PR #7, opened 2016-01-15
Add rotateEnabled to iOS and Android

christopherdro:


lelandrichardson:
LGTM. I wonder if we should ensure that all of these boolean properties have the same defaults on ios and android?

christopherdro:
I know zoom and rotate do, not sure about the other three. Can try those out later today.

PR #25, opened 2016-01-22
Fix when map is not loaded and app goes to background

jrichardlai:
I had an issue because I didn't set properly Google Service and the map was not rendering. So when leaving the app it just crashed because the map was not instantiated yet.

Not sure if thats the best way to handle it though.

### DIFF
--- a/android/src/main/java/com/AirMaps/AirMapView.java
+++ b/android/src/main/java/com/AirMaps/AirMapView.java
@@ -65,28 +65,6 @@ public class AirMapView extends MapView implements GoogleMap.InfoWindowAdapter, 
         super.onResume();
         super.getMapAsync(this);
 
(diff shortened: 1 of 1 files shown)


lelandrichardson:
@jrichardlai thanks for contributing!

What if instead we just moved the setting of the LifecycleEventListener to the `onMapReady` callback, where we know the map will be non-null after that?

jrichardlai:
Okay let me change that

jrichardlai:
No thanks for open sourcing this library, it is really useful!

jrichardlai:
@lelandrichardson does this looks fine to you? Seems I had to cast it to a `ThemedReactContext`.

lelandrichardson:
LGTM :+1:

PR #32, opened 2016-01-23
Fix showCallout bug for pin views on ios

### LINKED ISSUE #27: showCallout method usage
I'm trying programmatically show the callout of a particular marker by using the showCallout method under MapView.Marker but not having luck.

I've set the ref="myMarker" on a marker during render, and then in componentWillMount stage, use this.refs.myMarker.showCallout() but nothing seems to happen on the map. There's no error logged either.

Not sure if my usage is incorrect as I could not find an example usage in the folders. Is there a more detailed description on how to properly use the method?

Thanks

lelandrichardson:
Fixes #27 

This bug was was two-fold: pin views were always broken, and it looks like the callout views were broken since i added in the SMCalloutView library.

This should fix it!

### DIFF
--- a/example/App.js
+++ b/example/App.js
@@ -65,19 +65,19 @@ var App = React.createClass({
 
 
   render() {
-    return this.renderExamples([
-      [DisplayLatLng, 'Tracking Position'],
-      [ViewsAsMarkers, 'Arbitrary Views as Markers'],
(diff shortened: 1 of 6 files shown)


lelandrichardson:
(fixed in 0.2.1)

PR #58, opened 2016-01-30
passing style prop to AIRMap component

ex1machina:
I might be missing something but it seemed like the style prop wasn't being consumed anywhere so I couldn't do things like 

```
<MapView style={{flex: 1}} />
```


lelandrichardson:
@ex1machina styles passed in will already be included.  The `{...props}` call passes all props passed in to the component.

PR #80, opened 2016-02-15
add terrain map type

heydabop:


lelandrichardson:
@heydabop is this google maps only? is there a mapkit equivalent?

heydabop:
@lelandrichardson Oh, sorry, to my knowledge it's only Google Maps.

christopherdro:
@heydabop What would show on iOS if the map type was set to terrain? 

Could you also update the code comments and doc file to show that its Android specific?

Thanks!

heydabop:
@christopherdro I'm not sure I haven't tried. That being said if I try to set map to terrain in android and haven't applied this patch it crashes, IIRC. I can confirm this later, and change the documentation.

christopherdro:
@heydabop :Thum

heydabop:
Sorry, I cocked up my repo so I've had to make a new PR. [https://github.com/lelandrichardson/react-native-maps/pull/127](#127)

MiLeung:
Anyone know when the terrain map type will be supported on iOS?

Issue #77, opened 2016-02-12
Released unknown imageData reference


ghamaide:
I get this error on Android only when I try to update markers.
Here is my markers code : 

```
<MapView.Marker
  ref={restaurant.id}
  key={restaurant.id}
  coordinate={coordinates}
  onSelect={this.onSelect}
  pinColor={myRestaurant ? 'green' : 'red'}>
  <PriceMarker budget={restaurant.price_range} />
  <MapView.Callout>
    <TouchableHighlight underlayColor='rgba(0, 0, 0, 0)' onPress={() => this.props.navigator.push(Restaurant.route({id: restaurant.id}, restaurant.name))}>
      <View style={{flexDirection: 'row', alignItems: 'center', justifyContent: 'flex-start'}}>
        <Image source={{uri: restaurant.pictures[0]}} style={{height: 50, width: 50, marginRight: 5}} />
        <View>
          <Text style={{color: '#333333', fontSize: (Platform.OS === 'ios' ? 15 : 14), fontWeight: '500', marginBottom: 5}}>{restaurant.name}</Text>
          <Text style={{color: '#333333', fontSize: 13}}>{restaurant.food[1]}</Text>
        </View>
      </View>
    </TouchableHighlight>
  </MapView.Callout>
</MapView.Marker>
```

And here is the error : 

![screenshot_2016-02-12-15-51-49](https://cloud.githubusercontent.com/assets/6986811/13010242/1ad465c4-d1a1-11e5-8747-a5f0eae1c48f.png)

badasve:
I have the same error, it seems to be related to garbage collection of markers not  on the map anymore, any idea how to solve it ?

heydabop:
I was experiencing a similar error when I had an array of markers being rendered via a .map() operation, very similar to what happens in [DefaultMarkers.js#L60](https://github.com/lelandrichardson/react-native-maps/blob/master/example/examples/DefaultMarkers.js#L60). If I cleared the array, on the next render pass I would get this error.
I noticed that if I added `marker = null;` right after `marker.remove();` in [AirMapMarker.java#241](https://github.com/lelandrichardson/react-native-maps/blob/master/android/src/main/java/com/AirMaps/AirMapMarker.java#L241), it would prevent this error. I haven't tested other side effects to this, but setting the marker to null after removing it from the map prevents the manager from attempting to update it any further. Not sure if there's a situation where the same marker is removed and then added to the map again.

badaz:
[...]

PR #83, opened 2016-02-17
Potential fix for #77

heydabop:
#77 
set marker as null once it's removed from the map

Not yet sure if there are situations where a marker is removed from and then re-added to the map.


lelandrichardson:
Awesome. Thanks for fixing this!

I think we are safe. If they end up getting added back on to the map then the marker will be recreated. I think this is the ideal solution since memory pressure is so important on android.

PR #84, opened 2016-02-17
Use duration when animating setRegion:

jedlau:
`duration` was being passed, but wasn't being used when calling `setRegion:`.


lelandrichardson:
Awesome! Thanks.

PR #93, opened 2016-02-23
live update pinColor on iOS

machard:


ios/AirMaps/AIRMapMarker.m
```diff
@@ -223,5 +223,13 @@ - (void)setImageSrc:(NSString *)imageSrc
                                                           }];
 }
 
+- (void)setPinColor:(UIColor *)pinColor
+{
+    _pinColor = pinColor;
+    
+    if ([_pinView respondsToSelector:@selector(setPinTintColor:)]) {
```
lelandrichardson: do we need to check if `_pinView` is nil here before we can call `respondsToSelector:` on it?

PR #97, opened 2016-02-25
fitToElements: Implemented concatenation of overlays and annotations

micksabox:
I've implemented the concatenation of overlays and annotations on iOS. I haven't tested out the new method for calculating the bounding rect for the annotations, because my app doesn't use individual annotations yet.

### DIFF
--- a/docs/mapview.md
+++ b/docs/mapview.md
@@ -44,8 +44,8 @@
 |---|---|---|
 | `animateToRegion` | `region: Region`, `duration: Number` | 
 | `animateToCoordinate` | `region: Coordinate`, `duration: Number` | 
-| `fitToElements` | `animated: Boolean` | 
-
+| `fitToElements` | `animated: Boolean` | Fits both markers and overlays on the map
+| `fitToMarkers` | `animated: Boolean` | Fits markers on the map
 
 
 ## Types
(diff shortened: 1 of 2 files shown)


lelandrichardson:
I wonder if we should expose another method that does this for annotations only?  Are there scenarios where someone would want to zoom to fit annotations, but not overlays?

micksabox:
@lelandrichardson I was thinking that too. I think semantically the method name 'fitToElements' is very general, and I would interpret it as fit to annotations and overlays. But I agree with you. For example, in real estate you would want to show the homes (as annotations), but showing the neighbourhood boundaries (as overlays). You might want to have control over which type of elements to fit too,

lelandrichardson:
Perhaps we have `fitToElements()` as well as `fitToMarkers()`?

Also, I'd like to make sure we mimic this behavior in android before publishing this. If you're up to implementing that, that would be super appreciated! Otherwise, i'll take a crack at it and then merge the two PRs together.

micksabox:
The other consideration is that I'm not sure if we can do this on Android, since I'm not familiar with that SDK. Sorry, I won't be much help with the Android part.

lelandrichardson:
@micksabox no worries. If you want to take a crack at it, the API is fairly straightforward... would involve a couple of lines here: https://github.com/lelandrichardson/react-native-maps/blob/master/android/src/main/java/com/AirMaps/AirMapView.java#L344

In any event, when I get a chance I will do it myself if you have not and we can get this merged in.

Do you think in the mean time you can at least add an additional iOS method for `fitToMarkers()` that works the same way as `fitToElements()` did before?

micksabox:
@lelandrichardson just pushed the old method for `fitToElements()` renamed as `fitToMarkers()` and updated docs.

rborn:
@micksabox  we're trying to cleanup the repo, so I'll close this PR. Please ping me if you think you can continue working on it and I'll re-open 🤗

PR #98, opened 2016-02-25
forward onDestroy when removing view

sodik82:
PR regarding memory leak reported at https://github.com/lelandrichardson/react-native-maps/issues/95

implementation shows an improvement but still does not seem to cover the issue completely.

### DIFF
--- a/android/src/main/java/com/AirMaps/AirMapManager.java
+++ b/android/src/main/java/com/AirMaps/AirMapManager.java
@@ -80,7 +80,6 @@ public class AirMapManager extends ViewGroupManager<AirMapView> {
     
     @Override
     public void onDropViewInstance(AirMapView view) {
-        view.onPause();
         view.doDestroy();
         super.onDropViewInstance(view);
     }    
--- a/android/src/main/java/com/AirMaps/AirMapManager.java
+++ b/android/src/main/java/com/AirMaps/AirMapManager.java
@@ -77,6 +77,12 @@ public class AirMapManager extends ViewGroupManager<AirMapView> {
 
         return view;
     }
+    
+    @Override
+    public void onDropViewInstance(AirMapView view) {
+        view.onDestroy();
+        super.onDropViewInstance(view);
+    }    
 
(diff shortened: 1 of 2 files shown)


sodik82:
Cleaning up `LifecycleEventListener` helps to free additional resources.

experimentsin:
Looks really encouraging to me from an Android perspective. I'm afraid I'm not familiar enough with the React Native native side framework lifecycle or threading model on Android to know if there are any gotchas in this. Do all the callbacks changed here happen on the same (main UI) thread?

One small thing: depending on timing I guess a MapView could get unmounted (leading to an `onDropViewInstance()` call) a fraction before a pending `onHostPause()`, or vice versa, which could conceivably result in `onPause()` being called twice on the underlying `AirMapView`.

sodik82:
To be honest, I don't know RN threading model neither. If it is executed in one thread, it should not happen that `onPause` is called twice as the lifecycle callback is removed. 
However maybe it could happen that `onPause` is called as a result of lifecycle change (e.g. app is put to background) and then view is unmounted (or otherwise destroyed).

Is it problem if `onPause()` is called twice - from Android perspective - or is it just no operation then?

christopherdro:
/cc @mkonicek @kmagiera @satya164
Any chance any of you can take a peek at this and give us some input?

experimentsin:
> Is it problem if `onPause()` is called twice - from Android perspective - or is it just no operation then?

Unfortunately it depends entirely on how this particular `onPause()` is implemented inside `com.google.android.gms.maps.MapView` which I don't think is open source. You wouldn't get `onPause()` twice in a row from a conventional direct forwarding of Android's lifecycle callbacks, so it's likely untested and just depends on how Google have coded it.

sodik82:
thanks again :)

I have implement additional logic to prevent duplicate `onPause()`

lelandrichardson:
Thanks for your help with this. I've tested it out and everything looks good!  Sorry we dragged our feet on merging this...

PR #116, opened 2016-03-03
De/select annotation via map for show/hideCallout

jedlau:
**Problem:** If an annotation is displayed by calling
`showCallout()`, it can't be deselected by tapping away on the map. It
can only be dismissed by explicitly calling `hideCallout()`.

**Analysis:** When `showCallout()` is called on an annotation, it isn't
displayed through the map via `selectAnnotation:`. As a result, the
annotation is never deselected (and the `didDeselectAnnotationView:`
delegate method is never called), because the map didn't know it was
selected in the first place.

Fix:
- In `showCallout()`/`hideCallout()`, use
  `selectAnnotation:`/`deselectAnnotation:` through the map.
- In `showCalloutView:`/`hideCalloutView:`, don't call `setSelected:` on the
  MKAnnotationView. Quoting from the link below: "You should not call
  this method directly. An MKMapView object calls this method in
  response to user interactions with the annotation."

References:
- https://developer.apple.com/library/prerelease/ios/documentation/MapKit/Reference/MKAnnotationView_Class/index.html#//apple_ref/occ/instm/MKAnnotationView/setSelected:animated:

### DIFF
--- a/ios/AirMaps/AIRMapMarker.m
+++ b/ios/AirMaps/AIRMapMarker.m
@@ -134,8 +134,6 @@
 {
     MKAnnotationView *annotationView = [self getAnnotationView];
 
-    [self setSelected:YES animated:NO];
-
     id event = @{
             @"action": @"marker-select",
             @"id": self.identifier ?: @"unknown",
@@ -167,8 +165,6 @@
 {
     // hide the callout view
     [self.map.calloutView dismissCalloutAnimated:YES];
-
-    [self setSelected:NO animated:NO];
 
     id event = @{
             @"action": @"marker-deselect",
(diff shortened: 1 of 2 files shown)


christopherdro:
@jedlau Thanks for the PR! Any idea what the behavior on Android is?

jedlau:
No idea about Android.

brandonfancher:
I would love to see this merged as this problem is also affecting me.

lelandrichardson:
It would like to create a selection example to see how ios/android behave differently (if at all). Ideally the behavior is consistent.

JLLLinn:
Facing this issue right now. Right now without this PR do we have any way to select a certain marker in order to show the callout, so that user can click elsewhere to dismiss later?

rborn:
I'll close this for being very old. Please ping me if you consider it should re-opened after a full sync with the current stage of the module.
Thank you for you effort 😻

Issue #119, opened 2016-03-06
onRegionChange gives wrong location if map is tilted


heydabop:
If the map is in a tilted perspective, the coordinates returned by `onRegionChange` no longer line up with the center of the screen, or to coordinates the map programmatically gets navigated to.
To test this just call `animateToCoordinate` on a titled MapView and then check the values returned by `onRegionChange` or `onRegionChangeComplete`, and they won't be equal.

christopherdro:
@heydabop Could you try using to the `pitchEnabled` prop to disable tilting on that specific view?

heydabop:
This was fixed in my PR [#128](https://github.com/lelandrichardson/react-native-maps/pull/128)

PR #128, opened 2016-03-10
android: use camera target instead of region center for onRegionUpdate lat/lng

heydabop:
if the user tilts the camera the center of the viewed region is far off in the
distance, which would cause any actions based on onRegionChange lat/lng values
to appear off center

this also fixes lat/lng values being too far north/south if the user is very
zoomed out due to mercator distoritions


heydabop:
This should fix [#119](https://github.com/lelandrichardson/react-native-maps/issues/119)

lelandrichardson:
good catch :+1:

PR #136, opened 2016-03-15
Fix crash on iOS 7 and 8 with iOS 9+ API calls

jerolimov:
This fixes a crash when iOS 7 and 8 mapview properties `showsCompass`, `showsScale` or `showsTraffic` are defined. These properties are only available on iOS 9+.

Similar to the origin react-native MapView (RCTMap.m) this checks if the super class (MKMapView) has selectors to set and get these parameters. If not it does nothing or return false.

### DIFF
--- a/ios/AirMaps/AIRMap.m
+++ b/ios/AirMaps/AIRMap.m
@@ -204,4 +204,49 @@ const CGFloat AIRMapZoomBoundBuffer = 0.01;
     }
 }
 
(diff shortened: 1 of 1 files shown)


jerolimov:
@lelandrichardson ps: Thanks for sharing the great lib / api. Great work. :+1:

lelandrichardson:
@jerolimov Excellent! Thank you.

PR #142, opened 2016-03-20
Add pitchEnabled property to MapView

jadnco:


jadnco:
@lelandrichardson @christopherdro Any chance on getting this merged in? :)

christopherdro:
@jadnco Sorry for the late response, will check on that tonight?

christopherdro:
@jadnco Looks good. Could you squash your commits before we merge this in?

lelandrichardson:
@jadnco what is the meaning of the center coordinate of the region when tilt is enabled?

jadnco:
@lelandrichardson Nothing would change from what it already means. Tilt only changes when users start interacting with the map (swiping up/down with two fingers). That feature is already enabled by default, so adding the `pitchEnabled` property just allows developers to disable it.

Does that answer the question?

christopherdro:
👍 Thank you.

mbmelgo:
how to set the pitch of the map when the map was loaded?

Issue #138, opened 2016-03-15
Java compilation error for getPackages()


Pocket-titan:
Hi,

In Installation.md, it states that (when manually installing), the following code should be added to MainActivity.java: 
```
...
@Override
 protected List<ReactPackage> getPackages() {
   return Arrays.<ReactPackage>asList(
     new MainReactPackage()
     new AirPackage() // <---- and This!
   );
 }
```

However, for me, this results in a compileDebugJavaWithJavac error when building.
Adding a "," after MainReactPackage() fixes the error, so:

```
...
new MainReactPackage()
new AirPackage() 
```

would become:

```
...
new MainReactPackage(),
new AirPackage()
```

If this is an error others also experience, I can make a quick pull request to Installation.md for the missing colon.

christopherdro:
@Pocket-titan Good catch!

Can you submit a PR for that? Thanks!

Pocket-titan:
@christopherdro Sure! Pull request is here: #145

PR #145, opened 2016-03-21
Add comma

Pocket-titan:
Elements must be separated by a comma, was not the case before resulting in a compilation error.
Reference: #138


christopherdro:
thanks!

Closes #138

PR #159, opened 2016-03-30
On pan drag

jrichardlai:
Allow pan drag event from the map, this allow to drawing more smoothly polyline and polygons.

iOS seems to be fine, but Android seems a little laggy on Simulator and would crash after lots of drawings, ( which has not happened yet on the phone for me ), but I'd assume its about the high number of `setState` calls.

Note: for iOS `scrollEnabled` should be set to false to trigger the event


lelandrichardson:
@jrichardlai do you have a gif of this in action? Pretty cool.

jrichardlai:
Hey @lelandrichardson, sorry been pretty busy lately, here is a gif. In Android it's pretty laggy and would crash if you do a setState for all coordinates, so I'm adding a point for our app only after a delta.
We've been using it in prod for a week with no problems so far. Do you know a better way to send the drag with Android.

out

christopherdro:
@jrichardlai LGTM! How has this been performing in the last week. Still good?

/cc @lelandrichardson Can you look over this and approve?

lelandrichardson:
@jrichardlai @christopherdro this looks good to me. It's correct to assume that there is no performance degradation if you aren't using this event, right?

christopherdro:
@jrichardlai Could we also add 'Note: for iOS scrollEnabled should be set to false to trigger the event' inside the docs somewhere?

jrichardlai:
@lelandrichardson @christopherdro yes no issues noted so far.

For the performance in iOS I am still adding a `GestureResponder` so `handleMapDrag` will still be called but will return if `onPanDrag` is not set.
However on Android, its using `getExportedCustomDirectEventTypeConstants` and so `onPanDrag` is called on the drag, we didn't noticed an issue with scrolling but I think the performance could be improved, is there a way to return like with iOS?
I'll update the doc to have a note for the iOS scrollEnabled.

lelandrichardson:
hmm. not sure the right way to solve that...

lelandrichardson:
:shipit:

jrichardlai:
@lelandrichardson I'm gonna do more testing for performance on Android

jrichardlai:
@lelandrichardson after some testing on a less powerful phone I noticed that there was unfortunately some lag issues because of the onPanDrag call. I made a little tweak https://github.com/lelandrichardson/react-native-maps/pull/159/commits/6253f7bee1778dee6bffbd4a4472c0dcd3c96bed is there another way you think we can fix that?

jrichardlai:
@lelandrichardson I think this PR is gtg been using this code for few weeks now

android/lib/src/main/java/com/airbnb/android/react/maps/AirMapView.java
```diff
@@ -80,12 +82,12 @@ public AirMapView(ThemedReactContext context, AirMapManager manager) {
 //                return false;
 //            }
 
-                    @Override
-                    public boolean onScaleBegin(ScaleGestureDetector detector) {
-                        view.startMonitoringRegion();
-                        return true; // stop recording this gesture. let mapview handle it.
-                    }
-                });
```
jrichardlai: I'll fix the spaces

jrichardlai:
Thanks for merging @christopherdro ! 👍

yaronlevi:
Hi can you please reference a place in the docs where it explains about OnPanDrag or how to achieve the line draw you built in the GIF above?

I've looked all over the docs and can not find this.
Any help will be much appreciated.

jrichardlai:
You can look at the example https://github.com/airbnb/react-native-maps/blob/master/example/examples/PolylineCreator.js let me know if that helps

yaronlevi:
@jrichardlai Thank! just what I was looking for. Going to try it out.

cgelosulit:
"Was this removed in the example?"

PR #168, opened 2016-04-07
Remove hard-coded play-services API version in deps

rreusser:
**Problem:** Cannot be used out of the box with [`react-native-gcm-android`](https://github.com/oney/react-native-gcm-android) since that required play-services `8.3.0`. Resulting name collision results in cryptic runtime errors. For comparison, [`react-native-gcm-android` installation instructions](https://github.com/oney/react-native-gcm-android#installation) include:

```
    compile 'com.google.android.gms:play-services-gcm:8.3.0' // <- Add this line
```

**Solution:** Feel free to accept or reject this PR, but the convention of moving dependencies to the app makes a certain amount of sense to me so that the repo doesn't need to be forked in order to fix the issue. This approach requires negligible extra work, has the same result, and makes the dependency explicit and somewhat easier to recognize and track down.

Thanks for the plugin!

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -17,6 +17,4 @@ android {
 
 dependencies {
     compile 'com.facebook.react:react-native:0.19.+'
-    compile "com.google.android.gms:play-services-base:8.4.0"
-    compile 'com.google.android.gms:play-services-maps:8.4.0'
 }
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -37,11 +37,13 @@ or do it manually as described below:
      project(':react-native-maps').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-maps/android')
    ```
 
-2. in `android/app/build.gradle` add:
+2. in `android/app/build.gradle` add the following lines. If your project requires a different version of the play services API, you may need to, for example, downgrade to 8.3.0.
    ```
    dependencies {
        ...
        compile project(':react-native-maps')
+       compile 'com.google.android.gms:play-services-base:8.4.0'
+       compile 'com.google.android.gms:play-services-maps:8.4.0'
    }
    ```
 


christopherdro:
This works for me. @lelandrichardson Is this ok for you?

rreusser:
(I should add that I'm new to android—so if you're aware of reasons to do it one way or the other, you almost certainly know better than I do! Just happened to see a couple forks here and that that only change the play services version, so this seemed perhaps desirable.)

christopherdro:
@rreusser Ya it's a tough one to call. I've noticed that with bigger projects it's nice to have third party module stripped of their dependencies and force the user to set them up in their own app.

I like this change and it could potentially save a lot of headache.

rreusser:
Ahhhh. @christopherdro — This might have to get reverted… This worked fine up until I removed all of the intermediate files and node_modules and rebuilt. Now I'm getting:

```
node_modules/react-native-maps/android/src/main/java/com/AirMaps/AirMapCircle.java:5: error: package com.google.android.gms.maps does not exist
import com.google.android.gms.maps.GoogleMap;
...
```

It looks like `react-native-gcm-android` [does actually include play-services](https://github.com/oney/react-native-gcm-android/blob/master/android/build.gradle#L21) in its own `build.gradle`, though a different version.

Sorry for the confusion/waste of time. Long story short, I think I failed to *fully* test this on a fresh clone to ensure it wasn't only working because it had worked before. I'm reading up on the [dependency management docs](https://docs.gradle.org/current/userguide/dependency_management.html) now… :disappointed:

rreusser:
Eek. Okay, the official docs say:

> Manage transitive dependencies
>
> If you try to manage transitive dependencies yourself, the end of the story is that your build becomes brittle: no one dares to change your dependencies because the risk of breaking the build is too high. The project classpath becomes a complete mess, and, if a classpath problem arises, hell on earth invites you for a ride.

It sounds like I misunderstood the nature of the conflicting dependencies. I thought manual management was the proper solution, but it sounds like the true solution is to properly understand dependency resolution.

normanrz:
I have the same issue that the dependencies aren't resolved properly. Hope this PR can be reverted.

rreusser:
@normanrz very sorry. My bad. :disappointed: In the mean time, it should work to use:

```
"react-native-maps": "git@github.com:lelandrichardson/react-native-maps.git#096a7f23eec8ed1e60715b2e06c7aea0e095ef52"
```

normanrz:
No worries. I re-added the deps in the module's `build.gradle` for now.

PR #178, opened 2016-04-12
Cocoapods spec + installation instructions

aforty:
I use cocoapods for some other dependencies (not react-native related) and rnpm never plays nice with it. The podspec file is all that's needed for someone to install via cocoapods.

You don't have to publish this to cocoapods (although you can if you want), but you do have to keep the version updated within the podspec file going forward.

I hope you'll accept this to make some of our lives easier. Thanks for the awesome package!

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -16,6 +16,12 @@ or do it manually as described below:
 
 ## iOS
 
+### Cocoapods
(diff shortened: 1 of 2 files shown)


christopherdro:
Thanks!

PR #181, opened 2016-04-14
Remove locale from Google API Console url

jrichardlai:


KGALLET:
Oops didn't see that, good job :) +1

PR #188, opened 2016-04-20
Remove style instructions from README

npomfret:
removed the style guide in README.md for rendering a map with initialRegion because its not required (at least as of RN 0.23), and added some more instructions on getting google play services updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,22 +31,6 @@ declaratively controlling features on the map.
 
 ### Rendering a Map with an initial region
 
-When you render a MapView, this one need to be in absolute position with the top, left, right and bottom props set.
-If not you gonna have a blank view.
-
-## Styles
-```jsx
-const styles = StyleSheet.create({
-  map: {
-    position: 'absolute',
-    top: 0,
-    left: 0,
-    right: 0,
-    bottom: 0,
-  },
-});
-```
-
 ## MapView
 ```jsx
   <MapView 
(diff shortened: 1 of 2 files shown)


jrichardlai:
Thanks for the PR! I think "Genymotion" is more correct, also I think we should remove then the `styles.map` from the following section in the doc.

jrichardlai:
Seems there was a an update of the doc recently https://github.com/lelandrichardson/react-native-maps/pull/120 not sure which version of RN was used. I'm not using the absolute positioning so I think it's fine to remove it. @lelandrichardson what do you think?

lelandrichardson:
Are we sure the style is completely unneeded now?  What changed?

If we don't need it, let's definitely remove it... but maybe an alternative approach would be to have that part of the documentation being like "Your map not showing up? Maybe try this..." or something.

If it's completely unnecessary in RN latest though i'm all for getting rid of it.

npomfret:
I tested on RN 0.23.1 on both Android and IOS.

christopherdro:
I vote we start a troubleshooting section and have this in there as a possible workaround for a map not showing up.

lelandrichardson:
@christopherdro 👍  I support this idea.

jrichardlai:
Sounds good to me 👍 , would be good since there seems to be various issues.

jrichardlai:
Made a PR to update the doc https://github.com/lelandrichardson/react-native-maps/pull/193

christopherdro:
@npomfret Thanks for submitted this PR.

Closing for #193.

PR #192, opened 2016-04-22
converted to RN 0.24+

jgibbons:


christopherdro:
Is this backwards compatible with older versions of RN?

jrichardlai:
Merging this as RN 0.26 is coming up

christopherdro:
Please update the docs, preferably at the beginning of the README explaining this change and what versions of react native this will effect.

jrichardlai:
Yep sounds good!

christopherdro:
👍  Thank you.

jrichardlai:
@christopherdro I did some test, and it seems that this actually work out of the box from [0.18](https://github.com/facebook/react-native/releases/tag/v0.18.0), which is when react was used from npm. But otherwise it would also work if react is added as a dependency to the project for version lower than 0.17. I will write a note about it.

christopherdro:
That sounds great! We can add this to the 'Compatibility' section.

scottluptowski:
@jrichardlai Thanks for this update!

Will `package.json` be updated to reflect this change? I am still getting a deprecation warning in my app because my `package.json` is pointing to 0.3.2, which according to Git was 3 weeks ago. I'd prefer for my app to track a version number instead of `master` of a git branch.

Thanks again

PR #194, opened 2016-04-23
animateToCoordinate arg is a Coordinate object

yonahforst:
if you nest it inside `{region: Coordinate, duration: Number}`


christopherdro:
We still need to have duration in there, right?

yonahforst:
Ok, this is my misunderstanding. I thought the documentation meant a single argument with region and duration key. I didn't realize that it's actually two arguments: coordinate, duration.

https://github.com/lelandrichardson/react-native-maps/blob/master/components/MapView.js#L338

PR #195, opened 2016-04-24
Fit to supplied markers

AidenMontgomery:
In iOS it is now possible to provide a set of Marker IDs and have the
map change region to fit them on screen.


jrichardlai:
Thanks for the PR, do you have an example that you can add?

AidenMontgomery:
I can sort one out for you. I will try and get it done this evening.

AidenMontgomery:
I realised that I hadn't included my changes for Android, and then found that I had lost them.
I am trying to implement the change for Android as well, but I am having some issues with an error that I can't track down.

AidenMontgomery:
I have added Android support for this feature and updated the documentation, including the demo on the main Readme.
There is now a new example in the example application which shows focussing on different sets of markers.

ios/AirMaps/AIRMapMarkerManager.m
```diff
@@ -33,7 +33,7 @@ - (UIView *)view
     return marker;
 }
 
-//RCT_EXPORT_VIEW_PROPERTY(identifier, NSString)
+RCT_EXPORT_VIEW_PROPERTY(identifier, NSString)
```
jrichardlai: @lelandrichardson is there a reason why this was commented out? I saw this https://github.com/lelandrichardson/react-native-maps/issues/37 would this resolve the issue?

jrichardlai:
Do we need to add the `android/gradle/wrapper/gradle-wrapper` and `android/gradle/wrapper/gradle-wrapper.properties` files?

AidenMontgomery:
@jrichardlai I don't think so. I thought I had removed all of the new files that I didn't explicitly create. I will remove them now.

jrichardlai:
@AidenMontgomery looks good! I made a PR to your branch to use the animated bool! Good job! the 50 padding looks fine I think its okay to have this as default.
@lelandrichardson thoughts?

AidenMontgomery:
@jrichardlai Thanks for the addition. Would you like the gif's updating too?

jrichardlai:
A gif would be good to showcase :) !

AidenMontgomery:
There are two in the pull request already, in the main ReadMe file. There are two problems with them though. The Android gif is not using Genymotion, I haven't found the time to get maps working in there yet. The second issue is that they now don't match the updated version of the sample, as all transitions are animated.

christopherdro:
@lelandrichardson LGTM. Any thoughts?

ios/AirMaps/AIRMapMarkerManager.m
lelandrichardson: I was playing around with this at some point but never really got it to work the way I wanted. I was hoping it could be more automatic.  Another thing is that we should make the public JS api accept a string or a number, rather than just a string (since numeric ids are common).

lelandrichardson:
this LGTM. As I mentioned above, we could make the JS api allow for non-string identifiers, but that could be done in a followup PR as well.

If `identifier` becomes a thing, we should also start passing it back in all of the events.

AidenMontgomery:
I've had a look at accepting either a string or a number. I'll admit that I am not too sure what I am doing at this point, but I have set the identifier property to id in iOS and set the PropType to be PropType.oneOfType([PropType.string, PropType.number]), this seems to be working ok for iOS.
When it comes to Android and Java, I don't know what the equivalent of the id type would be, to let the property be one or the other type.
I can include what I have done so far in this PR, or if you are happy to accept this one now, then I could do another PR with my progress and someone with better Java skills could finish off the Java side.

lelandrichardson:
@AidenMontgomery Java doesn't have a great way to handle this yet (that I know of). I don't think this should block this PR though!

christopherdro:
@AidenMontgomery Would you mind rebasing this against master. I think the merge conflicts should just be from the documentation. Thanks!

AidenMontgomery:
I think that this is now completed and I don't think that I have broken anything.
There were a lot of conflicts with the snapshot taking feature, as we had inserted the new functionality in the same location in the files.

I attempted to run this through the example project, but had a lot of issues, due to the move of createClass from react-native to react in v26.

AidenMontgomery:
This had become way too stale. So I cleaned everything up and I will open a new pull request with a re-based and not messed up branch.

PR #203, opened 2016-04-29
Taking snapshots of map for ios

rusel1989:
Added functionality to take snapshot of map with annotation views for ios. You can configure image size and center point. I have tested it both in simulator and real device (iPhone SE). I also work on android implementation as i need to use it in both platforms for app I am building. So I will soon send pull request with android version as well (If this one is accepted :-)). PR contains example and readme update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,42 @@ render() {
 }
 ```
 
+### Take Snapshot of map
+currently only for ios, android implementation WIP
+
+```jsx
+getInitialState() {
+  return {
+    coordinate: {
+      latitude: LATITUDE,
+      longitude: LONGITUDE,
+    },
+  };
+}
+
+takeSnapshot () {
+  // arguments to 'takeSnapshot' are width, height, coordinates and callback
+  this.refs.map.takeSnapshot(300, 300, this.state.coordinate, (err, snapshot) => {
+    // snapshot contains image 'uri' - full path to image and 'data' - base64 encoded image
+    this.setState({ mapSnapshot: snapshot })
+  })
+}
+
+render() {
+  return (
+    <View>
+      <MapView initialRegion={...} ref="map">
+        <MapView.Marker coordinate={this.state.coordinate} />
+      </MapView>
+      <Image source={{ uri: this.state.mapSnapshot.uri }} />
+      <TouchableOpacity onPress={this.takeSnapshot}>
+        Take Snapshot
+      </TouchableOpacity>
+    </View>
+  );
+}
+```
+
 ### Troubleshooting
 
 #### My map is blank
--- a/components/MapView.js
+++ b/components/MapView.js
@@ -335,6 +335,12 @@ var MapView = React.createClass({
     this._runCommand('fitToElements', [animated]);
   },
 
+  takeSnapshot: function (width, height, centerPoint, callback) {
+    const region = this.props.region || this.props.initialRegion;
+    centerPoint = centerPoint || { latitude: region.latitude, longitude: region.longitude }
+    this._runCommand('takeSnapshot', [width, height, centerPoint, callback]);
+  },
+
   _getHandle: function() {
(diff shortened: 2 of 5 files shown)


jrichardlai:
Thanks, this is pretty cool! I tested it works pretty well. Did you look at Android? If you have it it will be better to merge this together, otherwise that's fine with me.
Also is there always a "Documents" folder that is named "Documents" internal at the application ?

rusel1989:
Hey, thanks for response. I looked at the android, but sadly, it seems that currently there is no possibility to pass data back to js from native view, as you can see [here](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewManager.java#L129). 
I am not 100% sure with the Documents folder, i will look at it later today and eventually fix it. Also i will change the method to take whole region object instead of just center point so you can adjust zoom level of snapshot. I will update pull request later today.

rusel1989:
Checked the Documents folder it is always there see [link](https://developer.apple.com/library/ios/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html). Also changed the centerPoint parameter to region and updated example, so let me know if it is ready for merge or if I should make some more changes. 
When it will be possible to make this work in android I will immediately add this functionality for android as well.

christopherdro:
@rusel1989 Looks great!

I would suggest we use RCTPromiseResolveBlock & RCTPromiseRejectBlock in place of the standard callback. This would then require alter takeSnapshot() a bit but would keep things asynchronous.

We can also offer a option to store the photo in the documents or temp directory. Having access to the temp directory is useful if the app needs to take screenshots that don't need to be accessed again.
Here is an example of that. https://github.com/christopherdro/react-native-html-to-pdf/blob/master/RNHTMLtoPDF/RNHTMLtoPDF.m#L77-L84

Re: Android. What kind of data are trying to pass back?

rusel1989:
@christopherdro Thanks 

Okay i will switch takeSnapshot method to promises and add directory option, i will also change the method to take options object instead of arguments in case anyone would like to add more options.

In Android I would like to pass same data as for ios - uri and base64 encoded image ... but with receiveCommand it is not possible to pass anything back to js. Do you have some suggestions how to achieve that ?

christopherdro:
@rusel1989 I think sending base64 data across the bridge can be expensive therefore recommend just sending the back the file path.

Unfortunately I'm not familiar with receiveCommand, but if you were have android take a snapshot you should be able to return the filepath.

BTW, does taking screenshots currently work if you have markers, polylines, or polygons rendered on the map?

FWIW, taking snapshots of views is currently available in core for iOS. There is a PR open to implement this on android but it seems that there is some discussion on merging it in.
https://github.com/facebook/react-native/pull/6431/files

Maybe we can combine our efforts and get this merged into core?

christopherdro:
Ok just tested this out and seems to work nicely. 
I think it might makes this more complicated than they need to be try swapping the standard callbacks out for promises. 

I'm going to pull this in now and work on the feature for Android. I most likely make a small update that will allow you to pass in width, height and other options and a single object. This will allow users to specify if they want the images save in a temp or perm directory and an option for the returned data format.

DVassilev:
@rusel1989  Is it possible to have the snapshot also capture the markers?

rusel1989:
@DVassilev Hey, it renders all markers by default even the custom ones

DVassilev:
@rusel1989 I was making snapshots and they didn't include the markers, just the maps.

rusel1989:
@DVassilev That is weird, it works for me, can you provide gist with your code ??

DVassilev:
Not around laptop now but I basically generate a bunch of Markers with a custom image and text, when I try to take snapshot (done as per the example) it just generates an image of the map. Will see if I can post it later.

Issue #151, opened 2016-03-23
When showsUserLocation is true initialRegion is always the user's location on iOS


lucasfeliciano:
If you enable the prop `showsUserLocation` even if you pass `initialRegion` the map is centralised by the user's position on iOS

lucasfeliciano:
If i use `region` props it works

lucasfeliciano:
Reopening it because

christopherdro:
It appears this is the default behavior.
>
This property does not indicate whether the user’s position is actually visible on the map, only whether the map view should try to display it. Setting this property to YES causes the map view to use the Core Location framework to find the current location and try to display it on the map. **As long as this property is YES, the map view continues to track the user’s location and update it periodically. The default value of this property is NO.**

Reference: https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKMapView_Class/index.html#//apple_ref/occ/instp/MKMapView/showsUserLocation

heydabop:
Sorry to bring this back up, but I think you're misreading the intention there.

>the map view continues to track the user’s location and update it periodically.

I'm fairly certain it just means it updates the marker for the user's location periodically, not that it updates the map's region periodically. I don't see anything in that description regarding the map's location.

Also, you explicitly tell it to follow a user's location in [AirMap.m:179](https://github.com/lelandrichardson/react-native-maps/blob/master/ios/AirMaps/AIRMap.m#L179). I commented this line out and it seemed to work fine. On that note, @lucasfeliciano, if you want the map to stop forcing itself to the user's location, just comment out that line and rebuild your project.

PR #212, opened 2016-05-05
Separate showing and following user

athaeryn:
As mentioned in #151, setting `showsUserLocation` also animates the map to the users location, but this is not always desired. To allow more control, I've added another property, `followsUserLocation`. This allows individual control of these two behaviors.


christopherdro:
Thanks for the PR!

Is this a issue on Android as well?

If `followUserLocation` if for iOS only can you add `@platform ios` to the bottom of props comment.?

athaeryn:
Not an issue on Android. I'll add the `@platform ios` line!

christopherdro:
Thank you!

AugustoAleGonZhipCode:
Do you have any idea how to do the followsUserLocation true works on Android? I have a problem with that.

PR #214, opened 2016-05-06
Update README.md

jrichardlai:
Add note about changes in requires in react-native-maps


jrichardlai:
@christopherdro maybe the note header is too much?

christopherdro:
LGTM. Thanks for doing that so quick.

PR #225, opened 2016-05-13
Android code cleanup and preparation for Maven deployment

felipecsl:
* Set the package name to `com.airbnb.android.react.maps` instead of `com.AirMaps`
* Adds gradle scripts for Maven deployment
* Removes dependency on NDK (seems to be unnecessary)
* General cleanup and organization around the android source code (formatting, project structure, etc).


android/lib/src/main/java/com/airbnb/android/react/maps/maps/SizeReportingShadowNode.java
```diff
@@ -1,18 +1,19 @@
 /**
  * Copyright (c) 2015-present, Facebook, Inc.
  * All rights reserved.
- *
+ * <p/>
```
lelandrichardson: ???

felipecsl: weird, seems to be added automatically by Android Studio. Will remove

android/lib/src/main/java/com/airbnb/android/react/maps/maps/AirMapManager.java
```diff
@@ -211,11 +210,13 @@ public void receiveCommand(AirMapView view, int commandId, @Nullable ReadableArr
     }
 
     @Override
-    public @Nullable Map<String, Integer> getCommandsMap() {
+    public
+    @Nullable
+    Map<String, Integer> getCommandsMap() {
```
lelandrichardson: this seems like strange formatting? is this typical?

felipecsl: no, wrong too. Fixing

lelandrichardson:
Can you also update the naming in these instructions? https://github.com/lelandrichardson/react-native-maps/blob/master/docs/installation.md

lelandrichardson:
cc @jrichardlai @christopherdro

felipecsl:
usage instructions updated!

jrichardlai:
Thanks @felipecsl. 
@lelandrichardson I'm not sure what are the implications of using Maven.
Should the example project be updated?
Also was not possible to use `com.airbnb.android.react.maps` ?

felipecsl:
@jrichardlai the sample should probably be updated too. I see that it lives in a different root folder though. Ideally it should be under `android` too, so we can easily make it depend on the library. 
I can definitely do that as a follow up after this PR is merged.

lelandrichardson:
@jrichardlai updated to use only one `.maps` :)

lelandrichardson:
It's unclear to me what the right example project structure is because I think needing it to run on both ios and android might change some things...

android/lib/src/main/java/com/airbnb/android/react/maps/AirMapMarkerManager.java
```diff
@@ -135,40 +130,44 @@ public void removeViewAt(AirMapMarker parent, int index) {
     }
 
     @Override
-    public @Nullable Map<String, Integer> getCommandsMap() {
+    public
+    @Nullable
```
jrichardlai: is that intended ?

android/lib/src/main/java/com/airbnb/android/react/maps/AirMapMarkerManager.java
```diff
@@ -23,11 +22,7 @@
     public static final int SHOW_INFO_WINDOW = 1;
     public static final int HIDE_INFO_WINDOW = 2;
 
-    private AirMapCalloutManager calloutManager;
-
-    public AirMapMarkerManager(AirMapCalloutManager calloutManager) {
-        super();
-        this.calloutManager = calloutManager;
```
jrichardlai: We don't need it anymore?

felipecsl: this was never used

android/lib/src/main/java/com/airbnb/android/react/maps/AirMapMarkerManager.java
felipecsl: fixing

android/lib/src/main/java/com/airbnb/android/react/maps/AirMapMarkerManager.java
lelandrichardson: strange. wonder if this is a bug we didnt know about, or just an oversight? lol. I must have put it there for a reason???

felipecsl:
@lelandrichardson hmm that's true, i didn't realize that.
We can always use the trick `project(':react-native-maps').projectDir = new File(rootProject.projectDir, ...` to make it work, so I think it won't be a problem

jrichardlai:
lgtm! For my knowledge whats the advantage of using Map instead of HashMap?

android/lib/src/main/java/com/airbnb/android/react/maps/AirMapMarkerManager.java
felipecsl: no idea 😕

felipecsl:
@jrichardlai it is a common java idiom to use `List` and `Map` instead of `ArrayList` of `HashMap` whenever possible. That is because whether it's an ArrayList or LinkedList is an implementation deatail and you don't need to make your type so restrictive, since all you care about is that it's a List anyways. Same goes for Map, it's just more flexible this way

lelandrichardson:
@jrichardlai regarding the implications of maven...

The React Native community seems to be hell-bent on using NPM for everything. When integrating RN with existing large code bases that use existing platform conventions and package managers (gradle/maven/cocoapods/etc), this can come as quite a shock and I think is quite honestly hurting RN's adoption amongst existing native developers.

For use inside of Airbnb, we are going to start using Maven and Cocoapods to depend on even 3rd party native modules that aren't publishing them themselves. We will still be publishing the source through NPM which users can install the old way if they'd like,

@felipecsl anything to add to this?

felipecsl:
@lelandrichardson that's absolutely accurate. Just to make it 100% exact, we're been already using Maven and Cocoapods for 3rd party dependencies on Android and iOS respective for quite a while, as these are the de-facto standards for these platforms

felipecsl:
Ah I guess you mean 3rd party react native modules specifically. So yeah, you're right!

lelandrichardson:
@felipecsl Right. by "start using", i really meant "start using for react native" :)

lelandrichardson:
k. Going to merge this in! Thanks, @felipecsl

jrichardlai:
Thanks for the explanation @felipecsl.
@lelandrichardson  👍

jeveloper:
Can you clarify what changes i would need to make to use version 0.5 
" Updated package to `com.airbnb.android.react.maps`" , is this something i would need to modify in MainActivity 

Or perhaps installation will be updated.

thank you

felipecsl:
@jeveloper the usage instructions are still being updated/fixed (see #237), but mainly you have to use `MapsPackage` instead of `AirPackage`

jeveloper:
Thank you @felipecsl  👍 

would it be better to just remove npm and install it again , not sure if there is a rnpm relink of some kind that will make all necessary changes

PR #241, opened 2016-05-19
Allow other versions of react-native than 0.25.1

jrichardlai:
Right now it seems that android is locked to 0.25.1.

I know we made the change recently after react-native-maps has been put on maven, just wondering if it is necessary to lock the version going forward or using  'com.facebook.react:react-native:+' is fine.

### DIFF
--- a/android/lib/build.gradle
+++ b/android/lib/build.gradle
@@ -16,7 +16,7 @@ android {
 }
 
 dependencies {
-  compile 'com.facebook.react:react-native:0.25.1'
+  compile 'com.facebook.react:react-native:+'
   compile "com.google.android.gms:play-services-base:8.4.0"
   compile 'com.google.android.gms:play-services-maps:8.4.0'
 }


jrichardlai:
@lelandrichardson @felipecsl thoughts ?

felipecsl:
This is usually considered a bad idea because you can't have repeatable builds since the version could change and you're unaware of that. It's strongly recommended that you lock to a specific version, that is, don't use the `+` option.
When other apps include this library, if they already use a version of RN newer than 0.25.1, Gradle will automatically pick the newest version in case one is used.
To summarize, I think this is a bad idea 😄

jrichardlai:
@felipecsl that sounds good, thanks for the answer. So it seems that its better to set a version of RN here so is it possible to set it to a lower one? Previously 0.18.0 was supported out of the box. Not sure if we should use this version as it is from January.

felipecsl:
You don't want to set one that's too old either since it may break for people using the current.
I guess the usual plan is to just set it to whatever is the current latest version :)

christopherdro:
Have any of you tried to build the example on android with the latest changes?

jrichardlai:
@felipecsl I tried again the current master and I can't run the example project, using this branch fixed the issue. 

Prior to that I'd get the issue:

```
* What went wrong:
A problem occurred configuring project ':app'.
> A problem occurred configuring project ':react-native-maps'.
   > Could not resolve all dependencies for configuration ':react-native-maps:_debugCompile'.
      > Could not find com.facebook.react:react-native:0.25.1.
        Searched in the following locations:
file:/Users/jrichardlai/.m2/repository/com/facebook/react/react-native/0.25.1/react-native-0.25.1.pom
file:/Users/jrichardlai/.m2/repository/com/facebook/react/react-native/0.25.1/react-native-0.25.1.jar
https://jcenter.bintray.com/com/facebook/react/react-native/0.25.1/react-native-0.25.1.pom
https://jcenter.bintray.com/com/facebook/react/react-native/0.25.1/react-native-0.25.1.jar
file:/Users/jrichardlai/Projects/react-native-maps/node_modules/react-native/android/com/facebook/react/react-native/0.25.1/react-native-0.25.1.pom
file:/Users/jrichardlai/Projects/react-native-maps/node_modules/react-native/android/com/facebook/react/react-native/0.25.1/react-native-0.25.1.jar
file:/usr/local/opt/android-sdk/extras/android/m2repository/com/facebook/react/react-native/0.25.1/react-native-0.25.1.pom
file:/usr/local/opt/android-sdk/extras/android/m2repository/com/facebook/react/react-native/0.25.1/react-native-0.25.1.jar
file:/usr/local/opt/android-sdk/extras/google/m2repository/com/facebook/react/react-native/0.25.1/react-native-0.25.1.pom
file:/usr/local/opt/android-sdk/extras/google/m2repository/com/facebook/react/react-native/0.25.1/react-native-0.25.1.jar
Required by:
android:react-native-maps:unspecified

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
```

I also tried changing it to `compile 'com.facebook.react:react-native:0.26.1'` the same issue comes up.

@christopherdro can you try out the change ?

christopherdro:
I tried this yesterday and still couldn't get it to work. Will give it a shot again and loop back.

christopherdro:
I suggest the build tools version on the lib and example stays the same with React Native to avoid confusing have needing two different build tools versions.

Or we can be ambitious and update RN android to use the latest build tools version :)

jrichardlai:
@christopherdro I agree on the build tool, using the same one as React Native would reduce overhead. `com.android.tools.build:gradle:1.3.1` seems to be the one used by the default template of RN.

christopherdro:
It appears thats the gradle file checked into maven needs to be editing as well. I'm not too familiar with how that works. @lelandrichardson @felipecsl Any ideas?

mangogogos:
Any updates on this? I fail to see why you would require this package to be locked to a specific React Native version and as it is this is blocking me from doing work on upgrading my app to RN 0.26

mangogogos:
Any updates on this? As it is I need to compile this project via compile 'com.airbnb.android:react-native-maps:0.5.0' rather than from the node_modules gradle project i.e. compile project(':react-native-maps')

christopherdro:
Pinging @lelandrichardson @felipecsl

PR #252, opened 2016-05-25
Add support note for onSelect and onDeselect

jrichardlai:
`onSelect` and `onDeselect` are only available for iOS at the moment.


christopherdro:
@jrichardlai Would you mind adding `@platform ios` inside `Marker.js` where the prop type is defined?

PR #256, opened 2016-05-27
Marker select deselect doc update

jrichardlai:
`onSelect` and `onDeselect` are only available for iOS at the moment.

Added `@platform ios` for `onDeselect ` in `MapMarker.js` but it was already there for `onSelect`


christopherdro:
Thank you sir!

PR #258, opened 2016-05-28
Fix custom marker anchor points

### LINKED ISSUE #230: Android custom marker incorrect anchor point
Hello,

I have a map with custom marker points on it and they are no longer in the correct position on the map.

### Incorrect

Earlier in the week the position of the markers were all correct (1st image), then on Wednesday I noticed that the markers were in a different location and on zoom in and out they moved around weird, I went back to previous commits of my code but nothing resolved the issue. I believe this issue came up from a google play services update that occurred on Tuesday.

I can somewhat fix the problem by setting an anchor point on the marker, the problem with that is only android v4.4.2 and down are affected by the issue and the anchor points vary from device to device. (there is no problem on 4.4.4+ and iOS)

I believe what is happening is that the custom marker bitmap size is larger than the actual view. The view itself always looks correct (26x26) but the map is behaving like the image is double that size, so the anchor point is completely off.

Any help on this issue would be greatly appreciated,

code:
```js
class CustomMarker extends React.Component {
  render() {
    return (
      <View style={styles.markerWrapper}>
        <Text style={styles.markerText}>{this.props.markerNumber}</Text>
      </View>
    )
  }
}
//...
renderMarker(i) {
    let markerNumber = i+1;
    return (
      <MapView.Marker
        //...
        coordinate={{
          latitude: lat,
          longitude: lng,
        }}>
        <CustomMarker markerNumber={markerNumber} />
      </MapView.Marker>
    );
  }
//...
let styles = StyleSheet.create({
  markerWrapper: {
    borderRadius: 13,
    width: 26,
    height: 26,
    alignItems: "center",
    justifyContent: "center",
    backgroundColor: cssVar("brandColor"),
  },
  markerText: {
    color: "#FFF",
    fontWeight: "700",
  },
});
```

note: I have also tried a simple square view with just a background color and the problem still persists.

levic92:
fixes #230

### DIFF
--- a/android/lib/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
+++ b/android/lib/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
@@ -275,6 +275,18 @@ public class AirMapMarker extends AirMapFeature {
         }
 
         marker.setIcon(getIcon());
+        
+        if (anchorIsSet) {
+            marker.setAnchor(anchorX, anchorY);
+        } else {
+            marker.setAnchor(0.5f, 1.0f);
+        }
+        
+        if (calloutAnchorIsSet) {
+            marker.setInfoWindowAnchor(calloutAnchorX, calloutAnchorY);
+        } else {
+            marker.setInfoWindowAnchor(0.5f, 0);
+        }
     }
 
     public void update(int width, int height) {


android/lib/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
```diff
@@ -275,6 +275,18 @@ public void update() {
         }
 
         marker.setIcon(getIcon());
+        
+        if (anchorIsSet) {
+            marker.setAnchor(anchorX, anchorY);
+        } else {
+            marker.setAnchor(0.5f, 1.0f);
```
jrichardlai: Where does those value comes from?

android/lib/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
```diff
@@ -275,6 +275,18 @@ public void update() {
...
+        }
+        
+        if (calloutAnchorIsSet) {
+            marker.setInfoWindowAnchor(calloutAnchorX, calloutAnchorY);
+        } else {
+            marker.setInfoWindowAnchor(0.5f, 0);
```
jrichardlai: Here too?

jrichardlai:
Thanks for the PR, just double checking where the hardcoded values come from?

heydabop:
Not sure where the values come from, but this works great for me. 👍

levic92:
The hard coded values are to ensure the default behaviour is preserved when no anchor is set via the options.

0.5f, 1.0f is the default for the marker. It gets positioned on the middle bottom (in relation to the coordinates on the map)

0.5f, 0 is the default for the callout window. It get positions on the middle top (in relation to the coordinates on the map)

So the default is that both the marker and callout window are centered horizontally on the point on the map. The bottom of the marker is directly above the point, and the bottom of the callout window is directly above the marker.

jrichardlai:
Sound good to me 👍 @levic92, is there a simple way to test out the issue? If you can add it to the custom marker example that would be great!

mvaivre:
@levic92 @jrichardlai I've just tried the current master, and the anchor still seems off, at least when the Marker mounts. If I'm in debug mode with hot reload on and I change the anchor value, then it "corrects itself" to the right position. Weird.

Issue #95, opened 2016-02-24
Memory leak on Android


sodik82:
Hi there,

I mount and unmount `<MapView>` component (with single marker) quite often. And during my test on real device the app is killed after some time by OS. I have investigated it more and found out that the memory usage is really high (I can get to 1.5GB after 25-30 cycles).

```
Applications Memory Usage (kB):
Uptime: 199716698 Realtime: 1702619320

** MEMINFO in pid 10768 [com.nativevibe] **
                   Pss  Private  Private  Swapped     Heap     Heap     Heap
                 Total    Dirty    Clean    Dirty     Size    Alloc     Free
                ------   ------   ------   ------   ------   ------   ------
  Native Heap        0        0        0        0    60402    60402    20493
  Dalvik Heap      165      104        0    82120   100544    84443    16101
 Dalvik Other        0        0        0      332                           
        Stack       68       68        0     1008                           
    Other dev   367039   366772       36        0                           
     .so mmap      409       12      148     2188                           
    .apk mmap       30        0        0        0                           
    .dex mmap       41        0        8        0                           
    code mmap      409        0        0        0                           
   image mmap      546       12        4     1876                           
   Other mmap       13        0        0        8                           
     Graphics    33444    33444        0        0                           
           GL   515760   515760        0        0                           
      Unknown      305      296        0    67744                           
        TOTAL   918229   916468      196   155276   160946   144845    36594
```

I am on Samsung S4 - Android 5.0.1. Quite funny is that same apk on different phone does not lead to such catastrophic results - the memory usage is still increasing but not in a such radical way.

Official documentation of [GMap](https://developers.google.com/android/reference/com/google/android/gms/maps/MapView) claims that the Users of this class must forward all the life cycle methods from the Activity or Fragment.

So I did make some experiments and I have added following code to `AirMapManager`

```
    @Override
    public void onDropViewInstance(AirMapView view) {
        super.onDropViewInstance(view);
        view.onDestroy();
    }
```

and it helped to significantly reduce the memory overhead - however the results were not perfect. However I don't know if it is the "right" place to call such method since I don't know the internals of react-native or to verify that onDestroy is really called since the method is final in `AirMapView`

lelandrichardson:
Hi @sodik82 ,

Thanks for looking into this issue!  Do you think you could put up a PR of the change?

Also, cc'ing @mkonicek to see if he can confirm that this would be an appropriate place to put this

sodik82:
[...]

PR #264, opened 2016-05-31
Fix: Map not removing markers and other features causing memory leak

DennisMG:
Issue #95


jrichardlai:
Thanks for the PR @DennisMG taking a look now

jrichardlai:
Confirming that `feature.getFeature()` returns null when feature.removeFromMap(map) is called. The code looks correct.

Thank you! merging

lelandrichardson:
I'm a little late to comment as I'm on vacation, but just wanted to say thanks for this fix! Great work!

PR #270, opened 2016-06-04
[MapMarker] Return id and coordinates on marker press

christopherdro:
According to the docs `onPress` should return coordinates of the maker, similiar to `onSelect`. This was working fine for Android but was missing from iOS.

/cc @jrichardlai


jrichardlai:
It worked without commenting out `//RCT_EXPORT_VIEW_PROPERTY(identifier)` ?

symmetriccurve:
@christopherdro . I do not get this working. can you please thrown a Example

christopherdro:
That's strange that it's not working for you guys. Will take a look in a bit.

christopherdro:
Ok what I am noticing is that it only works with custom markers and not the default ios pin.

symmetriccurve:
Thanks. With with custom markers  it Works.

Issue #271, opened 2016-06-04
App crashing on sony experia with android 6


npomfret:
I'm getting this error (only on a sony experia):

```
java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.res.Configuration android.content.res.Resources.getConfiguration()' on a null object reference
                                                             at android.util.ResolutionOverride.<init>(ResolutionOverride.java:56)
                                                             at android.view.SurfaceView.init(SurfaceView.java:207)
                                                             at android.view.SurfaceView.<init>(SurfaceView.java:187)
                                                             at maps.X.x.<init>(Unknown Source)
                                                             at maps.X.v.<init>(Unknown Source)
                                                             at maps.E.u.<init>(Unknown Source)
                                                             at maps.E.u.a(Unknown Source)
                                                             at maps.ag.ad.a(Unknown Source)
                                                             at maps.ag.t.a(Unknown Source)
                                                             at maps.ag.Q.a(Unknown Source)
                                                             at vn.onTransact(:com.google.android.gms.DynamiteModulesB:66)
                                                             at android.os.Binder.transact(Binder.java:387)
                                                             at com.google.android.gms.maps.internal.IMapViewDelegate$zza$zza.onCreate(Unknown Source)
                                                             at com.google.android.gms.maps.MapView$zza.onCreate(Unknown Source)
                                                             at com.google.android.gms.dynamic.zza$3.zzb(Unknown Source)
                                                             at com.google.android.gms.dynamic.zza$1.zza(Unknown Source)
                                                             at com.google.android.gms.maps.MapView$zzb.zzzW(Unknown Source)
                                                             at com.google.android.gms.maps.MapView$zzb.zza(Unknown Source)
                                                             at com.google.android.gms.dynamic.zza.zza(Unknown Source)
                                                             at com.google.android.gms.dynamic.zza.onCreate(Unknown Source)
                                                             at com.google.android.gms.maps.MapView.onCreate(Unknown Source)
                                                             at com.AirMaps.AirMapView.<init>(AirMapView.java:71)
                                                             at com.AirMaps.AirMapManager.createViewInstance(AirMapManager.java:70)
                                                             at com.AirMaps.AirMapManager.createViewInstance(AirMapManager.java:27)
                                                             at com.facebook.react.uimanager.ViewManager.createView(ViewManager.java:44)
```

As are a few other people:

https://code.google.com/p/gmaps-api-issues/issues/detail?id=9071

[...]

PR #276, opened 2016-06-08
0.4 stable

locly:
Fix for#165 and #271 - Android 6.0 crashing on devices such as Xperia Z5 Compact, Z5 Bond and others. 

It passes the activity to the package, manager and view so that it can be used in place of the app context when initialising the MapView component.

The new code as also been tested on the following devices to ensure it continues to work as before:
Nexus 4 (Android 5), Nexus 5 (Android 6), Nexus 6P (Android 6)

It requires a change to an app's MainActivity.java to pass `this` when creating the AirPackage: `new AirPackage(this)` rather than `new AirPackage()` in the current version.

### DIFF
--- a/android/src/main/java/com/AirMaps/AirMapManager.java
+++ b/android/src/main/java/com/AirMaps/AirMapManager.java
@@ -1,6 +1,7 @@
 package com.AirMaps;
 
 import android.view.View;
+import android.app.Activity;
 
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.ReactContext;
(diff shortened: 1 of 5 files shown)


jrichardlai:
Thanks for finding the issue and fixing it! I think the example app will need to be updated too.

christopherdro:
@locly Nice! Yea, could you update the example to pass `this` when declaring the package. Thanks!

locly:
Yes - have updated the example to declare the package with `new AirPackage(this)`

jrichardlai:
@locly looks good! thanks for the PR merging

jrichardlai:
@locly just realized this was for the 0.4-stable branch. I think this should also work for master, any reason not too?
I can do it for master unless you want to tackle this.

PR #290, opened 2016-06-13
iOS: Fix a marker not being shown when using a custom image

maximeaubaret:
When using a custom image for the marker, and when the ```imageLoader``` sets the image before the ```reactSetFrame``` method is called on the marker, the size of the view is incorrectly set, resulting in an invisible marker.
This commit makes sure that, when an image is set before the ```reactSetFrame```, we transfer the size of the image to the marker.


jeveloper:
Thank you

maximeaubaret:
You're welcome :)

PR #313, opened 2016-06-22
Follow user location continuously, if specified

raubas:
In my opinion this should be the developers choice, without this prop it is unnecessary hard to develop a mapkit-like behaviour for ios where the map follows the users location.

<Button 
icon={like.the.arrow.on.ios.maps}
placement={top.left.or.anywhere.really.on.the.map}
onPress={this.setState(followuser: toggle true/false} />

<Map followuser={this.state.followuser}>

For Android you'll still need a custom camera controller, but that should be the developers choice how to handle this situation. The followuser is very smooth for iOS.!

### DIFF
--- a/ios/AirMaps/AIRMapManager.m
+++ b/ios/AirMaps/AIRMapManager.m
@@ -434,7 +434,7 @@ static int kDragCenterContext;
         [mapView setRegion:region animated:YES];
 
         // Move to user location only for the first time it loads up.
-        mapView.followUserLocation = NO;
+        // mapView.followUserLocation = NO;
     }
 }
 


jrichardlai:
Does setting `followsUserLocation` on the Mapview not working?

raubas:
It zooms to the user, but then gets set to No. (Stops following).

My PR lets you choose this and control it yourself.

VinceBT:
Is there a way I can access the native map from your component in order to call the method ?

yonahforst:
+1 for this. I thought the `followsUserLocation` name was a little confusing since it just zooms to the current location and doesn't actually follow changes.

patrickml:
+1 Just changed it locally and it fixed it

Issue #158, opened 2016-03-30
Prevent mapToolbar to show when selecting a marker


mvaivre:
Strange behaviour related to the selection of a marker: a map toolbar appears... after rotating the screen to landscape mode. Maybe a MapView prop to allow or not the mapToolbar to show would be nice!

![screenshot_20160330-103031](https://cloud.githubusercontent.com/assets/6715943/14136507/693105f6-f663-11e5-99a7-3f9678328cf6.png)

heydabop:
I actually have a patch for this, I just haven't taken the extra time to make it a prop.

```
diff --git a/android/src/main/java/com/AirMaps/AirMapView.java b/android/src/main/java/com/AirMaps/AirMapView.java
index f028f3b..b07f370 100644
--- a/android/src/main/java/com/AirMaps/AirMapView.java
+++ b/android/src/main/java/com/AirMaps/AirMapView.java
@@ -107,6 +107,7 @@ public class AirMapView
         this.map = map;
         this.map.setInfoWindowAdapter(this);
         this.map.setOnMarkerDragListener(this);
+               this.map.getUiSettings().setMapToolbarEnabled(false);
 
         manager.pushEvent(this, "onMapReady", new WritableNativeMap());
```

mvaivre:
@heydabop Ok nice! Someone to add this as a prop? I could try to do it, but I don't have any experience with Java / writing modules for RN.

christopherdro:
@heydabop Any chance of putting a PR together for this?

heydabop:
Sure, I'll look into making it a prop.

heydabop:
PR is up [#316](https://github.com/lelandrichardson/react-native-maps/pull/316)

PR #316, opened 2016-06-22
made android maps toolbar controllable by prop

heydabop:
Resolves [#158](https://github.com/lelandrichardson/react-native-maps/issues/158)

### DIFF
--- a/components/MapView.js
+++ b/components/MapView.js
@@ -132,6 +132,14 @@ var MapView = React.createClass({
     pitchEnabled: PropTypes.bool,
 
     /**
+     * `Android only`
+     * If `false` will hide 'Navigate' and 'Open in Maps' buttons on marker press
+     * Default value is `true`.
+     *
+     */
+    toolbarEnabled: PropTypes.bool,
+
+    /**
      * A Boolean indicating whether the map shows scale information.
      * Default value is `false`
      *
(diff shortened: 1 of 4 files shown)


components/MapView.js
```diff
@@ -132,6 +132,14 @@ var MapView = React.createClass({
     pitchEnabled: PropTypes.bool,
 
     /**
+     * `Android only`
```
christopherdro: nit: remove this line and add `@platform android` at the end of the prop description.

christopherdro:
Sweet! Outside of the small change it looks good to me.

heydabop:
@christopherdro Ah, not sure how I missed all the other parts doing that. Fixed it!

christopherdro:
@felipecsl I know you just did a release to maven before this PR. Is it easy to update the the build on maven to include this?

felipecsl:
@christopherdro unlike NPM, Maven artifacts are immutable. Once you upload it you can't modify it (I believe you cant even delete it). So no, we'd need a new release.

christopherdro:
@felipecsl Ok. Does it make sense to try and plan out a steady release schedule? Say every 2-3 weeks? If we release a new version that doesn't have any native code changes, would we still be required to release a new version to maven that matches?

felipecsl:
sounds good to me! I believe we should try to keep all versions in sync even if there are no changes. It would be pretty confusing if the npm package is at version 0.9.0 but the latest Android jar is 0.6.0

christopherdro:
Sounds like a plan!

PR #321, opened 2016-06-23
Pass through style even when map isn't ready

astreet:
We encountered an issue at a hackathon where props (and therefore styles) was not being passed through: this means that if the map isn't ready (`this.state.isReady` is false), it is invisible unless it was given a width and height by its parent. This is an issue because on genymotion, where maps aren't available, isReady never becomes true and we can never see the error message.

### DIFF
--- a/components/MapView.js
+++ b/components/MapView.js
@@ -437,6 +437,7 @@ var MapView = React.createClass({
       props.handlePanDrag = !!props.onPanDrag;
     } else {
       props = {
+        ...this.props,
         region: null,
         initialRegion: null,
         onChange: this._onChange,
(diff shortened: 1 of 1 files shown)


jrichardlai:
Thank you for the PR I will double check tomorrow

jrichardlai:
Sorry tested the pr it breaks the example app Tracking Position example, and other ones too.

lelandrichardson:
@astreet I am intentionally not passing the other props in though, because some of them need to be "set" for the first time when the map is actually ready. i think the fix is to pass style through explicitly, but none of the others. Thoughts?

astreet:
Sure, I'm not sure what the other props we'd be grabbing here are so you'd probably know better than me :) Style is all I needed to fix my issue.

PR #323, opened 2016-06-24
fixed a bug where map completely failed to render in some situations

kulicuu:
Hello there, and thankyou for this wonderful module !

I'm pretty new to React-Native, and was at a hackathon yesterday at Facebook Tel Aviv.  There was a problem where the map component failed to render in some situations.  Some guys from Facebook London were there and they found this fix.  Maybe you'll accept it;  it's just one line in the `else` block of MapView's `render`, adding `...this.props` to pass through.

Cheers!

### DIFF
--- a/components/MapView.js
+++ b/components/MapView.js
@@ -437,6 +437,7 @@ var MapView = React.createClass({
       props.handlePanDrag = !!props.onPanDrag;
     } else {
       props = {
+        ...this.props,
         region: null,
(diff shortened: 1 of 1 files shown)


kulicuu:
Ah I see #321 astreet already got to it.

jrichardlai:
Thanks for the PR, I'll double check the other PR I will close this one

PR #325, opened 2016-06-24
[AirMapMarker][android] fix custom marker combined with MapMarker ima…

gilbox:
…ge prop

previously, if you had a custom marker, the `MapMarker`'s `image` prop was ignored on android. 

image


gilbox:
ping @felipecsl @lelandrichardson

felipecsl:
I don't quite understand what's going on here, but looks 👍 . @lelandrichardson might have more context

PR #333, opened 2016-06-27
Fix warning for iOS "loadImageWithTag is deprecated" for RN 0.28.0

joelcloralt:
There is a warning after upgrading React Native to 0.28.0 about `loadImageWithTag` being deprecated. I saw it when loading a custom image for a marker. I updated the function to what I think is the new way of loading these images. 

I'm not sure if `clipped` should be `YES` by default but I think that's the case now.

Hope this helps :)

### DIFF
--- a/ios/AirMaps/AIRMapMarker.m
+++ b/ios/AirMaps/AIRMapMarker.m
@@ -209,9 +209,10 @@
         _reloadImageCancellationBlock();
         _reloadImageCancellationBlock = nil;
     }
-    _reloadImageCancellationBlock = [_bridge.imageLoader loadImageWithTag:_imageSrc
+    _reloadImageCancellationBlock = [_bridge.imageLoader loadImageWithTag:[RCTConvert NSURLRequest:_imageSrc]
                                                                      size:self.bounds.size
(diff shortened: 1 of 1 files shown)


jrichardlai:
Looks good, just double checked too https://github.com/facebook/react-native/blob/235749ba196adb157df256a76589aad5a8281c1c/Libraries/Image/RCTImageLoader.m#L749 seems to be the default

jrichardlai:
@joelcloralt sorry I misunderstood the story, I thought loadImageWithTag was defined in previous versions of RN, but it seems not. I'm gonna revert it

joelcloralt:
:( ok.

This change should still be valid for RN 0.28+ right?

farwayer:
Build failed with this commit on 0.28. It will be nice to revert asap.

jrichardlai:
Reverted @farwayer

ddresch:
Also fails with RN 0.30.0. Any solutions for now, cause the warning is also present with 0.30.0?

ddresch:
Sorry ignore my last comment found this PR which fixes this issue.

https://github.com/lelandrichardson/react-native-maps/pull/389/files

PR #335, opened 2016-06-28
[android] installation instructions to actually use node_modules

igrayson:
* And basic gradle.properties reorg to allow that

Previous `installation.md` was configuring android projects to pull their `react-native-maps` artifact from maven, and not `node_modules`. It was also uselessly [adding a settings.gradle](https://github.com/lelandrichardson/react-native-maps/blob/master/docs/installation.md#android) project for `:react-native-maps` which was never used.

This change fixes those instructions, and related project structure, to allow us to build android projects with `react-native-maps` coming from `node_modules`, just as ios projects are built.

I'm not informed about our reasoning for publishing to maven at all, but I don't believe this PR breaks that functionality.

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -13,17 +13,3 @@
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
 #Thu May 12 17:26:12 PDT 2016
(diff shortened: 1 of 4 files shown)


jrichardlai:
@felipecsl does this PR looks good to you?

felipecsl:
yeah that should work as well 👍

Issue #64, opened 2016-02-02
Vertical scroll not working when included in "ScrollView" - Android


dchdimitrov:
MapView for iOS is with greater priority than ScrollView. When you try to scroll, you scroll the map. If the content of the ScrollView should be scrolled, the user has to touch something outside the map. I think that the case should be the same here.

ericmuyser:
It doesn't seem to be working in iOS either. I'd like to wrap it in a ScrollView so that I can scroll below the map. I think the map is absolutely positioned so that's why. I think a workaround would be to use an interaction to manipulate the map properties manually.

loox-:
I met the same problem on Android: it is nearly impossible to scroll map in the same direction of scrollview (either vertical or horizontal)
Did anyone solve this?

ericmuyser:
I'm trying to grok the issue but I believe I solved this by wrapping everything inside the ScrollView with a View, including the MapView.

```js
<ScrollView>
  <View>
    <MapView></MapView>
  </View>
  <View>
    <MyComponentHere></MyComponentHere>
  </View>
</ScrollView>
```

AndroidMatrixnmedia:
good# 

> [`_****_`](url)

mohamedpsx:
On ios everything works perfectly but this issue still persists on android ? any progress on this or any ideas that might help ?

rborn:
We're trying to cleanup the repo, so please comment/mention me if this is still an issue. Thank you.

jpls93:
I'm still experiencing this issue. Only on Android. Works perfectly fine in iOS.

```
"react-native": "0.49.3",
```

diksha-valuecoders:
I'm still experiencing the this issue. I am using ScrollView vertical within the viewTransformer and the ScrollView horizontal. The parent scroll works properly as expected but the vertical scroll which is inside the view transformer is not working properly. 
Is there any solution for a table of large no. of data which scroll horizontal as well as vertical with zooming.

PR #343, opened 2016-06-29
Fixing scrolling map inside a scrollView

mohamedpsx:
@lelandrichardson && @jrichardlai this is the same issue mentioned in #64 , it turns out that this is a known bug in MapView i just implemented the fix suggested on [stackoverflow]: http://stackoverflow.com/questions/6546108/mapview-inside-a-scrollview

Already tested this solution on another project and it works perfectly!


mohamedpsx:
@felipecsl sorry for tagging most of the contributors, but there is plenty of good PR's not getting merged. you guys should be more active as maintainers

lelandrichardson:
LGTM. Thanks!

lelandrichardson:
And sorry for the slow turnarounds. Hopefully we can give this project more of the attention it deserves soon.

PR #368, opened 2016-07-08
More RN 29 compat changes

felipecsl:
@lelandrichardson 

Still getting this error though:

screen shot 2016-07-08 at 4 48 08 pm


android/lib/src/main/java/com/airbnb/android/react/maps/AirMapManager.java
```diff
@@ -40,24 +39,10 @@
     );
 
     private ReactContext reactContext;
+    private final Context appContext;
 
-    private Context appContext;
-    private AirMapMarkerManager markerManager;
-    private AirMapPolylineManager polylineManager;
-    private AirMapPolygonManager polygonManager;
-    private AirMapCircleManager circleManager;
```
felipecsl: none of these were being used

android/lib/build.gradle
```diff
@@ -15,7 +15,7 @@ android {
 }
 
 dependencies {
-  compile 'com.facebook.react:react-native:+'
+  compile 'com.facebook.react:react-native:0.29.0'
   compile "com.google.android.gms:play-services-base:8.4.0"
```
christopherdro: Is it possible to eliminate react native dependency completely?

lelandrichardson: Nothing would build if the dependency isn't there. RN APIs are used all over the place in this project

android/lib/src/main/java/com/airbnb/android/react/maps/AirMapManager.java
lelandrichardson: 👍  I put them there a long time ago, as there was some work that I was doing that required some cross-talk between them, but I don't think that code ever got committed.

android/lib/build.gradle
christopherdro: aren't those APIs available through the main app ?

lelandrichardson: You can't know for sure...  I believe you are more or less asking for something similar to "Peer Dependencies" in NPM. I don't believe there is anything like that for gradle.

According to @felipecsl, If we specify `react-native:0.29.0` here, if another project uses `react-native:0.30` or something, gradle will only choose to use `0.30`. Perhaps Felipe can confirm.

PR #381, opened 2016-07-11
Fix setup of example app for Android

pedro:
Fix maven repo location and just use whatever version of react-native is under `node_modules`.

This matches what react-native itself does, eg:

https://github.com/facebook/react-native/blob/9b184cc0f43194f8db73d2fea0b23b1284c487f9/local-cli/generator-android/templates/src/app/build.gradle#L131

### DIFF
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -100,6 +100,6 @@ android {
 
 dependencies {
     compile "com.android.support:appcompat-v7:23.4.0"
-    compile "com.facebook.react:react-native:0.29.0"  // From node_modules
+    compile "com.facebook.react:react-native:+"  // From node_modules
     compile project(':react-native-maps')
 }
--- a/example/android/build.gradle
+++ b/example/android/build.gradle
@@ -18,7 +18,7 @@ allprojects {
         jcenter()
         maven {
             // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
-            url "$projectDir/../../node_modules/react-native/android"
+            url "$rootDir/../node_modules/react-native/android"
         }
     }
 }


example/android/app/build.gradle
```diff
@@ -100,6 +100,6 @@ android {
 
 dependencies {
     compile "com.android.support:appcompat-v7:23.4.0"
-    compile "com.facebook.react:react-native:0.29.0"  // From node_modules
+    compile "com.facebook.react:react-native:+"  // From node_modules
```
felipecsl: I don't understand why people continuously want to do this... Using `+` is **not** okay

pedro: hmm it's what react-native itself does.

This is just for the example app, but I'd love to see if there's a better way to define this dependency for the package itself as react-native moves so fast.

felipecsl: I think we can probably omit this line completely, since gradle should pick up the `react-native` dependency transitively from `react-native-maps`.

felipecsl: Also, even if you specify an explicit version like `0.29.0`, gradle will automatically switch to a newer version if any of your other dependencies depends on a newer version of `react-native`

pedro: This is obviously up to you folks, but I'd make the example app match whatever `react-native init` does, to ensure this works regular apps out of the box.

Good point on the dependency! So if react-native-maps is on 0.29 my app can be on that version or any newer? Right now I'm still on 0.28, so that's why I initially thought the `+` would be nice.

felipecsl: Correct. [Here's](https://docs.gradle.org/current/userguide/dependency_management.html) the documentation on how it works:

```
Newest: The newest version of the dependency is used. This is Gradle's default strategy, and is often an appropriate choice as long as versions are backwards-compatible.
```

React Native in uses weird and non-standard ways of doing some stuff, this is one of them. We'll try to have them change their minds on this :)

pedro: aaaaaaah, this makes perfect sense. You rock, thank you for the patience! :}

felipecsl: you're welcome!

PR #382, opened 2016-07-11
Take prop to hide the "my location" button in Android

pedro:
This button is enabled by default when showsUserLocation is true; it's the gray button that shows on the top right and centers on the user location on press.

This is useful if you want to implement a different version of that in React Native, for instance.


felipecsl:
LGTM

VinceBT:
Can we have a release to include this feature soon ?

kkaczmarczyk:
I would appreciate release as well. Is there any concrete date for the release including this fix yet?

PR #386, opened 2016-07-12
Fit to supplied markers

AidenMontgomery:
This is a cleaned up version of an earlier pull request, which is based on a more recent fork of the master branch.


docs/mapview.md
```diff
@@ -53,6 +53,7 @@
 | `animateToRegion` | `region: Region`, `duration: Number` |
 | `animateToCoordinate` | `region: Coordinate`, `duration: Number` |
 | `fitToElements` | `animated: Boolean` |
+| `fitToSpecifiedMarkers` | `markerIDs: String[]` |
```
christopherdro: `fitToSuppliedMarkers`

christopherdro:
@AidenMontgomery Thanks for redoing the PR!

Looks good to me but I haven't gotten a chance to test it.

/cc @jrichardlai @lelandrichardson @felipecsl 

I think we should get this merged in.

android/lib/src/main/java/com/airbnb/android/react/maps/AirMapView.java
```diff
@@ -506,6 +508,39 @@ public void fitToElements(boolean animated) {
...
+            markerIDs[i] = markerIDsArray.getString(i);
+        }
+
+        boolean addedPosition = false;
+
+        for (AirMapFeature feature : features) {
+            if (feature instanceof AirMapMarker) {
+                String identifier = ((AirMapMarker)feature).getIdentifier();
+                Marker marker = (Marker)feature.getFeature();
+                if (Arrays.asList(markerIDs).contains(identifier)) {
```
felipecsl: extract `Arrays.asList(markerIDs)` as a variable outside of the loop so you dont allocate a new array every time

felipecsl:
looks good to me with minor nits 👍

AidenMontgomery:
Thanks for the feedback. I will make the changes and do some more testing this evening, then update the PR.

AidenMontgomery:
I haven't testing this on Android yet, but I have run the example project on the iOS Simulator and I think that it looks ok. I am going to try and get it running on the Android emulator now.

christopherdro:
@AidenMontgomery I should have some time today to test these on a few android devices.

christopherdro:
Remind me again why you wouldn't use fitToElelments and instead of having to specify an array of marker ids?

AidenMontgomery:
Unless it has changed recently, fitToElements shows every marker on the map. In my use case I needed to zoom to the closest 2 markers on the map, relative to another marker. However, as soon as the user zooms out they need to be able to see all of the markers in the area. I could have done this by removing all other markers, then reacting to the region changed gesture and adding all of the other markers back on the map, but this felt simpler to me.

erickreutz:
Would love to be able to use this!

christopherdro:
@felipecsl I think your comments were addressed. Should we get this merged in?

felipecsl:
yep, looks good to me!

PR #388, opened 2016-07-12
Update gradle dependencies

magrinj:
I just change all the dependencies version in `build.gradle` to avoid compilation error with multiple library using multiple versions

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -16,7 +16,7 @@ android {
 }
 
 dependencies {
-    compile 'com.facebook.react:react-native:0.19.+'
-    compile "com.google.android.gms:play-services-base:8.4.0"
-    compile 'com.google.android.gms:play-services-maps:8.4.0'
+    compile 'com.facebook.react:react-native:+'
+    compile "com.google.android.gms:play-services-base:+"
+    compile 'com.google.android.gms:play-services-maps:+'
 }


christopherdro:
@magrinj We keep going back and forth on this issue. @felipecsl Says a version has to be specified in order to publish to maven. 

/cc @lelandrichardson

felipecsl:
@magrinj thanks for the contribution, but this has been discussed multiple times already in other PRs

christopherdro:
@felipecsl I actually just opened a thread on the react native core contributor group on facebook trying to figure out the best way to handle peer dep version of react native. 

The reality of the situation is that this causes conflicts with other libraries so I'm trying to figure out the best way to find a solution.

christopherdro:
Is it the warning thats of concern? I remember reading that it was not possible to publish to maven if you used '+'

felipecsl:
@christopherdro yes, that's one problem. It's also unsafe and not recommended, since you don't really know what version of RN will be used.
Here's more info on the matter: http://blog.danlew.net/2015/09/09/dont-use-dynamic-versions-for-your-dependencies/

magrinj:
Ok I understand, but It's still a big problem when you use external libraries and these libraries use another version of play-services.
Can we apply the `+` only on this two packages, or maven still being upset by that ?

```
compile "com.google.android.gms:play-services-base:+"
compile 'com.google.android.gms:play-services-maps:+'
```

felipecsl:
That's only a problem if you need to use a version older than the one that the project is set to (currently 8.4). Is that the case?

magrinj:
My problem is that, I've two library, one using version 9 (Firebase) and react-native-maps using version 8.4, my app currently compile, but last week when I got a compilation error and I replace this two lines my app compile again. I think gradle ignore the older version of play-services and choose the last one, and that's conflict with the library using the older one, but I'm not sure of that.

felipecsl:
Yes, gradle will ignore the older one and use the latest in that case

magrinj:
Ok so, I think that's why I got the compilation error, I can't reproduce it right now, but when I came across it I fix it by replacing the versioning of the play-services in `react-native-maps` package

PR #389, opened 2016-07-12
Update AirMapMarker to use loadImageWithURLRequest

paulshen:
[`loadImageWithTag`](https://github.com/facebook/react-native/blob/ba4c34c0db004cda9fb561b3bdf5224d518ab32d/Libraries/Image/RCTImageLoader.m#L755) is deprecated and shows warnings on RN 0.29.

### Test Plan

On iOS, render a map with a marker that has image prop set. See marker is rendered correctly and no deprecation warnings.

```
<MapView ...>
  <MapView.Marker ... image={require('./pin.png')} />
</MapView>
```

### DIFF
--- a/ios/AirMaps/AIRMapMarker.m
+++ b/ios/AirMaps/AIRMapMarker.m
@@ -209,20 +209,21 @@
         _reloadImageCancellationBlock();
         _reloadImageCancellationBlock = nil;
     }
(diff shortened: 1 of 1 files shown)


christopherdro:
@paulshen Nice. Thanks!

PR #390, opened 2016-07-12
Update the Android installation instructions

felipecsl:
This seems to be causing a lot of confusion

@lelandrichardson @christopherdro

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -40,6 +40,8 @@ To install using Cocoapods, simply insert the following line into your `Podfile`
 
 ## Android
 
+For React Native v0.29.0 or above:
+
 1. in your `build.gradle` add:
  ```
  ...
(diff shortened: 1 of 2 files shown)


christopherdro:
We should still leave the old setup and label the new one with `For React Native >= v0.29`. Similiar to how code push has their docs setup. 
https://github.com/Microsoft/react-native-code-push#plugin-configuration-android

felipecsl:
Fair enough. I added the old instructions back

christopherdro:
Maybe its worth it to also specify what version supports RN 29 somewhere in the docs?

felipecsl:
Yes, that's definitely a good idea. I think it should go in the `CHANGELOG.md` with the release notes

docs/installation.md
```diff
@@ -40,81 +40,71 @@ To install using Cocoapods, simply insert the following line into your `Podfile`
...
+1. in your `build.gradle` add:
+```groovy
+...
+dependencies {
+  ...
+  compile 'com.airbnb.android:react-native-maps:0.7.1'
+}
+```
+
+3. in your application object, add:
```
ahmed1490: number 2 is missing

PR #395, opened 2016-07-15
Lower dependency on GMS packages

pedro:
I can't seem to run 8.4 on older devices (trying a galaxy s3 mini).

7.8 works fine on that device and seems to cover the functionality we need, though I didn't verify every single feature here.

### DIFF
--- a/android/lib/build.gradle
+++ b/android/lib/build.gradle
@@ -17,6 +17,6 @@ android {
 
 dependencies {
   compile 'com.facebook.react:react-native:0.29.0'
-  compile "com.google.android.gms:play-services-base:8.4.0"
-  compile 'com.google.android.gms:play-services-maps:8.4.0'
+  compile "com.google.android.gms:play-services-base:7.8.0"
+  compile 'com.google.android.gms:play-services-maps:7.8.0'
 }


felipecsl:
what issue are you having? do you have a stacktrace or more details on the problem?

pedro:
The app compiles but just doesn't render the map.

Tried changing the styles but nothing – only rendered when I lowered this version dependency. I actually knew where to look because we had this issue before with another Map component we wrote early on for Android.

I don't see anything useful in the device logs, looking for gsm|google|maps :{ let me know if there's anything in particular I can provide/look for to help!

felipecsl:
I'm very hesitant to revert to a Google Play version that's even older than the current one. 7.8 is 1 year old already. I suspect we're doing something wrong or using some deprecated API that's currently broken. I'll take a look into it

pedro:
For sure, up to you folks of course!

Just sharing this patch in case it helps others, I'll close for now then.

I take no one else ran into issues with older phones?

felipecsl:
That's my take, maybe others are not concerned about it as much :)
Thanks for the contribution anyways!
Do you think it's a device issue instead of Android API level issue? I assume it something that doesn't work with older API levels. It could also be something specific to a device model, but that's much less likely

PR #411, opened 2016-07-21
Supports React Native 0.30 on both platforms again

skellock:
Here's a few things that will make `react-native-maps` work again now that React Native 0.30 is here.

Before you read to code, i just want ensure I'm not trolling you here. I adore `react-native-maps` and just want it to work out of the box and survive upgrades of React Native.

### Android

* Set the dependency to for React Native to `+`.    I know y'all have issues with this, but I believe this is straight forward and works.  <3
* Downgrade play services back to 8.4.0 as this will cause android clients to need to upgrade to work.  Just a compatibility thing.

### iOS

* Two small header imports were broken.

### Package.json

* Set the `rnpm` hint in `package.json` to point it to the right code so it gets linked properly.
* Changed the dependency to peerDependency.  Just relies on the host to install the right version.  This change is similar in spirit to the old `*` you just replaced.

### How To Test

`npm i --save skellock/react-native-maps#friends-with-30` on a fresh `react-native init` project if you'd like to test in advance.

Just to reinterate.  These changes are just to play well with RN upgrades and make this awesome project work out of the box again.

### Thanks!

Cheers and thanks for all your hard work on this project.

### DIFF
--- a/android/lib/build.gradle
+++ b/android/lib/build.gradle
@@ -17,6 +17,6 @@ android {
 
 dependencies {
   compile 'com.facebook.react:react-native:+'
-  compile "com.google.android.gms:play-services-base:9.2.0"
-  compile 'com.google.android.gms:play-services-maps:9.2.0'
+  compile "com.google.android.gms:play-services-base:8.4.0"
+  compile 'com.google.android.gms:play-services-maps:8.4.0'
 }
--- a/android/lib/build.gradle
+++ b/android/lib/build.gradle
@@ -16,7 +16,7 @@ android {
 }
 
 dependencies {
-  compile 'com.facebook.react:react-native:0.29.2'
+  compile 'com.facebook.react:react-native:+'
   compile "com.google.android.gms:play-services-base:9.2.0"
   compile 'com.google.android.gms:play-services-maps:9.2.0'
 }
--- a/ios/AirMaps/AIRMap.h
+++ b/ios/AirMaps/AIRMap.h
@@ -9,7 +9,6 @@
 
 #import <MapKit/MapKit.h>
 #import <UIKit/UIKit.h>
-#import <React/RCTComponent.h>
 
 #import "RCTConvert+MapKit.h"
 #import "RCTComponent.h"
--- a/ios/AirMaps/AIRMapCallout.h
+++ b/ios/AirMaps/AIRMapCallout.h
@@ -4,7 +4,7 @@
 //
 
 #import <Foundation/Foundation.h>
-#import "React/RCTView.h"
+#import "RCTView.h"
 
 
 @interface AIRMapCallout : RCTView
--- a/package.json
+++ b/package.json
@@ -20,9 +20,9 @@
     "google-maps",
     "mapkit"
   ],
-  "dependencies": {
-    "react": "15.2.0",
-    "react-native": "^0.29.2"
+  "peerDependencies": {
+    "react": ">=15.2.0",
+    "react-native": ">=0.30.0"
   },
   "devDependencies": {
     "react": "15.2.0",
--- a/package.json
+++ b/package.json
@@ -27,5 +27,10 @@
   "devDependencies": {
     "react": "15.2.0",
     "react-native": "^0.29.2"
+  },
+  "rnpm": {
+    "android": {
+      "sourceDir": "./android/lib"
+    }
   }
 }


android/lib/build.gradle
```diff
@@ -16,7 +16,7 @@ android {
 }
 
 dependencies {
-  compile 'com.facebook.react:react-native:0.29.2'
-  compile "com.google.android.gms:play-services-base:9.2.0"
-  compile 'com.google.android.gms:play-services-maps:9.2.0'
+  compile 'com.facebook.react:react-native:+'
+  compile "com.google.android.gms:play-services-base:8.4.0"
+  compile 'com.google.android.gms:play-services-maps:8.4.0'
```
felipecsl: What's your rationale behind this change?

skellock: In my Android sim (Android 6 (23) - avd + Google APIs on Nexus 5X) maps won't show.  `logcat` said that Play Services was out of date.  This will require updates on devices.

Also, (just fyi, not really an issue), it does require us to update our dev Google Play services to version 31 (in Android SDK Manager), but that's not a huge deal, re-running that as a dev isn't a big problem and we can easily put that in a readme file.

I figure, once we start using 9.x features, it's time to set the dependency?  Thinking outloud.

felipecsl: We've had other problems with 8.4 recently and it is a pretty old version already.
When an app requires a version of GPS that's newer than the one that is currently installed, it should automatically update in the background, so that shouldn't be a problem. I'd rather fix the issue with 9.x instead of going back to 8.4 that's 1+ year old.
As for the react-native version, there was plenty of discussion about this in other PRs already and the tl;dr is to not use the wildcard dependency. Gradle will automatically pick the latest available version if you depend on a newer one.

christopherdro: @felipecsl At this point I'm starting to think what you guys are doing different internally that seems to work. As you can see, even after I brought this issue up many times people are still finding it problematic.

As far as play services is concerned. We've had a few concerns with new phones that have an outdated version of play services. In order for the phone the auto update it, the consumer needs to login with their google account. What was the rationale to bumping the versions of play services up? Was there a change that required this?

At this point, I just work off my own fork of the project.

felipecsl: Ok that's pretty bad. Yeah I don't quite get why this versioning is so problematic to people but I guess at this point it might be better to follow the RN standard then to avoid any more churn.
for GPS we were seeing this crash below a lot and it seems to be a bug on 8.4.0 with Marshmallow devices. Upgrading to 9.2.1 fixed it:

```
java.lang.IllegalArgumentException: Can only use lower 16 bits for requestCode
        at android.support.v4.app.BaseFragmentActivityEclair.checkForValidRequestCode(BaseFragmentActivityEclair.java:64)
        at android.support.v4.app.BaseFragmentActivityEclair.startIntentSenderForResult(BaseFragmentActivityEclair.java:45)
        at android.support.v4.app.FragmentActivity.startIntentSenderForResult(FragmentActivity.java:75)
        at com.google.android.gms.common.ConnectionResult.startResolutionForResult(Unknown:-1)
        at com.google.android.gms.common.api.internal.zzw$zzb.run(Unknown:-1)
        at android.os.Handler.handleCallback(Handler.java:739)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:158)
        at android.app.ActivityThread.main(ActivityThread.java:7224)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120)
```

skellock: I only have a problem with the version hard coding because it means `react-native-maps` is broken out of the box 3-5 days every 2 weeks.  In spirit, i agree with your take on "know thy dependencies".

Something I've always looked up to about this project was your README file that said, "Look, RN is new and unstable.  Stay up to date.  Here: have a * in our package.json". 

I just hate to see you guys doing grunt work when new releases come out.  :)

RE the 9.2 :  If you think that's the right thing to do, I guess we'll just have to have people upgrade.  I don't know of the bugs you're talking about, but you have a bazillion times more exposure to that than I do.  :)

Thanks again for considering these changes.

christopherdro: As far as outdated play services is concerned. I think I checked and I couldn't find how you can detect what play services version the phone is currently on.

If this is possible, I would suggest we do a check against whatever version this maps library requires. If the user has an outdated version we emit an event over the bridge so a proper alert/action can be given to the user.

felipecsl: ok at this point I'll just defer to you @christopherdro and @lelandrichardson on this change, I consider my cause defeated 😄

skellock: I'm happy to slice & dice this PR up into constituents.  

I'm also down with you guys just making the changes yourself.  I don't need credit, just a healthy & happy `react-native-maps`.  Lemme know if I can help.

felipecsl: thanks for the positive attitude, @skellock. And thanks for your contribution

bodyflex:
is there going to be a release with this change?

yeegr:
On Max OSX, just npm install the latest package, still getting the same error

`** BUILD FAILED **


The following build commands failed:
	CompileC /Users/yeegr/Desktop/t2/ios/build/Build/Intermediates/AirMaps.build/Debug-iphonesimulator/AirMaps.build/Objects-normal/x86_64/AIRMapCallout.o AirMaps/AIRMapCallout.m normal x86_64 objective-c com.apple.compilers.llvm.clang.1_0.compiler
(1 failure)
/Users/yeegr/Desktop/t2/node_modules/promise/lib/done.js:10
      throw err;
      ^

Error: xcodebuild process exited with code 65
    at ChildProcess.<anonymous> (runIOS.js:102:14)
    at emitTwo (events.js:100:13)
    at ChildProcess.emit (events.js:185:7)
    at maybeClose (internal/child_process.js:850:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:215:5)`

bnetter:
Same, still have the error code 65 even with this.

skellock:
Not sure if its been released just yet.  Probably soon though.

You can verify by `npm install --save lelandrichardson/react-native-maps`.  That'll get you the latest.  

You might be in an awkward state though due to the previous release.   I'm not sure if `react-native unlink react-native-maps` has been promoted from `rnpm` just yet.

So...

A little bit of cleanup in settings, build.gradle, and MainApplication.java and MainActivity.java might be required.  To see what that they end up looking like, try in a fresh project.

bnetter:
I'm having the error 65 on a fresh project:
+ Running `react-native init my-map-app`
+ Then adding your fork into the `package.json` by running `npm i --save skellock/react-native-maps#friends-with-30`
+ Then `rpnm link`
+ Then on the `ios` folder doing a `pod init`
+ Adding `pod 'react-native-maps', :path => '../node_modules/react-native-maps'` to the Podfile, and uncommenting `platform :ios, '9.0'`
+ Then running `pod install`
+ Then running `react-native run-ios`

And the error 65 appears. Not sure what went wrong. I have the latest `react-native-cli` release.

skellock:
What happens if you skip the `pod` steps?

bnetter:
Indeed, it works better !

PR #434, opened 2016-07-27
Added new method: setMapType

tpisto:
This method allows fast and efficient way to change the underlying map type. Supported map types are  'standard', 'satellite', 'hybrid', 'terrain'

### DIFF
--- a/android/lib/build.gradle
+++ b/android/lib/build.gradle
@@ -16,7 +16,7 @@ android {
 }
 
 dependencies {
-  compile 'com.facebook.react:react-native:0.29.2'
+  compile 'com.facebook.react:react-native:+'
   compile "com.google.android.gms:play-services-base:9.2.0"
   compile 'com.google.android.gms:play-services-maps:9.2.0'
 }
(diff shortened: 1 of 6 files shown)


spikebrehm:
Thanks for the PR @tpisto!

I'm curious, what's the use case for an imperative `setMapType()` method?  Is the `mapType` prop not sufficient for changing the map type?

jeveloper:
I'm confused, but this is already present and works great on both platforms. I can attest to that.

tpisto:
@spikebrehm Thank you for the comments. I found the imperative method very convenient and fast if you have following structure (taken from the front page readme.md) and big number of markers:

```
<MapView
  region={this.state.region}
  onRegionChange={this.onRegionChange}
>
  {this.state.markers.map(marker => (
    <MapView.Marker
      coordinate={marker.latlng}
      title={marker.title}
      description={marker.description}
    >
    <MapView.Callout>
        <MyCustomCalloutView {...marker} />
     </MapView.Callout>
   </MapView.Marker>
  ))}
</MapView>
```

There are ways to overcome this performance problem in other ways too, but for me this method offered usage convenience. That could be just me - but I thought maybe others could find this useful too. But if I'm wrong, it's ok just close this PR.

spikebrehm:
I don't see where `setMapType()` is called in your example.  Can you provide some sample code of how you're using it, so I can understand the performance implications?

Here's what it would look like to use the `mapType` prop:

```js
<MapView
  region={this.state.region}
  onRegionChange={this.onRegionChange}
  mapType={this.state.mapType}
>
  {this.state.markers.map(marker => (
    <MapView.Marker
      coordinate={marker.latlng}
      title={marker.title}
      description={marker.description}
    >
    <MapView.Callout>
        <MyCustomCalloutView {...marker} />
     </MapView.Callout>
   </MapView.Marker>
  ))}
</MapView>
```

spikebrehm:
@tpisto closing for now.  Feel free to re-open if you think this is necessary.

PR #439, opened 2016-07-29
Include import line in Java application

gohanman:
Adding `MapsPackage` to the list doesn't work if the import statement hasn't been added to the file.


NomadGraphix:
Using the documentation was not enough to figure this out. I had to StackOverflow it.

spikebrehm:
Thanks 👍

PR #445, opened 2016-08-05
Improvements to watch and copy script

spikebrehm:
* Add proper `devDependencies` to `package.json`
* Also watch and copy ios, android source code

to: @lelandrichardson


lelandrichardson:
Seems legit.

PR #450, opened 2016-08-09
Update RegionChangeEvent to work with RN 0.32.0-rc

yianyue:
May not be backward compatible.

For RN 0.32.0rc, constructor for Event was changed to accept only 1 parameter
See: https://github.com/facebook/react-native/commit/da063e3d55a4fcd617ac496d62aa051343f35b56


christopherdro:
What happens if you pass only 1 parameter on older versions of RN?

waqas19921:
Update required to get working with new react-native version 0.32.0
You can suggest users to use old version for react-native old version users

spikebrehm:
Fixed by #502.

PR #485, opened 2016-08-23
[CRASH] Added check on marker class in predicate

johnryan:
The current implementation will crash if you call fitToSuppliedMarkers while showsUserLocation is set to true because of an unchecked cast of the MKUserLocation marker.


spikebrehm:
Thanks! 🍻

PR #493, opened 2016-08-25
absoluteFillObject

stereodenis:
https://github.com/facebook/react-native/blob/8eed600d5b2be78c405f7ef139a7b0e4a2480c85/Libraries/StyleSheet/StyleSheet.js#L108


spikebrehm:
Thanks! 🍻

PR #501, opened 2016-08-25
Add ESLint and fix a number of linting violations

spikebrehm:
Add ESLint using the `eslint-config-airbnb` preset, and override a few of the rules to better fit with React Native (`prefer-object-spread`, `react/jsx-filename-extension`) or to suppress a number of the warnings until we can get everything cleaned up (`no-use-before-define`, `no-underscore-dangle`).

Add `npm run lint` command, which runs ESLint.  Currently, there are still 488 linting errors, which I'd like to fix in follow-up PRs to prevent a bunch of merge conflicts from popping up.

**Note**: This doesn't actually enforce linting to pass in CI before merging. That will happen once we address all the issues.

Please take a look @lelandrichardson @jrichardlai.

cc: @ljharb


.eslintignore
```diff
@@ -0,0 +1 @@
+node_modules
```
ljharb: this should already be gitignored, and as such, eslint should already ignore it - why is this needed?

components/MapMarker.js
```diff
@@ -1,24 +1,19 @@
...
   StyleSheet,
   Platform,
   NativeModules,
   Animated,
-} = ReactNative;
+  findNodeHandle,
+} from 'react-native';
 
-var resolveAssetSource = require('react-native/Libraries/Image/resolveAssetSource');
+const resolveAssetSource = require('react-native/Libraries/Image/resolveAssetSource');
```
ljharb: can this not be an `import`?

spikebrehm: Yes, I still need to do that for each of the files.

.eslintignore
spikebrehm: In my testing, without ignoring this, `eslint .` takes significantly longer.

PR #527, opened 2016-09-02
Fix path to react-native when developing outside of the example app

spikebrehm:
In #502, the local maven url pointing to the `react-native` node module
changed to work better for the example app. However, that broke the path
to `react-native` when developing in the context of the library, not the
example app. In this PR, specify both paths.

cc: @felipecsl @chrisknepper

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,6 +8,15 @@ buildscript {
       mavenLocal()
       jcenter()
       maven {
+          // For developing the library outside the context of the example app, expect `react-native`
+          // to be installed at `./android/node_modules`.
+          url "$projectDir/../node_modules/react-native/android"
+      }
+      maven {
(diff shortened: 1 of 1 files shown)


felipecsl:
nice! lgtm

spikebrehm:
Just published [0.8.1](https://github.com/lelandrichardson/react-native-maps/releases/tag/v0.8.1) that includes this change.

spikebrehm:
It turns out 0.8.1 was faulty, because my working directory had an additional example app containing the `react-native` node module, which got published as part of the NPM package (:facepalm:), causing a `@providesModule` error, so I've unpublished it and published [0.8.2](https://github.com/lelandrichardson/react-native-maps/releases/tag/v0.8.2).  Use that instead!

Issue #452, opened 2016-08-09
Android crashs on start. No static method zzic


flexy1994:
I've updated react-native `"react-native": "^0.29.2"` and also added the maps `react-native-maps:0.7.1`. Now the Android-App does not start anymore. 

"Unfortunately, MyApp has stopped."

pidcat:

![bildschirmfoto 2016-08-09 um 11 21 56](https://cloud.githubusercontent.com/assets/1196318/17512077/61677514-5e26-11e6-98fc-5396a6d323fe.png)

I have no clue what this stacktrace wants to tell me.

dbanck:
I'm getting the same error.

It might be working with another play-services version, as suggested here: https://github.com/googlemaps/android-maps-utils/issues/258

rstuven:
@dbanck That's it! I made sure all play-services-* in all packages were the same version and it worked.

KGALLET:
Same error, changed my `app/build.gradle` to :
```
    compile 'com.google.android.gms:play-services-auth:9.4.0' // should be at least 9.0.0
    compile 'com.google.android.gms:play-services-maps:9.4.0'
```

fix the issue

ddresch:
In my case it helped to set the version in my app/build.gradle like that, which forces to use this version.

```
compile ('com.google.android.gms:play-services-gcm:8.4.0') {  force = true;  }
```

Just in case you are not able or don't want to upgrade.

Nodonisko:
I write whole article about this problem:
[https://medium.com/@suchydan/how-to-solve-google-play-services-version-collision-in-gradle-dependencies-ef086ae5c75f#.9l0u84y9t](https://medium.com/@suchydan/how-to-solve-google-play-services-version-collision-in-gradle-dependencies-ef086ae5c75f#.9l0u84y9t)

Hope I helped you :)

Issue #420, opened 2016-07-24
My app crashes as soon as i load the map 


BigPun86:
Running on Mac, RN29.2, maps v 0.7.1, Android

```
Could not find generated setter for class com.airbnb.android.react.maps.SizeReportingShadowNode
                         E  FATAL EXCEPTION: main
                         E  Process: com.test.test, PID: 12879
                         E  java.lang.NoSuchMethodError: No static method zzb(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object; in class Lcom/google/android/gms/common/inte
                            rnal/zzx; or its super classes (declaration of 'com.google.android.gms.common.internal.zzx' appears in /data/app/com.test.test-1/base.apk)
                         E      at com.google.android.gms.maps.MapsInitializer.initialize(Unknown Source)
                         E      at com.airbnb.android.react.maps.AirMapManager.createViewInstance(AirMapManager.java:59)
                         E      at com.airbnb.android.react.maps.AirMapManager.createViewInstance(AirMapManager.java:27)
                         E      at com.facebook.react.uimanager.ViewManager.createView(ViewManager.java:44)
                         E      at com.facebook.react.uimanager.NativeViewHierarchyManager.createView(NativeViewHierarchyManager.java:206)
                         E      at com.facebook.react.uimanager.UIViewOperationQueue$CreateViewOperation.execute(UIViewOperationQueue.java:148)
                         E      at com.facebook.react.uimanager.UIViewOperationQueue$DispatchUIFrameCallback.dispatchPendingNonBatchedOperations(UIViewOperationQueue.java:874)
                         E      at com.facebook.react.uimanager.UIViewOperationQueue$DispatchUIFrameCallback.doFrameGuarded(UIViewOperationQueue.java:847)
                         E      at com.facebook.react.uimanager.GuardedChoreographerFrameCallback.doFrame(GuardedChoreographerFrameCallback.java:32)
                         E      at com.facebook.react.uimanager.ReactChoreographer$ReactChoreographerDispatcher.doFrame(ReactChoreographer.java:125)
                         E      at android.view.Choreographer$CallbackRecord.run(Choreographer.java:765)
                         E      at android.view.Choreographer.doCallbacks(Choreographer.java:580)
                         E      at android.view.Choreographer.doFrame(Choreographer.java:549)
                         E      at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:753)
[...]

PR #533, opened 2016-09-02
Bumped google-play-things in build.gradle to 9.+

i8wu:
Bumping build.gradle to use latest version of google-play, should fix crashing when conflicting with other node modules that use the 9.+ version

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -44,6 +44,6 @@ android {
 
 dependencies {
   compile 'com.facebook.react:react-native:0.32.0'
-  compile "com.google.android.gms:play-services-base:8.4.0"
-  compile 'com.google.android.gms:play-services-maps:8.4.0'
+  compile "com.google.android.gms:play-services-base:9.+"
+  compile 'com.google.android.gms:play-services-maps:9.+'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -44,6 +44,6 @@ android {
 
 dependencies {
   compile 'com.facebook.react:react-native:0.32.0'
-  compile "com.google.android.gms:play-services-base:9.+"
-  compile 'com.google.android.gms:play-services-maps:9.+'
+  compile "com.google.android.gms:play-services-base:9.4.0"
+  compile 'com.google.android.gms:play-services-maps:9.4.0'
 }


i8wu:
From my usage when different node_modules have different play versions it causes issues. In other words 8.4.0 doesn't seem to be a minimum, and causes app crash. See #452 #420 . This will cause issues with modules that are using 8.4.0, but I guess from the Compatibility line in the readme, I would assume we would want to strictly use the latest versions of google-play things unless there was a conflict with react-native-maps itself?

jeveloper:
You can keep it at 8.4.0 as i dont think there are any specific features from 9.x that are used 
Also it'll allow for this component to work on older devices (in fact a lot of factory built devices come with outdated google play ).

This will cause a white map btw , i can confirm. 
E.g. if the device has google play services 7.x , it'll be a white map(missing) due to minimum 8.4

i8wu:
Is it common for users not to upgrade their play services? I can't seem to find any statistics about this. Also, while keeping it at 8.4.0 will support older devices, it will cause crashes for modules that use newer versions of google play services. This means that modules that do use features from 9.x will be incompatible with react-native-maps.

felipecsl:
I'd definitely recommend bumping to v9

alexHlebnikov:
Why not update to latest version - 9.4.0?

spikebrehm:
> Why not update to latest version - 9.4.0?

Sounds good.  @felipecsl, our resident Android expert, is recommending that. (Thanks for bearing with my ignorance about Android 😸 .)

@i8wu can you update the PR?

i8wu:
Updated

felipecsl:
👍

spikebrehm:
Thanks!

spikebrehm:
Just published [0.8.1](https://github.com/lelandrichardson/react-native-maps/releases/tag/v0.8.1) that includes this change.

spikebrehm:
It turns out 0.8.1 was faulty, because my working directory had an additional example app containing the `react-native` node module, which got published as part of the NPM package (:facepalm:), causing a `@providesModule` error, so I've unpublished it and published [0.8.2](https://github.com/lelandrichardson/react-native-maps/releases/tag/v0.8.2).  Use that instead!

rops:
Guys we lost support for older devices by upgrading to 9+ though.

felipecsl:
@rops why is that? Are there any devices that are not compatible with GPS v9?

rops:
You would get this nice alert on a device running Android 19 :(

untitled

felipecsl:
yeah that sucks but you should always be able to update your GPS and eventually if you don't, most apps will stop working 😢

rops:
Can you update GPS without doing a full OS upgrade?

felipecsl:
I think so. AFAIK GPS upgrades are not tied to OS version. I scanned through the release notes and didn't find anythink mentioning that in the v9 release.
If you have an android 19 or older and could confirm here, that would be great

i8wu:
It seems my users with older versions of Google Play Services did not get the MapView display, but got a gray view with an "Update" button. When pressing the update button, my app crashes. If the user updates Play Services from the Play Store the MapView renders fine. Is there something in the MapView that needs to be caught?

PR #535, opened 2016-09-03
Add back NativeMethodsMixin

spikebrehm:
I removed `NativeMethodsMixin` in #511 because the module moved in
0.32.0, which made it throw errors. I assumed we didn't need it because
none of its methods were being called, but I didn't realize that we want
to include it to that consumers of the library can use those methods.

to: @lelandrichardson

### DIFF
--- a/components/MapCallout.js
+++ b/components/MapCallout.js
@@ -4,6 +4,7 @@ import {
   requireNativeComponent,
   StyleSheet,
 } from 'react-native';
+import NativeMethodsMixin from 'react/lib/NativeMethodsMixin';
 
 const propTypes = {
   ...View.propTypes,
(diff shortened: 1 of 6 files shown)


alexHlebnikov:
Why not just to return original code?
```
...
const MapCallout = React.createClass({
  mixins: [NativeMethodsMixin],
...
```

spikebrehm:
@alexHlebnikov 

> Why not just to return original code?

Good question. I updated the code to match our [React styleguide](https://github.com/airbnb/javascript/tree/master/react#class-vs-reactcreateclass-vs-stateless), and I was hoping to not have to un-do that.  Plus, we don't need the fancy wrapping of lifecycle methods, like `componentDidMount` etc., for this use case.

But, I'm going to be redoing this to only add in a `setNativeProps()` method.

PR #563, opened 2016-09-10
Add class prefix to EmptyCalloutBackgroundView

terribleben:
This helps when vendoring the library elsewhere, since you can search for symbols beginning with AIR- and SM-.


spikebrehm:
LGTM, thanks! 🍻

PR #564, opened 2016-09-10
[Android] Minor code cleanup

felipecsl:
* Remove unused variables and fields
* Fix compiler warnings (unchecked, deprecation, etc)

### DIFF
--- a/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
+++ b/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
@@ -1,27 +1,25 @@
 package com.airbnb.android.react.maps;
 
+import android.content.Context;
 import android.content.pm.PackageManager;
 import android.content.res.ColorStateList;
 import android.graphics.Bitmap;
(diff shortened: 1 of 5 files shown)


felipecsl:
@spikebrehm this should be a simple one I guess?

android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
```diff
@@ -404,20 +383,14 @@ public void addFeature(View child, int index) {
             AirMapPolyline polylineView = (AirMapPolyline) child;
             polylineView.addToMap(map);
             features.add(index, polylineView);
-            Polyline polyline = (Polyline) polylineView.getFeature();
-            polylineMap.put(polyline, polylineView);
```
spikebrehm: These maps aren't used, eh?  You've smoke tested the example app?

felipecsl: yeah, these maps are never queried

Issue #180, opened 2016-04-14
Error on using fitToElements


gathole:
To set the Google Map zoom level to show all the markers, I was suggested  to use **fitToElements** but I am getting error **"not included points"**

Here is my code and markers have 2 points:

``` javascript
  componentDidMount() {
    this.refs.map.fitToElements(true);
  }

  render() {
    var markers = this.props.markers || [];
    return (
      <MapView
        ref="map"
        style={[styles.map, this.props.mapStyle]}
        initialRegion={this.getInitialRegion(markers)}>
        {markers.map(marker => (
          <MapView.Marker 
            coordinate={marker.latlng}
            title={marker.title}
            description={marker.description}
            image={marker.image}/>
        ))}
      </MapView>
    );
  }

```

christopherdro:
Whats the data in `marker.latlng` look like?

gathole:
```javascript
let markers = [{
      latlng: {
        latitude: this.props.currentBooking.pickup_latitude,
        longitude: this.props.currentBooking.pickup_longitude
      },
      image: require('../images/customer.png')
    }, {
      latlng: {
        latitude: this.props.driverLocation.latitude,
        longitude: this.props.driverLocation.longitude
      },
      image: require('../images/onway.png')
    }];
```

Latitude and Longitude are string value... I think it supposed to be a Point... Let me know if I am correct and Thanks for reply

jrichardlai:
Not sure if thats the issue but can you cast the latitude and longitude to a float?

gathole:
I cast the latitude and longitude to a float but getting the same error "No included points". Can you please explain where should I mention points and what would be the values (x and y) to be passed in Point?

Let me clarify one more thing is that I am working on **Android**.

DennisMG:
The docs says `fitToElements` method receives argument `animated: Boolean` ... I'm not sure what a Animated: Boolean is!

DennisMG:
I found the solution, you just have to add an empty js object, like this: `this.refs.map.fitToElements(true,{});` I don't know why but it worked.

simonhintersonnleitner:
@DennisMG `this.refs.map.fitToElements(true,{});` do not work for me.

DennisMG:
I thought it solved the problem but sometimes throws the error described above and other times works perfectly

DennisMG:
try changing the fitToElements function in react-native-maps/android/src/main/java/com/AirMaps/AirMapView.java for this:

```
[...]

PR #567, opened 2016-09-11
Don't fit to elements if no positions added

ryankask:
This commit prevents the crash mentioned in #180. It is identical to the implementation in `AirMapView#fitToSuppliedMarkers()`.

This doesn't actually fix the issue mentioned in #180 because while the app no longer crashes, the elements aren't zoomed unless you use a setTimeout hack or something similar. @igrayson has written about this there.

It would be great to figure out why that isn't working, however.


spikebrehm:
LGTM, thanks.

PR #577, opened 2016-09-14
[WIP] Add Appium tests

jrichardlai:
This is something I had in my local for a while, I don't think this is ready.
Just wanted to push it see if it can bring any value to the project, and bring some discussions about testing.

**How this works**

I'm basing the tests on the example project, and do an integration testing using Appium.

`npm run compile:test-ios` will compile the app
`npm run mocha:test-ios` will run the test suite for iOS

Unfortunately for some reason map elements are hard to test.

For `Callouts.js` wanted to check if the marker callout was open but didn't find a way ( I can do that by having some text showing in a Text about the status of the callout ).
So right now the tests are not really interesting but they could be more if we think more about what the view is doing and showing.

**NOTES**

Android compilation does not seems to be working anymore, had it working before. How do you guys run the example project in local?

**TODOS**

I put it in those tests on Travis before, but they take a while to run :/, we can cache the compilation to make it faster.

Compilation:

compilation

Tests Running:

test-running


spikebrehm:
This is awesome!!! Thank you!!! We'll take a look soon.

Re: the Android compile issues, we've seen that reported a few times, and will look into it.

lelandrichardson:
YASSSSSSSSSS.  @jrichardlai awesome work. Let's definitely figure out a way to get this in!

jrichardlai:
**(iOS)** status note here:

I got the test running on Travis ( green yay ), it will take I think max 20 mins for the whole suite to run (right now run with 14mins but was not really a relevant suite), I think for this project it might be acceptable as there should not be a lot of PR's that would go in. I also separated eslint and Appium. Thoughts?

However I upgraded to XCode 8 and seems that there is several issues with Appium (BOO), as Travis supports it I think its better to be on XCode 8 anyways, but will need more investigation to fix the build.

lelandrichardson:
@jrichardlai awesome work. Yeah, we should all be working out of xcode8 if possible, so if things are breaking right now we should fix (separate PR maybe?).

.travis.yml
```diff
@@ -1,11 +1,32 @@
...
 
 cache:
   directories:
     - node_modules
     - example/node_modules
+    - ~/.nvm
+
+before_install:
+  - mkdir -p ~/.nvm
+  - which nvm || [ -f ~/.nvm/nvm.sh ] || (git clone https://github.com/creationix/nvm.git ~/.nvm && cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`)
```
spikebrehm: Do we have to manually install node now that the language is no longer `node_js`?

cli/compiler/AbstractCompiler.js
```diff
@@ -0,0 +1,62 @@
+import childProcess from 'child_process';
+import path from 'path';
+
+class Compiler {
```
spikebrehm: No big deal but may as well make the classname match the filename: `AssetCompiler`

cli/compiler/AbstractCompiler.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  makeDirectories() {
+    this.execute(`mkdir -p ${this.buildDirectory}`);
+    this.execute(`mkdir -p ${this.appDirectory}`);
+  }
+
+  calculateDirectories() {
+    this.rootDirectory = path.join(process.cwd(), this.projectPath);
+    this.platformDirectory = `${this.rootDirectory}/${this.platform}`;
```
spikebrehm: Could also use `path.join()` for these

package.json
```diff
@@ -6,7 +6,12 @@
   "version": "0.8.0",
   "scripts": {
     "start": "react-native start",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "compile:test-ios": "PLATFORM=ios babel-node cli/compile.js --presets react-native",
```
spikebrehm: For which language features do we need `babel-node`? Is Node 6.x's ES6 support not sufficient? Just trying to introduce fewer moving parts if we can get away with it.

test/helpers/appium.js
```diff
@@ -0,0 +1,69 @@
...
+
+const startAppium = ({ appiumServerConfig }) => (
+  new Promise((resolve) => {
+    const appiumProc = childProcess.spawn('appium', ['-p', appiumServerConfig.port]);
+
+    let loadedAppium = null;
+    appiumProc.stdout.on('data', (data) => {
+      if (loadedAppium) return;
+      console.log(`[APPIUM]: ${data}`);
+      if (data.indexOf('Appium REST http interface listener started') >= 0) {
```
spikebrehm: Can also do: 

```js
if (data.includes('Appium REST http interface listener started')) {
```

test/helpers/appium.js
```diff
@@ -0,0 +1,69 @@
+import 'colors';
+import wd from 'yiewd';
+import childProcess from 'child_process';
+
+const startAppium = ({ appiumServerConfig }) => (
+  new Promise((resolve) => {
+    const appiumProc = childProcess.spawn('appium', ['-p', appiumServerConfig.port]);
+
+    let loadedAppium = null;
```
spikebrehm: Should initialize to `false`?

cli/travis_before_install_appium.sh
```diff
@@ -0,0 +1,12 @@
+if [ -z "$TRAVIS_BUILD_DIR" ]; then
```
spikebrehm: Don't forget to chmod +x this file

spikebrehm:
I'm getting this error when running mocha:

```
[APPIUM]: Error [MJSONWP] Encountered internal error running command: Error: Xcode version [object Object] is not yet supported
```

Perhaps because i'm on Xcode 8?

test/integrations/Callouts.test.js
```diff
@@ -0,0 +1,37 @@
...
+import integration from '../helpers/integration';
+
+const { it, isIOS } = integration;
+
+describe('Callouts', () => {
+  // For some reason appium can't see if the marker is opened or not
+  it('should show custom callouts', function* callback(driver, done) {
+    yield driver.elementById('Custom Callouts').click();
+
+    yield driver.elementById('Tap on markers to see different callouts');
```
spikebrehm: What does this do, just calling `elementById()` but not clicking or anything? It waits for it?

test/integrations/Callouts.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+    // Click on the first callout
+    // It shows the callout
+    if (isIOS) {
+      yield driver.elementById('This is a title').click();
+    } else {
+      yield driver.elementByXPath(
```
spikebrehm: whoaaa xpath. This is necessary for Android for some reason?

spikebrehm:
Also, the Android build works for me locally.  Maybe try again?

.travis.yml
```diff
@@ -1,11 +1,32 @@
-language: node_js
-
-node_js:
-  - "6"
+language: objective-c
+osx_image: xcode7
```
jrichardlai: Will need to, I didn't changed this yet, I'm still looking on how to fix the test suite Appium 1.5.3 is the current latest released version, I think `v1.6.0-beta1` of Appium might supports xcode 8.

.travis.yml
jrichardlai: Yes if we use a different language then we will need to install it manually

cli/compiler/AbstractCompiler.js
jrichardlai: Thanks missed that. I think I'm just gonna change those into a bash script for readability, those js Classes were useful in my other projects where I use that for different environments.

package.json
jrichardlai: Yes thanks! Forgot to update this!

test/integrations/Callouts.test.js
jrichardlai: It was just an example, I could not compile anymore android, should not be needed

cli/travis_before_install_appium.sh
jrichardlai: Executing it with the dot command in .travis.yml

jrichardlai:
Thanks @lelandrichardson, @spikebrehm for looking at the PR!

package.json
spikebrehm: Although now that I see that the tests are using async/await and generators, we probably still need this?

test/integrations/PolygonCreator.test.js
```diff
@@ -0,0 +1,66 @@
+import 'babel-polyfill';
+import { TouchAction } from 'wd';
+import integration from '../helpers/integration';
+
+const { it } = integration;
+
+describe('FitToSuppliedMarkers', () => {
+  it.only('should Track the position', function* callback(driver, done) {
```
lelandrichardson: we need to get rid of this `.only`, no?

scalp42:
LGTM @jrichardlai

rborn:
@jrichardlai great job. Are you willing to update the PR for the last version of the module and be sure it still runs? 😁
Thanks a bunch  🤗

rborn:
@jrichardlai  we're trying to cleanup the repo, so I'll close this PR . I got no reply from you on my last message in December  so please ping me if you think you can continue working on it and I'll re-open 🤗

PR #591, opened 2016-09-18
Updated Mapview onRegionChange events notes

Alastairm:
Clarified the difference between onRegionChange and onRegionChange
notes for each match comments in MapView.js


spikebrehm:
👍 🍻 thanks

PR #592, opened 2016-09-18
Added support for AnimatedRegion without modifying the AnimatedImplementation.js of react-native

IjzerenHein:
Hi,

This pull requests adds support for AnimatedRegion without the need for modifying AnimatedImplemention.js.
I've added AnimatedRegion to the MapView js component so it can be included from there.
I've also updated the README and added an example.

Awesome library, keep up the good work 😃

cheers, Hein

### DIFF
--- a/components/AnimatedRegion.js
+++ b/components/AnimatedRegion.js
@@ -1,45 +1,50 @@
 /* eslint-disable */
+import {Animated} from 'react-native';
 
-class AnimatedRegion extends AnimatedWithChildren {
-  // latitude: AnimatedValue;
-  // longitude: AnimatedValue;
-  // latitudeDelta: AnimatedValue;
-  // longitudeDelta: AnimatedValue;
-  // _listeners: {[key: string]: {
+const AnimatedWithChildren = Object.getPrototypeOf(Animated.ValueXY);
+
+var _uniqueId = 1;
+
+export default class AnimatedMapRegion extends AnimatedWithChildren {
+  //latitude: AnimatedValue;
+  //longitude: AnimatedValue;
+  //latitudeDelta: AnimatedValue;
+  //longitudeDelta: AnimatedValue;
+  //_listeners: {[key: string]: {
   //  latitude: string,
   //  longitude: string,
   //  latitudeDelta: string;
   //  longitudeDelta: string,
-  // }};
+  //}};
 
   constructor(valueIn) {
     super();
-    const value = valueIn || { // probably want to come up with better defaults
+    var value = valueIn || { // probably want to come up with better defaults
       latitude: 0,
       longitude: 0,
       latitudeDelta: 0,
-      longitudeDelta: 0,
+      longitudeDelta: 0
     };
-    this.latitude = value.latitude instanceof Animated
+    this.latitude = value.latitude instanceof Animated.Value
       ? value.latitude
-      : new AnimatedValue(value.latitude);
-    this.longitude = value.longitude instanceof Animated
+      : new Animated.Value(value.latitude);
+    this.longitude = value.longitude instanceof Animated.Value
       ? value.longitude
-      : new AnimatedValue(value.longitude);
-    this.latitudeDelta = value.latitudeDelta instanceof Animated
+      : new Animated.Value(value.longitude);
+    this.latitudeDelta = value.latitudeDelta instanceof Animated.Value
       ? value.latitudeDelta
-      : new AnimatedValue(value.latitudeDelta);
-    this.longitudeDelta = value.longitudeDelta instanceof Animated
+      : new Animated.Value(value.latitudeDelta);
+    this.longitudeDelta = value.longitudeDelta instanceof Animated.Value
       ? value.longitudeDelta
-      : new AnimatedValue(value.longitudeDelta);
+      : new Animated.Value(value.longitudeDelta);
     this._listeners = {};
   }
 
   setValue(value) {
-    // this.latitude.setValue(value.latitude);
-    // this.longitude.setValue(value.longitude);
-    // this.latitudeDelta.setValue(value.latitudeDelta);
-    // this.longitudeDelta.setValue(value.longitudeDelta);
+    //this.latitude.setValue(value.latitude);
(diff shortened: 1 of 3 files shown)


spikebrehm:
Thanks, we'll give this a try! Note that there's a merge conflict.

PTAL @lelandrichardson

IjzerenHein:
Awesome!

Do I need to do anything about the merge conflict? I don't seem to be able to see the conflicts.

cheers

spikebrehm:
Re: the merge conflict, here's what i see on GitHub:

screen shot 2016-09-21 at 1 46 03 pm

IjzerenHein:
Yeah I'm seeing that too (except for the Merge Pull request option). But I can't seem to see anywhere what the actual conflicts are.

spikebrehm:
@IjzerenHein did you try rebasing from master?  

    git pull --rebase origin master

components/AnimatedRegion.js
```diff
@@ -1,45 +1,50 @@
 /* eslint-disable */
+import {Animated} from 'react-native';
 
-class AnimatedRegion extends AnimatedWithChildren {
-  // latitude: AnimatedValue;
-  // longitude: AnimatedValue;
-  // latitudeDelta: AnimatedValue;
-  // longitudeDelta: AnimatedValue;
-  // _listeners: {[key: string]: {
+const AnimatedWithChildren = Object.getPrototypeOf(Animated.ValueXY);
```
lelandrichardson: oooh. this is sneaky.

components/AnimatedRegion.js
```diff
@@ -1,45 +1,50 @@
...
-    const value = valueIn || { // probably want to come up with better defaults
+    var value = valueIn || { // probably want to come up with better defaults
       latitude: 0,
       longitude: 0,
       latitudeDelta: 0,
-      longitudeDelta: 0,
+      longitudeDelta: 0
     };
-    this.latitude = value.latitude instanceof Animated
+    this.latitude = value.latitude instanceof Animated.Value
```
lelandrichardson: hmm. We should avoid this if possible. It prevents us from allowing latitude/longitude values from being interpolated values, for instance.  I wonder if we can get the `Animated` class with more of your `getPrototypeOf` voodoo? :)

lelandrichardson:
wow @IjzerenHein this is a great PR!  I'd love to get this in. Do you think you can try to tweak so we can use `instanceof Animated` instead of `instanceof Animated.Value` checks?  Thanks!

IjzerenHein:
@lelandrichardson I'll have a look. Really busy at the moment so it may take a while.
cheers

components/AnimatedRegion.js
IjzerenHein: Well, it's not really that sneaky :) There's prototype chains in Javascript, so why not use them. Ideally react-native would allows us to create our own custom composition classes, but until that is the case we have to be a bit creative. This is quite solid, but I do think we should add a check to ensure that the parent prototype of Animated.ValueXY is indeed AnimatedWithChildren. I'll try to add some code that does this, so in case it breaks we'll immediately know.

components/AnimatedRegion.js
IjzerenHein: Indeed, good call. I modelled it after ValueXY which also only accepts number or instances of `Animated.Value`. It didn't occur to me that you may also want to interpolate the values. I'll give it a go and see whether this is achievable.

IjzerenHein: I've had a look but it doesn't seem possible to get `.interpolate, Animated.add, etc..` working without having to modify `AnimatedImplementation.js`. The crux is in the `maybeVectorAnim` function which explicitly checks for ValueXY class.
https://github.com/facebook/react-native/blob/5deb528695a8a47156a331b2a0182fabef0253d7/Libraries/Animated/src/AnimatedImplementation.js#L1831

 If we could replace that function with an extended one that support AnimatedRegion, then `.interpolate, Animated.multiply, etc..` should work with AnimatedRegion. It would have to be changed to this:

```js
var maybeVectorAnim = function(
  value: AnimatedValue | AnimatedValueXY | AnimatedRegion,
  config: Object,
  anim: (value: AnimatedValue, config: Object) => CompositeAnimation
): ?CompositeAnimation {
  if (value instanceof AnimatedValueXY) {
    var configX = {...config};
    var configY = {...config};
    for (var key in config) {
      var {x, y} = config[key];
      if (x !== undefined && y !== undefined) {
        configX[key] = x;
        configY[key] = y;
      }
    }
    var aX = anim((value: AnimatedValueXY).x, configX);
    var aY = anim((value: AnimatedValueXY).y, configY);
    // We use `stopTogether: false` here because otherwise tracking will break
    // because the second animation will get stopped before it can update.
    return parallel([aX, aY], {stopTogether: false});
  }
  else if (value instanceof AnimatedRegion) {
    var configLat = {...config};
    var configLng = {...config};
    var configLatDelta = {...config};
    var configLngDelta = {...config};
    for (var key in config) {
      var {lat, lng, latDelta, lngDelta} = config[key];
      if (lat !== undefined && lng !== undefined && latDelta !== undefined && lngDelta !== undefined) {
        configLat[key] = lat;
        configLng[key] = lng;
        configLatDelta[key] = latDelta;
        configLngDelta[key] = lngDelta;
      }
    }
    var aLat = anim((value: AnimatedRegion).latitude, configLat);
    var aLng = anim((value: AnimatedRegion).longitude, configLng);
    var aLatDelta = anim((value: AnimatedRegion).latitudeDelta, configLatDelta);
    var aLngDelta = anim((value: AnimatedRegion).longitudeDelta, configLngDelta);
    // We use `stopTogether: false` here because otherwise tracking will break
    // because the second animation will get stopped before it can update.
    return parallel([aLat, aLng, aLatDelta, aLngDelta], {stopTogether: false});
  }
  return null;
};
```

IjzerenHein:
I've created a new pull request for this (#608), which is rebased on the latest master and adds an additional check to ensure the AnimatedWithChildren class is correctly obtained.

spikebrehm:
Closing because this is superceded by #608.

PR #597, opened 2016-09-22
Fix list of examples on Android

spikebrehm:
In #548, we added Google Maps support for iOS. We changed the example
app to list examples based on whether they support Google Maps on iOS.
The logic was such that on Android it only showed the examples that are
iOS + Google Map compatible, however on Android this condition is
irrelevant, this commit changes it to show all examples on Android,
regardless.

### Before
screen shot 2016-09-22 at 10 37 24 am

### After
screen shot 2016-09-22 at 10 37 04 am

to: @gilbox


gilbox:
thanks for fixing! 👍

PR #608, opened 2016-09-25
Added support for AnimatedRegion without modifying the AnimatedImplementation.js of react-native

IjzerenHein:
This pull request replaces the old one (#592) (was having too much trouble rebasing due to upstream changes). This pull request is rebased on the latest master and adds an additional check to verify that the AnimatedWithChildren class has been correctly obtained.

### DIFF
--- a/components/AnimatedRegion.js
+++ b/components/AnimatedRegion.js
@@ -1,45 +1,38 @@
 /* eslint-disable */
+import {Animated} from 'react-native';
 
-class AnimatedRegion extends AnimatedWithChildren {
-  // latitude: AnimatedValue;
-  // longitude: AnimatedValue;
-  // latitudeDelta: AnimatedValue;
-  // longitudeDelta: AnimatedValue;
-  // _listeners: {[key: string]: {
-  //  latitude: string,
-  //  longitude: string,
-  //  latitudeDelta: string;
-  //  longitudeDelta: string,
-  // }};
+const AnimatedWithChildren = Object.getPrototypeOf(Animated.ValueXY);
+if (AnimatedWithChildren.name !== 'AnimatedWithChildren') console.error('AnimatedRegion could not obtain AnimatedWithChildren base class');
+// const __Animated = Object.getPrototypeOf(AnimatedWithChildren);
+// if (__Animated.name !== 'Animated') console.error('AnimatedRegion could not obtain Animated base class');
 
+var _uniqueId = 1;
+
+export default class AnimatedMapRegion extends AnimatedWithChildren {
   constructor(valueIn) {
     super();
-    const value = valueIn || { // probably want to come up with better defaults
+    var value = valueIn || { // probably want to come up with better defaults
       latitude: 0,
       longitude: 0,
       latitudeDelta: 0,
-      longitudeDelta: 0,
+      longitudeDelta: 0
     };
-    this.latitude = value.latitude instanceof Animated
+    this.latitude = value.latitude instanceof Animated.Value
       ? value.latitude
-      : new AnimatedValue(value.latitude);
-    this.longitude = value.longitude instanceof Animated
+      : new Animated.Value(value.latitude);
+    this.longitude = value.longitude instanceof Animated.Value
       ? value.longitude
-      : new AnimatedValue(value.longitude);
-    this.latitudeDelta = value.latitudeDelta instanceof Animated
+      : new Animated.Value(value.longitude);
+    this.latitudeDelta = value.latitudeDelta instanceof Animated.Value
       ? value.latitudeDelta
-      : new AnimatedValue(value.latitudeDelta);
-    this.longitudeDelta = value.longitudeDelta instanceof Animated
+      : new Animated.Value(value.latitudeDelta);
+    this.longitudeDelta = value.longitudeDelta instanceof Animated.Value
       ? value.longitudeDelta
-      : new AnimatedValue(value.longitudeDelta);
+      : new Animated.Value(value.longitudeDelta);
     this._listeners = {};
   }
 
   setValue(value) {
-    // this.latitude.setValue(value.latitude);
-    // this.longitude.setValue(value.longitude);
-    // this.latitudeDelta.setValue(value.latitudeDelta);
-    // this.longitudeDelta.setValue(value.longitudeDelta);
     this.latitude._value = value.latitude;
     this.longitude._value = value.longitude;
     this.latitudeDelta._value = value.latitudeDelta;
@@ -65,7 +58,7 @@ class AnimatedRegion extends AnimatedWithChildren {
       latitude: this.latitude.__getValue(),
       longitude: this.longitude.__getValue(),
       latitudeDelta: this.latitudeDelta.__getValue(),
-      longitudeDelta: this.longitudeDelta.__getValue(),
+      longitudeDelta: this.longitudeDelta.__getValue()
     };
   }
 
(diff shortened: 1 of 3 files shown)


lelandrichardson:
FYI @IjzerenHein , I hadn't really thought to do something like this (get `AnimatedWithChildren` through the prototype chain) because the choice to not export those classes was made explicitly, and I was trying to figure out a good way to extend this system.

I came up with a bit of a strategy, which I PR'd here: https://github.com/animatedjs/animated/pull/12  If you'd be interested in taking a look.

components/AnimatedRegion.js
```diff
@@ -1,45 +1,38 @@
...
-  // latitudeDelta: AnimatedValue;
-  // longitudeDelta: AnimatedValue;
-  // _listeners: {[key: string]: {
-  //  latitude: string,
-  //  longitude: string,
-  //  latitudeDelta: string;
-  //  longitudeDelta: string,
-  // }};
+const AnimatedWithChildren = Object.getPrototypeOf(Animated.ValueXY);
+if (AnimatedWithChildren.name !== 'AnimatedWithChildren') console.error('AnimatedRegion could not obtain AnimatedWithChildren base class');
```
lelandrichardson: will this get preserved through a minification pass?

IjzerenHein: Great question :)
I'll let you know when I know for sure :)

lelandrichardson: might not be terrible to just wrap it in `if(__DEV__)`

lelandrichardson:
I suppose if we added `addListener` to the prototype of `Animated`, then we'd be okay just using the `instanceof Animated` check, or am I missing something?. IIRC from talking to @vjeux `addListener` could be on the Animated "interface" if we wanted it to be, there just wasn't a need for it at the time.

vjeux:
I don't really care where this lives :) If putting it on the prototype makes it easier for you, feel free to

components/AnimatedRegion.js
IjzerenHein: I made a release build but I couldn't get it to trigger the error.
Still, I think you're right and we should stay on the safe side here and wrap it in a __DEV__

IjzerenHein:
@lelandrichardson I added the __DEV__ check 👍 

As for the interpolate discussion. I'd love to take a closer look at that but frankly I've got a big deadline coming up I don't have the time nor the need for it (at the moment).
I'll revisit this in a later stage. Ping me in case I forget.

cheers, Hein

lelandrichardson:
@IjzerenHein for sure. this is strictly better than how it was before.

This LGTM.

mjsisley:
I am unable to get the example running on your PR (after changing Animated.Region to MapView.AnimatedRegion. Running the example project with this PR (and also the simpler example from the readme, results in: `undefined in not an object (evaluating 'this.state.XXX). 

Is it possible to add a simple example?

IjzerenHein:
@mjsisley Hmm that's odd. Could you check whether the `MapView.AnimatedRegion` property contains a valid reference to the class? Could you also include a stacktrace of the error?

So, `MapView.AnimatedRegion` is basically a direct replacement for `Animated.Region`, so it should work exactly the same. It does have some restrictions and should be used slightly as `Animated.Value` (you can't use `Animated.spring` or `Animated.timing` but should call `.spring` and `.timing` directly on the AnimatedRegion).

Example:

```js
export default class MyView extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      coordinate: new MapView.AnimatedRegion({
        latitude: 5.2443,
        longitude: 6.1223
      })
    }
  }

  render() {
    return <MapView>
      <MapView.Marker.Animated coordinate={this.state.coordinate} />
    </MapView>;
  }

  setCoordinateImmediate() {
    this.state.coordinate.setValue({
      latitude: 7.234,
      longitude: 1.233
    });
  }

  setCoordinateAnimated() {
    this.state.coordinate.timing({
      latitude: 7.234,
      longitude: 1.233,
      duration: 1000
    }).start();
  }
}
```

mjsisley:
The full stack trace when the [original example](https://github.com/airbnb/react-native-maps/blob/master/example/examples/AnimatedViews.js) is used with your PR subbing MapView.AnimatedRegion for Animated.Region:

```
undefined is not an object (evaluating 'this.state.panY') 
onMoveShouldSetPanResponder AnimatedViews.js:292 
executeDispatchesInOrderStopAtTrueImpl EventPluginUtils.js:137 executeDispatchesInOrderStopAtTrue EventPluginUtils.js:153 
setResponderAndExtractTransfer ResponderEventPlugin.js:334 
extractEvents ResponderEventPlugin.js:443 
extractEvents EventPluginHub.js:194 
handleTopLevel ReactEventEmitterMixin.js:28 
<unknown> ReactNativeEventEmitter.js:125 
perform Transaction.js:138 
batchedUpdates ReactDefaultBatchingStrategy.js:63 
batchedUpdates ReactUpdates.js:98 
_receiveRootNodeIDEvent ReactNativeEventEmitter.js:124 
receiveTouches ReactNativeEventEmitter.js:186 
__callFunction MessageQueue.js:229 
<unknown> MessageQueue.js:111 
guard MessageQueue.js:42 
callFunctionReturnFlushedQueue MessageQueue.js:110
```

It seems the problem is with the usage PanResponder, and thus unrelated to the PR. Just looked closer at the examples and realized that AnimatedViews example is also the only example using the PanResponder... so this is a separate issue.

Thanks for you help!

Plugging the following into the AirMapsExplorer project does work (slight modification of your example for the example project):

```jsx
import React from 'react';
import {
  StyleSheet,
  View,
  Animated,
} from 'react-native';

import MapView from 'react-native-maps';

class MyView extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      coordinate: new MapView.AnimatedRegion({
        latitude: 5.2443,
        longitude: 6.1223
      })
    }
  }

  render() {
    return <View style={styles.container}>
      <MapView style={styles.map}>
        <MapView.Marker.Animated coordinate={this.state.coordinate} />
      </MapView>
    </View>;
  }

  setCoordinateImmediate() {
    this.state.coordinate.setValue({
      latitude: 7.234,
      longitude: 1.233
    });
  }

  setCoordinateAnimated() {
    this.state.coordinate.timing({
      latitude: 7.234,
      longitude: 1.233,
      duration: 1000
    }).start();
  }
}

  container: {
    ...StyleSheet.absoluteFillObject,
  },
  map: {
    backgroundColor: 'transparent',
    ...StyleSheet.absoluteFillObject,
  },
});

module.exports = MyView;
```

IjzerenHein:
Alright, great to hear that has been resolved @mjsisley 👍

components/AnimatedRegion.js
```diff
@@ -1,45 +1,40 @@
...
-  //  latitude: string,
-  //  longitude: string,
-  //  latitudeDelta: string;
-  //  longitudeDelta: string,
-  // }};
+const AnimatedWithChildren = Object.getPrototypeOf(Animated.ValueXY);
+if (__DEV__) {
+  if (AnimatedWithChildren.name !== 'AnimatedWithChildren') console.error('AnimatedRegion could not obtain AnimatedWithChildren base class');
+}
+// const __Animated = Object.getPrototypeOf(AnimatedWithChildren);
```
spikebrehm: Remove these comments?

spikebrehm:
Awesome! Thanks for your hard work! 🍻

Issue #611, opened 2016-09-26
Build fails when not using CocoaPods


sjmueller:
I have a react-native project that only uses carthage, no cocoapods. When using the latest react-native-maps from master (manual setup), the build fails on `PhaseScriptExecution Check\ Pods\ Manifest.lock`:

![image](https://cloud.githubusercontent.com/assets/29085/18823074/431c998a-836b-11e6-8e15-b6332be545b1.png)

This is most likely due to the build scripts that were added in support of google maps:
https://github.com/airbnb/react-native-maps/commit/f78c0ee674a2a67bc72f843779cd3eb0f16e1675#diff-b5c34708840d3b818caefbc631f8e514R398

Is the plan going forward to only support projects that use cocoapods?

simonmitchell:
Same issue here. We've never used CocoaPods due to issues we had with it, and never plan on doing so. Will you be supporting manual install?

mjsisley:
I am attempting to run with cocoapods, and am also getting this error. I can't run the example project on iOS when following the directions.

simonmitchell:
I managed to get it working with cocoapods in the end (manual was just not going to work at all) but had to make the change in my pull request here: https://github.com/airbnb/react-native-maps/pull/614

gilbox:
My b. I'll look into it.

gilbox:
https://github.com/airbnb/react-native-maps/pull/620

mjsisley:
#620 solved this issue. Thanks @gilbox.

For me, using cocoapods 1.0.1, xcode 8.0, and the example (and instructions) it also took the following (as a record for anyone looking):

1. senpng's comment on https://github.com/facebook/react-native/issues/9014
2. http://stackoverflow.com/questions/26263637/apple-mach-o-linker-error-cocoapods

spikebrehm:
Closed by #620

PR #614, opened 2016-09-26
Fix/nullability

simonmitchell:
Fixes compiler error with trying to specify nullability on a BOOL


spikebrehm:
@simonmitchell in #611 you said that this fix is necessary to get the project to build with CocoaPods.  How does this change affect CocoaPods builds?

In [this commit](https://github.com/airbnb/react-native-maps/commit/36476025deed5faac2a469c2fa6076ad4e9802aa#diff-a537138599914b3a6ff2af1efb65be4eR442), @naoufal added default values to the `fitToCoordinates` options, so `animated` should not be `null` unless you're manually passing it in, right?

gilbox:
@spikebrehm i had to make this same fix for a from-scratch simple non-cocoapods test project.

spikebrehm:
👍

simonmitchell:
Yeah sorry for the confusion, wasn't related to the CocoaPods issue, was just telling that person that even after I used CocoaPods I still needed to make changes.

PR #615, opened 2016-09-26
Added support for drawing polylines on snapshots on iOS

IjzerenHein:
This pull request fixes the issue that polylines are not drawn on snapshots on iOS.

Specifically, it adds infrastructure (the AIRMapSnapshot protocol) to which overlays can conform. The AIRMapSnapshot protocol has one method called `drawToSnapshot` which can optionally be implemented by overlays.

Additionally, this pull request adds the `drawToSnapshot` implementation for the AIRMapPolyline class, causing it to be rendered to snapshots. It should be fairly easy to extends the overlay to support snapshot drawing as well.

The example below shows the image that was generated using the `takeSnapshot` function, including the polyline overlay:

image


spikebrehm:
Nice, thanks!

IjzerenHein:
Awesome, thanks for pulling!

FYI, I'm working on the Android implementation of taking snapshots. I'll hopefully push that soon.

jeveloper:
Thanks @IjzerenHein  !

Any chance you can provide an example?
Is that something end user can drag and it draws a line or you pass an array of x,y and it draws lines ?

thanks

IjzerenHein:
@jeveloper Serge, no you have to add `<Polyline ... />` as a child of `<MapView>` and then you can call `takeSnapshot` to render the map & polyline to an image. The documentation for Polyline and takeSnapshot is in the main README.

cheers

jeveloper:
thanks for sharing that info :)

PR #613, opened 2016-09-26
change loadImageWithURLRequest for RN 0.35

sjmueller:
Signature changed here:

https://github.com/facebook/react-native/commit/820b1c0e46f781f6677caa3a62b59076bf2700e5

PR #623, opened 2016-09-28
Zoom over images

### LINKED ISSUE #522: can't pinch to zoom when touching marker
When I want to zoom map and initial touches are on markers — it's not zooming

sjmueller:
This change allows the map to respond properly to pan and zoom gestures that originate over Image markers, and would close #522.

There are probably consequences to the way this fix is implemented:
- Some gestures may not work properly for markers, such as dragging/moving, ~~although single taps seem to still be received~~ no events respond besides tapping to reveal a callout. Taps within the callout work though
- Perhaps we should call `super.hitTest` anyways, to check for  nil so that we don’t respond to non-map related events
- If apple changed the order or hierarchy of subviews, well this would obviously break

### DIFF
--- a/ios/AirMaps/AIRMapMarker.m
+++ b/ios/AirMaps/AIRMapMarker.m
@@ -215,6 +215,7 @@
                                                                          clipped:YES
                                                                       resizeMode:RCTResizeModeCenter
                                                                    progressBlock:nil
+                                                                partialLoadBlock:nil
                                                                  completionBlock:^(NSError *error, UIImage *image) {
                                                                      if (error) {
                                                                          // TODO(lmr): do something with the error?
--- a/ios/AirMaps/AIRMap.m
+++ b/ios/AirMaps/AIRMap.m
@@ -155,7 +155,16 @@ const CGFloat AIRMapZoomBoundBuffer = 0.01;
     UIView *calloutMaybe = [self.calloutView hitTest:[self.calloutView convertPoint:point fromView:self] withEvent:event];
     if (calloutMaybe) return calloutMaybe;
 
-    return [super hitTest:point withEvent:event];
+    // If it's not a callout, then always return the MKNewAnnotationContainerView which will handle pinch & zoom properly
+    // - MKMapView
+    // - - UIView
+    // - - - MKBasicMapView
+    // - - - - _MKMapLayerHostingView
+    // - - - MKScrollContainerView
+    // - - - - MKOverlayContainerView
+    // - - - MKNewAnnotationContainerView
+    // - - MKAttributionLabel
+    return ((UIView *)[((UIView *)[self.subviews objectAtIndex:0]).subviews objectAtIndex:2]);
 }
 
 #pragma mark SMCalloutViewDelegate


ios/AirMaps/AIRMapMarker.m
```diff
@@ -215,6 +215,7 @@ - (void)setImageSrc:(NSString *)imageSrc
                                                                          clipped:YES
                                                                       resizeMode:RCTResizeModeCenter
                                                                    progressBlock:nil
+                                                                partialLoadBlock:nil
                                                                  completionBlock:^(NSError *error, UIImage *image) {
```
christopherdro: This line can be removed since #613 was merged.

carleryd:
I badly want this feature, but this does not work for my custom markers on either 0.8.2, 0.9.0 or 0.10.0 with React Native 0.34.
My custom markers are created with a remote image like:
```<MapView.Marker> <Image source={{uri: <remote_image>}}/> </MapView.Marker>```

**Comment by sjmueller** [comment]
@carleryd yep, unfortunately right now this PR is a tradeoff, because markers don't respond to taps with this change.  The exception here is if you have a callout, tapping on a marker will reveal it. Then from within the callout you can have touchable events work properly.

I've spent a great deal of time trying to figure out ways to have the best of both worlds, but unfortunately coming up short.  The challenge I got stuck on was detecting what type of event is transpiring within the hitTest. Typically a UITouch event object would contain an array of touches that could be inspected to determine the gesture, but in our case the hitTest is invoked with an event that doesn't contain this information.

We arent the only one that has experienced the mysterious empty set of touches:
http://stackoverflow.com/questions/387938/uievent-passed-into-uiview-hittest

Will keep investigating when I have another block of time.

**Comment by carleryd** [comment]
Ah thats too bad, I've spent a fair amount of time trying to find workarounds in React Native to solve the problem, unfortunately I don't know enough native iOS development to help you out.
One thing I have noticed though is that markers behave correctly the first time they are rendered.
So if I render a list of markers like this:
```
render() {
    markers.map((marker) => <MapView.Marker key={"id" + Date.new()} />
}
```
The markers blink every time I move around the map, but they are both clickable and zoomable.
I explain the problem more in my issue https://github.com/airbnb/react-native-maps/issues/544
Good luck solving it, I am really looking forward to it 👍

**Comment by stereodenis** [comment]
@carleryd btw, you can get uniq key like this

```
render() {
    markers.map((marker, index) => <MapView.Marker key={index} />
}
```

carleryd:
Yes but then they would get re-rendered and once that happens they block
zoom. With a new key every time it works

On Oct 7, 2016 10:47 AM, "Denis" <notifications@github.com> wrote:

@carleryd <https://github.com/carleryd> btw, you can get uniq key like this

render() {
    markers.map((marker, index) => <MapView.Marker key={index} />
}

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/airbnb/react-native-maps/pull/623#issuecomment-252185793>,
or mute the thread

carleryd:
Any update on this?
I`m currently familiarizing myself with the marker code but have no previous iOS native development, so if anyone wanna collaborate on this, reach me on c.arleryd@gmail.com :)

carleryd:
I managed to get it working @sjmueller !
I am using this PR and instead of using onPress in my RN code, I use onSelect, which I found out was the only way to get a basic annotation to respond to click as well(I assume this has something to do with the hittest now returning MKNewAnnotationContainerView?).
The drawback of this is that I can no longer interact with images inside of the marker, but this is far better than not being able to pinch to zoom!

Exilz:
@sjmueller I can confirm your PR works like a charm  ! It solves this huge UX problem nicely.
I didn't notice any drawback in my application.
I hope this gets merged soon.

Thank you !

tinygipxy:
I got same issue with onSelect, it only work for first select then must select other

Exilz:
@tirnovali @sjmueller 
I just tried to apply these changes on the latest version of this repo and it doesn't work anymore.

I can't zoom and drag over images like expected, but now I can't click at all on my markers.
`onPress` is never called. It used to work once or twice before not firing it again, as I said earlier.

henrikra:
@sjmueller Is this pr still relevant?

Exilz:
@henrikra yes, the problem is still there on the latest released version.

compojoom:
I just tested this patch and I think that we need another go at this. We shouldn't merge this. 

Applying the patch leads to a lot of problems the biggest one being **onPress can't be used anymore**.

If one switches to onSelect, then the map seems at first to behave as expected but any button inside the marker can't respond to touch events. This basically means that we can use this only for the most simple of markers, but anything that needs user interaction isn't going to work.

rborn:
I'll close this for being very old and based on @compojoom comments. Please ping me if you consider it should re-opened after a full sync with the current stage of the module.
Thank you for you effort 😻

PR #625, opened 2016-09-28
Added support for taking snapshots on Android

IjzerenHein:
This pull request adds support for `takeSnapshot` on Android.

The takeSnapshot function was purposefully implemented on the AirMapModule class and not as a command on AirMapManager. The reason was that with using commands it is not possible to call the callback that is provided. The callback cannot be extracted from `ReadableMap`. This made it unsuitable for this specific function. I've taking inspiration from `react-native-view-shot` and was able to implement the functionality on the AirMapModule instead.

The region argument is not used on Android and the width and height are interpreted differently. This is because the `GoogleMaps.snapshot` doesn't take the size or region as input. This creates a feature difference with the iOS version.

Additionally, I've made `takeSnapshot` promise aware. This means it now returns a promise with the resulting snapshot value (the uri) (or rejects the promise in case of an error).

Android screenshot:
image

### DIFF
--- a/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
+++ b/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
@@ -1,0 +1,126 @@
+package com.airbnb.android.react.maps;
+
+import android.app.Activity;
+import android.util.DisplayMetrics;
+import android.util.Base64;
+import android.graphics.Bitmap;
+import android.net.Uri;
+import android.view.View;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.OutputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.ReactContextBaseJavaModule;
+import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.bridge.Promise;
+import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.uimanager.UIManagerModule;
+import com.facebook.react.uimanager.UIBlock;
+import com.facebook.react.uimanager.NativeViewHierarchyManager;
+
+import com.google.android.gms.maps.GoogleMap;
+
+
+public class AirMapModule extends ReactContextBaseJavaModule {
+
+    public AirMapModule(ReactApplicationContext reactContext) {
+        super(reactContext);
+    }
+
+    @Override
+    public String getName() {
+        return "AirMapModule";
+    }
+
+    public Activity getActivity() {
+        return getCurrentActivity();
+    }
+
+    @ReactMethod
+    public void takeSnapshot(final int tag, final ReadableMap options, final Promise promise) {
+
+        // Parse and verity options
+        final ReactApplicationContext context = getReactApplicationContext();
+        final String format = options.hasKey("format") ? options.getString("format") : "png";
+        final Bitmap.CompressFormat compressFormat =
+            format.equals("png") ? Bitmap.CompressFormat.PNG :
+            format.equals("jpg") ? Bitmap.CompressFormat.JPEG : null;
+        if (compressFormat == null) {
+            promise.reject("AIRMap.takeSnapshot", "Unsupported image format: " + format + ". Try one of: png | jpg | jpeg");
+            return;
+        }
+        final double quality = options.hasKey("quality") ? options.getDouble("quality") : 1.0;
+        final DisplayMetrics displayMetrics = context.getResources().getDisplayMetrics();
+        final Integer width = options.hasKey("width") ? (int)(displayMetrics.density * options.getDouble("width")) : 0;
+        final Integer height = options.hasKey("height") ? (int)(displayMetrics.density * options.getDouble("height")) : 0;
+        final String result = options.hasKey("result") ? options.getString("result") : "file";
+
+        // Add UI-block so we can get a valid reference to the map-view
+        UIManagerModule uiManager = context.getNativeModule(UIManagerModule.class);
+        uiManager.addUIBlock(new UIBlock() {
+            public void execute (NativeViewHierarchyManager nvhm) {
+                AirMapView view = (AirMapView) nvhm.resolveView(tag);
+                if (view == null) {
+                    promise.reject("AirMapView not found");
+                    return;
+                }
+                if (view.map == null) {
+                    promise.reject("AirMapView.map is not valid");
+                    return;
+                }
+                view.map.snapshot(new GoogleMap.SnapshotReadyCallback() {
+                    public void onSnapshotReady(Bitmap snapshot) {
+
+                        // Convert image to requested width/height if neccesary
+                        if (snapshot != null && width != 0 && height != 0 && (width != snapshot.getWidth() || height != snapshot.getHeight())) {
+                            snapshot = Bitmap.createScaledBitmap(snapshot, width, height, true);
+                        }
+                        if (snapshot == null) {
+                            promise.reject("Failed to generate bitmap");
+                            return;
+                        }
+
+                        // Save the snapshot to disk
+                        OutputStream outputStream = null;
+                        try {
+                            if ("file".equals(result)) {
+                                File tempFile = File.createTempFile("AirMapSnapshot", "." + format, context.getCacheDir());
+                                outputStream = new FileOutputStream(tempFile);
+                                snapshot.compress(compressFormat, (int)(100.0 * quality), outputStream);
+                                outputStream.close();
+                                outputStream = null;
+                                String uri = Uri.fromFile(tempFile).toString();
+                                promise.resolve(uri);
+                            }
+                            else if ("base64".equals(result)) {
+                                outputStream = new ByteArrayOutputStream();
+                                snapshot.compress(compressFormat, (int)(100.0 * quality), outputStream);
+                                outputStream.close();
+                                outputStream = null;
+                                byte[] bytes = ((ByteArrayOutputStream) outputStream).toByteArray();
+                                String data = Base64.encodeToString(bytes, Base64.NO_WRAP);
+                                promise.resolve(data);
+                            }
+                        }
+                        catch (Exception e) {
+                            promise.reject(e);
+                        }
+                        finally {
+                            if (outputStream != null) {
+                                try {
+                                    outputStream.close();
+                                } catch (IOException e) {
+                                    e.printStackTrace();
+                                }
+                            }
+                        }
+                    }
+                });
+            }
+        });
+    }
+}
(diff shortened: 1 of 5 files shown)


IjzerenHein:
Additionally, I'd like to make a proposal for a new and improved `takeSnapshot` API. The API is inspired by that `react-native-view-shot` but shall still be compatible with the old `takeSnapshot` function. When only 1 argument is specified, the function uses the **new** API:

```js
const promise = map.takeSnapshot({
  width: 100, // optional, when omitted the view width is used
  height: 100, // optional, when omitted the view height is used
  result: 'file', // values: 'file', 'base64' (default: 'file')
  format: 'png', // file-format: 'png', 'jpg' (default: 'png')
  quality: 0.8, // compression-quality (only relevant for jpg) (default: 1.0)
});
```

The result of the promise is a file-uri when `format='file'` or a base-64 encoded string
when `format='base64'`.
Region has been removed as it cannot be implemented on Android. On both iOS and Android the currently selected region shall be used.


And for backwards compatibility the function can be used the old way:

```js
map.takeSnapshot(300, 200, region, (err, result) => {
  ...
});
```

I'd like to hear your thoughts on this.

cheers, Hein

android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
```diff
@@ -0,0 +1,126 @@
...
+                        }
+                        if (snapshot == null) {
+                            promise.reject("Failed to generate bitmap");
+                            return;
+                        }
+
+                        // Save the snapshot to disk
+                        OutputStream outputStream = null;
+                        try {
+                            if ("file".equals(result)) {
```
spikebrehm: Perhaps we can pull these strings out to CONSTANTs?

components/MapView.js
```diff
@@ -446,8 +446,45 @@ class MapView extends React.Component {
...
 
   takeSnapshot(width, height, region, callback) {
-    const finalRegion = region || this.props.region || this.props.initialRegion;
-    this._runCommand('takeSnapshot', [width, height, finalRegion, callback]);
+    return new Promise((resolve, reject) => {
+      let options;
+      if (typeof width === 'object') {
+        options = width;
+      } else {
+        options = {
```
spikebrehm: Can you add a deprecation warning for this case? That way we can remove the old API in the next breaking release.

spikebrehm:
This is awesome, great!  I'm not super excited about the forking between iOS & Android, but it's opaque to the end developer so that's fine.

I like your new new API.  We can introduce that with deprecation warning and then cut over to the new API in the subsequent breaking release.

IjzerenHein:
@spikebrehm Good calls! I'm gonna make those changes and let you know.

What exactly did you mean by "forking between iOS & Android" ?

android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
IjzerenHein: DONE!

components/MapView.js
IjzerenHein: Done. I've updated the function so that it uses the old API when 4 arguments are specified (and only when on iOS).  In all other cases the new API is used.
The deprecation message is generated using `console.warn`.

IjzerenHein:
Alright, I've implemented the new takeSnapshot API. I've updated the iOS code to support optional `width, height and region` and to support the new `format, quality and result` arguments. There is a special case in the iOS native code to support the legacy API.

The API now looks like this:

```
const snapshot = this.refs.map.takeSnapshot({
    width: 300,      // optional, when omitted the view-width is used
    height: 300,     // optional, when omitted the view-height is used
    region: {..},    // iOS only, optional region to render: {latitude, longitude, latitudeDelta, longitudeDelta}
    format: 'png',   // image formats: 'png', 'jpg' (default: 'png')
    quality: 0.8,    // image quality: 0..1 (only relevant for jpg, default: 1)
    result: 'file'   // result types: 'file', 'base64' (default: 'file')
});
```

christopherdro:
Looks great! Should we try and get this merged in for the next release ?

IjzerenHein:
Any chance we can get this in for the next release?
If there's anything I need to change for that to happen let me know.

android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
```diff
@@ -0,0 +1,126 @@
...
+                        }
+                        catch (Exception e) {
+                            promise.reject(e);
+                        }
+                        finally {
+                            if (outputStream != null) {
+                                try {
+                                    outputStream.close();
+                                } catch (IOException e) {
+                                    e.printStackTrace();
```
felipecsl: remove this line, just ignore the exception

android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
```diff
@@ -0,0 +1,126 @@
...
+                                promise.resolve(uri);
+                            }
+                            else if (result.equals(SNAPSHOT_RESULT_BASE64)) {
+                                outputStream = new ByteArrayOutputStream();
+                                snapshot.compress(compressFormat, (int)(100.0 * quality), outputStream);
+                                outputStream.close();
+                                outputStream = null;
+                                byte[] bytes = ((ByteArrayOutputStream) outputStream).toByteArray();
+                                String data = Base64.encodeToString(bytes, Base64.NO_WRAP);
+                                promise.resolve(data);
```
felipecsl: it would be nice if you could break down this logic into smaller methods. maybe `takeBase64Snapshot()` and `takeFileSnapshot()`?

android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
```diff
@@ -0,0 +1,126 @@
...
+                if (view == null) {
+                    promise.reject("AirMapView not found");
+                    return;
+                }
+                if (view.map == null) {
+                    promise.reject("AirMapView.map is not valid");
+                    return;
+                }
+                view.map.snapshot(new GoogleMap.SnapshotReadyCallback() {
+                    public void onSnapshotReady(Bitmap snapshot) {
```
felipecsl: annotate with `@Nullable`

android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
```diff
@@ -0,0 +1,126 @@
...
+
+                        // Convert image to requested width/height if neccesary
+                        if (snapshot != null && width != 0 && height != 0 && (width != snapshot.getWidth() || height != snapshot.getHeight())) {
+                            snapshot = Bitmap.createScaledBitmap(snapshot, width, height, true);
+                        }
+                        if (snapshot == null) {
```
felipecsl: invert the order of these 2 `if`s

android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
```diff
@@ -0,0 +1,126 @@
...
+                        }
+
+                        // Save the snapshot to disk
+                        OutputStream outputStream = null;
+                        try {
+                            if (result.equals(SNAPSHOT_RESULT_FILE)) {
+                                File tempFile = File.createTempFile("AirMapSnapshot", "." + format, context.getCacheDir());
+                                outputStream = new FileOutputStream(tempFile);
+                                snapshot.compress(compressFormat, (int)(100.0 * quality), outputStream);
+                                outputStream.close();
```
felipecsl: `close()` can throw `IOException`.  A common pattern is to create a method `closeQuietly()` that looks like this:
```java
    public static void closeQuietly(Closeable closeable) {
        if (closeable == null) return;
        try {
            closeable.close();
        } catch (IOException ignored) {
        }
    }
```

android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
felipecsl: you could use `closeQuietly` here too

android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
```diff
@@ -0,0 +1,126 @@
...
+                                outputStream = new ByteArrayOutputStream();
+                                snapshot.compress(compressFormat, (int)(100.0 * quality), outputStream);
+                                outputStream.close();
+                                outputStream = null;
+                                byte[] bytes = ((ByteArrayOutputStream) outputStream).toByteArray();
+                                String data = Base64.encodeToString(bytes, Base64.NO_WRAP);
+                                promise.resolve(data);
+                            }
+                        }
+                        catch (Exception e) {
```
felipecsl: what exceptions are you catching exactly? can you be more specific or narrow down its scope?

android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
IjzerenHein: Thanks so much for all the feedback @felipecsl . I will update the code accordingly and re-test and the update the PR. Hopefully this week, but probably next week.
Cheers, Hein

IjzerenHein:
Hi, it look a little longer but I've updated the code and addressed all the issues you pointed out.
Hope this gets in the next release.
regards, Hein

spikebrehm:
Thanks, will take a look soon.

IjzerenHein:
Awesome, thanks @spikebrehm

IjzerenHein:
@spikebrehm Any idea when the next release will be?
Cheers, Hein

components/MapView.js
```diff
@@ -445,9 +445,79 @@ class MapView extends React.Component {
...
+   * Takes a snapshot of the map and saves it to a picture
+   * file or returns the image as a base64 encoded string.
+   *
+   * @param config Configuration options
+   * @param [config.width] Width of the rendered map-view (when omitted actual view width is used).
+   * @param [config.height] Height of the rendered map-view (when omitted actual height is used).
+   * @param [config.region] Region to render (Only supported on iOS).
+   * @param [config.format] Encoding format ('png', 'jpg') (default: 'png').
+   * @param [config.quality] Compression quality (only used for jpg) (default: 1.0).
+   * @param [config.result] Result format ('file', 'base64') (default: 'file').
```
spikebrehm: Maybe add `legacy` here for now?

components/MapView.js
```diff
@@ -445,9 +445,79 @@ class MapView extends React.Component {
...
+    // Sanitize inputs
+    const config = {
+      width: args.width || 0,
+      height: args.height || 0,
+      region: args.region || {},
+      format: args.format || 'png',
+      quality: args.quality || 1.0,
+      result: args.result || 'file',
+    };
+    if ((config.format !== 'png') &&
```
spikebrehm: Consider pulling these out to constants at the top of the file, like:

```js
const VALID_FORMATS = ['png', 'jpg'];
const VALID_RESULTS = ['file', 'base64'];

// ...

if (!VALID_FORMATS.include(config.format)) {
  throw new Error('Invalid format specified');
}
```

Also it may be nice to use the [`invariant`](https://www.npmjs.com/package/invariant) package like:

```js
invariant(VALID_FORMATS.include(config.format), 'Invalid format specified');
```

spikebrehm:
Published as [v0.12.2](https://github.com/airbnb/react-native-maps/releases/tag/v0.12.2)

IjzerenHein:
Awesome, thanks so much @spikebrehm !

PR #628, opened 2016-09-29
Add clarification on initialRegion behaviour

haikyuu:
**Use case**: specifying an `initialRegion` when `fitToSuppliedMarkers` is used in `ComponentDidMount`


haikyuu:
Actually, using `fitTiSuppliedMarkers` should NOT be used in `ComponentDidMount` (unless put in a setTimeout) as it will **[ruin the performance](https://facebook.github.io/react-native/docs/performance.html#javascript-frame-rate)** (blocked the UI for ~8seconds in my case !!)

christopherdro:
Thank you!

npineda:
@haikyuu won't ``fitToCoordinates``all have issues if it is put inside the ``componentDidMount`` method with being put in a ``setTimeout`` function? I get this error without putting the timer:

``Attempt to invoke virtual method 'void com.google.android.gms.maps.GoogleMap.setPadding(int,int,int,int)' on a null object reference``

Last call in stack my stack was:

``fitToCoordinates
AirMapView.java:539``

The way the docs read it seems like using ``setTimeout`` only applies to``fitToSuppliedMarkers`` so I wasn't sure if ``fitToCoordinates`` was bugging out for similiar reasons.

haikyuu:
I think this is a separate issue, no idea why it's happening !
Try to fill an issue and provide more context.
Good luck. 👍

PR #631, opened 2016-09-30
[google][ios] Make provider="google" use same region format as MapKit

gilbox:
This effects events on iOS: the `onRegionChange` event as well as `onRegionChangeComplete` (the latter is currently only available with google).

gb-region

@spikebrehm @lelandrichardson

### DIFF
--- a/ios/AirGoogleMaps/AIRGoogleMap.m
+++ b/ios/AirGoogleMaps/AIRGoogleMap.m
@@ -12,13 +12,36 @@
 #import "RCTConvert+MapKit.h"
 #import "UIView+React.h"
 
-id cameraPositionAsJSON(GMSCameraPosition *position) {
-  // todo: convert zoom to delta lat/lng
+id regionAsJSON(MKCoordinateRegion region) {
   return @{
-           @"latitude": [NSNumber numberWithDouble:position.target.latitude],
-           @"longitude": [NSNumber numberWithDouble:position.target.longitude],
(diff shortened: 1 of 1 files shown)


ios/AirGoogleMaps/AIRGoogleMap.m
```diff
@@ -12,15 +12,38 @@
 #import "RCTConvert+MapKit.h"
 #import "UIView+React.h"
 
-id cameraPositionAsJSON(GMSCameraPosition *position) {
-  // todo: convert zoom to delta lat/lng
+id regionAsJSON(MKCoordinateRegion region) {
   return @{
-           @"latitude": [NSNumber numberWithDouble:position.target.latitude],
-           @"longitude": [NSNumber numberWithDouble:position.target.longitude],
-           @"zoom": [NSNumber numberWithDouble:position.zoom],
```
spikebrehm: Is this a breaking change, because `zoom` is no longer passed through?

spikebrehm:
Nice, seems legit 👍

ios/AirGoogleMaps/AIRGoogleMap.m
spikebrehm: Although, this was released in an rc.0, so we can make a "breaking" change still.

gilbox: yep ( ◠‿◠ )

PR #632, opened 2016-09-30
[google][ios] fix markers never going away for provider="google"

gilbox:
Once created, markers were never going away. This PR fixes that. It also adds the ability to test this functionality in "Default Markers" demo.

gb-markers

@spikebrehm @lelandrichardson 
@felipecsl


spikebrehm:
👍

PR #649, opened 2016-10-05
[google] refactor AIRGoogleMapMarker to use DummyView

gilbox:
- DummyView is used to create virtual tree
- Add CustomMarkers example
- Prevent annoying errors from showing up with OS_ACTIVITY_MODE disable, see here https://github.com/facebook/react-native/issues/9921#issuecomment-250078139

Adding children to markers is still buggy, but this PR at least improves the situation when adding an `image` prop to the marker.

@spikebrehm @lelandrichardson


spikebrehm:
Seems legit.

spikebrehm:
LGTM

PR #664, opened 2016-10-07
If we've disabled scrolling within the map, then don't capture the touch events

mikelambert:
This allows the containing scrollview to perform a scrollview scroll by dragging on the map. (Previously, the map would absorb the touches, and then not scroll the map _or_ the scrollview, which was bad.)

### DIFF
--- a/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
+++ b/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
@@ -578,13 +578,16 @@ public class AirMapView extends MapView implements GoogleMap.InfoWindowAdapter,
 
         switch (action) {
             case (MotionEvent.ACTION_DOWN):
-                this.getParent().requestDisallowInterceptTouchEvent(true);
+                if (map != null && map.getUiSettings().isScrollGesturesEnabled()) {
+                    this.getParent().requestDisallowInterceptTouchEvent(true);
+                }
                 isTouchDown = true;
                 break;
             case (MotionEvent.ACTION_MOVE):
                 startMonitoringRegion();
                 break;
             case (MotionEvent.ACTION_UP):
+                // Clear this regardless, since isScrollGesturesEnabled() may have been updated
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
```diff
@@ -578,13 +578,16 @@ public boolean dispatchTouchEvent(MotionEvent ev) {
 
         switch (action) {
             case (MotionEvent.ACTION_DOWN):
-                this.getParent().requestDisallowInterceptTouchEvent(true);
+                if (map != null && map.getUiSettings().isScrollGesturesEnabled()) {
+                    this.getParent().requestDisallowInterceptTouchEvent(true);
```
gilbox: @mikelambert All I did was checkout your branch and change `true` to `false, and then the included **Static Map** demo worked as I expected. Did you try testing the demo? On iOS it allows you to drag over the map and it scrolls the ScrollView. In android, before I made this change it would drag the map, *not* the ScrollView. After the change, it would scroll the ScrollView with the map remaining unchanged.

mikelambert: So you're saying before this change, Android and iOS had different behavior?

If I look at the code for "Static Map", I see <MapView scrollEnabled={false} ... >. By that logic, we shouldn't scroll inside the map at all, and shouldn't capture those drags, but should instead let them scroll the ScrollView.

Which is what you describe as happening on iOS, and on Android after my change is applied. So I'm confused what the problem is?

mikelambert: Oh wait, I may have misread your message. I'm going to build and run Static Map now.

gilbox: quick update: this pr definitely hasn't been forgotten. The current situation is causing a bug in our app, but I haven't had time to loop back to it yet.

felipecsl:
thanks @mikelambert!

PR #676, opened 2016-10-11
[android] Add parameter to disable the moving on marker press

mlanter:
On iOS (Apple Maps) clicking a marker doesn't move the map but on Android clicking a marker causes the map to move.  This PR adds a property that lets that behavior be disabled.

Test plan (on Android):
* Verify the callout still appears when clicking a pin
* Verify the map moves if the property isn't specified
* Verify the map doesn't move if the property is set to false.

Before:
before

After:
after

### DIFF
--- a/android/src/main/java/com/airbnb/android/react/maps/AirMapManager.java
+++ b/android/src/main/java/com/airbnb/android/react/maps/AirMapManager.java
@@ -155,27 +155,27 @@ public class AirMapManager extends ViewGroupManager<AirMapView> {
         view.map.getUiSettings().setRotateGesturesEnabled(rotateEnabled);
     }
 
-    @ReactProp(name="cacheEnabled", defaultBoolean = false)
+    @ReactProp(name = "cacheEnabled", defaultBoolean = false)
     public void setCacheEnabled(AirMapView view, boolean cacheEnabled) {
         view.setCacheEnabled(cacheEnabled);
     }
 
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/airbnb/android/react/maps/AirMapManager.java
```diff
@@ -165,6 +165,11 @@ public void setLoadingEnabled(AirMapView view, boolean loadingEnabled) {
         view.enableMapLoading(loadingEnabled);
     }
 
+    @ReactProp(name="moveOnMarkerPress", defaultBoolean = true)
+    public void setMoveOnMarkerPress(AirMapView view, boolean moveOnPress) {
```
christopherdro: space between `=`

christopherdro:
LGTM! Small nit pick comment

android/src/main/java/com/airbnb/android/react/maps/AirMapManager.java
mlanter: Just to clarify, should I add spaces around the `=` or remove it?  The other methods in the file use both standards.

christopherdro:
@spikebrehm @lelandrichardson LGTM! 

Thoughts on getting this is?

spikebrehm:
Beautiful, thank you!

spikebrehm:
LMK if you need me to cut a new release for this.

christopherdro:
Would be great with the commit for react native 0.35.

PR #696, opened 2016-10-15
Update installation.md

securingsincity:


christopherdro:
Hey @securingsincity! Thanks for the PR.

I actually was looking at this the other day and am thinking it will be a good idea to add the documentation. 

> Note: As shown above, com.google.android.geo.API_KEY is the recommended metadata name for the API key. A key with this name can be used to authenticate to multiple Google Maps-based APIs on the Android platform, including the Google Maps Android API. For backwards compatibility, the API also supports the name com.google.android.maps.v2.API_KEY. This legacy name allows authentication to the Android Maps API v2 only. An application can specify only one of the API key metadata names. If both are specified, the API throws an exception.
Source: https://developers.google.com/maps/documentation/android-api/signup

Is there a chance you can update this PR to take the note above into account?

There are a few other notes we could add about setting up restricted usage where the apps package name and SHA-1 signing-certificate fingerprint need to be added to the list.

I can update the docs with this info after we get this merged in.

christopherdro:
Thanks!

PR #708, opened 2016-10-19
Support iOS SDK < 10 ( XCode < 8 )

### LINKED ISSUE #629: Xcode 7.3.1 Compatibility (or not?)
When attempting to build a project containing `react-native-maps` at the latest commit in `master`, [`be2e6b3`](https://github.com/airbnb/react-native-maps/commit/be2e6b32b3e4cb87b124a9d78de402fe6b725f34), for iOS in Xcode 7.3.1, the build fails with the following error:

`Cannot find protocol declaration for 'CAAnimationDelegate'`

As I [understand it](http://stackoverflow.com/questions/38972027/cannot-find-protocol-declaration-for-caanimationdelegate), this is because `CAAnimationDelegate` was added to the iOS 10 SDK, which is only available in Xcode 8. However, as of the current version of `react-native`, `0.34`, Xcode 7.3.1 can still be used to build to iOS, though I am unsure if this is officially supported.

I don't know if it's worth thinking about Xcode 7.x support, or maintaining Xcode 7.x support in `react-native-maps`, since it is an old version, but I thought it would be good to discuss it, so we can clarify this in the docs and avoid any confusion.

rops:
Fixes #629


spikebrehm:
LGTM

rops:
Ops forgot to squash commits :/ Aren't you guys using Merge&Squash?

PR #721, opened 2016-10-23
[iOS]  Added showsUserLocation property support for Google Maps

julien-rodrigues:
Hi,

I needed to have the position of the user displayed on the map. And I saw the support on GoogleMaps for iOS was not there.

I added the support of the prop `showsUserLocation` for GoogleMaps on iOS.

The code just sets the option `myLocationEnabled` on the map.

Documentation for this is available here: https://developers.google.com/maps/documentation/ios-sdk/map#my_location


julien-rodrigues:
Thanks :)

PR #728, opened 2016-10-25
[marker flicker]  Fix flicker of map pins on state change

mlanter:
Fixes https://github.com/airbnb/react-native-maps/issues/683 caused by https://github.com/airbnb/react-native-maps/pull/548.

In #548, MapMarker.js was changed from 
`const AIRMapMarker = requireNativeComponent('AIRMapMarker', MapMarker);` to 
`const AIRMapMarker = this.getAirComponent();`.  

`AirMapMarker` is what is returned in the render method.

The bug was that `this.getAirComponent()` was returning a new native component each time causing a re-render.  This only happened for the default component, not the other ones since they were cached.  The fix is to cache the default component.

Note: this seems to affect map markers with arbitrary views inside them the most.  

Before:
before 2

After:
after

### DIFF
--- a/components/decorateMapComponent.js
+++ b/components/decorateMapComponent.js
@@ -48,7 +48,7 @@ export default function decorateMapComponent(Component, { componentType, provide
     if (components[provider]) return components[provider];
 
     if (provider === PROVIDER_DEFAULT) {
-      components.default = getDefaultComponent();
+      if (!components.default) components.default = getDefaultComponent();
       return components.default;
     }
 
--- a/components/decorateMapComponent.js
+++ b/components/decorateMapComponent.js
@@ -48,8 +48,8 @@ export default function decorateMapComponent(Component, { componentType, provide
     if (components[provider]) return components[provider];
 
     if (provider === PROVIDER_DEFAULT) {
-      if (!components.default) components.default = getDefaultComponent();
-      return components.default;
+      components[PROVIDER_DEFAULT] = getDefaultComponent();
+      return components[PROVIDER_DEFAULT];
     }
 
     const providerInfo = providers[provider];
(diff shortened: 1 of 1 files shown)


mlanter:
@spikebrehm @lelandrichardson @gilbox

components/decorateMapComponent.js
```diff
@@ -48,7 +48,7 @@ export default function decorateMapComponent(Component, { componentType, provide
     if (components[provider]) return components[provider];
 
     if (provider === PROVIDER_DEFAULT) {
-      components.default = getDefaultComponent();
+      if (!components.default) components.default = getDefaultComponent();
```
gilbox: thanks for catching this! I wonder if it would be better to do:

```
components[DEFAULT_PROVIDER] = getDefaultComponent();
```

instead.... It's just a little weird because the value of `DEFAULT_PROVIDER` is `null`.

mlanter: Yeah that's a bit clearer (and can use the existing cache return at the beginning of the function).

components/decorateMapComponent.js
```diff
@@ -48,8 +48,8 @@ export default function decorateMapComponent(Component, { componentType, provide
     if (components[provider]) return components[provider];
 
     if (provider === PROVIDER_DEFAULT) {
-      components.default = getDefaultComponent();
-      return components.default;
+      components[PROVIDER_DEFAULT] = getDefaultComponent();
```
mlanter: Don't need to check if already set, because it was then it would have been returned earlier.

mlanter:
@gilbox

gilbox:
domo arigatou!

max-mykhailenko:
When do you plan to release this?

max-mykhailenko:
@gilbox How I can help you with release?

alvelig:
Looks like does not happen in recent versions (0.19.0) Let us know if still persists.

hugoh1995:
Still having this flickering bug.

EhsanSarshar:
still bug exist in version 0.27.1

PR #547, opened 2016-09-07
Use latest version of RN to build Android

fmatoss:
Currently it's not possible to build an Android app when using RN 0.32.1 because it's looking for version 0.32.0, this commit fixes this issue.

PR #731, opened 2016-10-26
Use latest Google Play Services

mlanter:
Switch to use the latest Google Play Services.  

The pain point is that by specifying a specific version it doesn't work with other libraries which specify the latest version of Google Play Services (which most do).  It causes a crash because when it is compiled it will include some of the latest version and some of 9.4.0 and then when it runs it crashes.

It's generally safe to assume users have the latest Google Play Services as the updates are pushed automatically.  As Google says "Google Play services gives you the freedom to use the newest APIs for popular Google services without worrying about device support. Updates to Google Play services are distributed automatically by the Google Play Store and new versions of the client library are delivered through the Android SDK Manager."

This matches a similar decision that was made to use `react-native:+` in #547

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -53,7 +53,7 @@ android {
 }
 
 dependencies {
-  compile 'com.facebook.react:react-native:+'
-  compile "com.google.android.gms:play-services-base:9.4.0"
-  compile 'com.google.android.gms:play-services-maps:9.4.0'
+  compile "com.facebook.react:react-native:+"
+  compile "com.google.android.gms:play-services-base:+"
+  compile "com.google.android.gms:play-services-maps:+"
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -54,6 +54,6 @@ android {
 
 dependencies {
   compile "com.facebook.react:react-native:+"
-  compile "com.google.android.gms:play-services-base:+"
-  compile "com.google.android.gms:play-services-maps:+"
+  compile "com.google.android.gms:play-services-base:9.6.1"
+  compile "com.google.android.gms:play-services-maps:9.6.1"
 }


alexHlebnikov:
If newest version of GPS will be not installed on user's device, the app will crash when user will try to interact with map.

mlanter:
It's a tradeoff between crashing for users that don't have the latest GPS versus crashing for projects that use newer versions of GPS.  

I'd lean towards requiring the latest GPS, because

- Google says it is safe to assume the latest Google Play Services
- The react-native-maps project has tended to error on the side of supporting only the latest versions of dependencies given the rapid change of pace (based on the Compatibility part of the readme)
- We'll be able co-exist with other react-native libraries, most of which require the latest GPS.
- Last time this was discussed in #533, the decision was to support the latest GPS with the tradeoff of requiring users have the latest version (which Google says is a safe assumption).

mlanter:
cc: @spikebrehm @felipecsl @lelandrichardson

android/build.gradle
```diff
@@ -53,7 +53,7 @@ android {
 }
 
 dependencies {
-  compile 'com.facebook.react:react-native:+'
-  compile "com.google.android.gms:play-services-base:9.4.0"
-  compile 'com.google.android.gms:play-services-maps:9.4.0'
+  compile "com.facebook.react:react-native:+"
+  compile "com.google.android.gms:play-services-base:+"
+  compile "com.google.android.gms:play-services-maps:+"
```
felipecsl: use version `9.6.1` explicitly

mlanter: Just curious, what's the advantage of explicitly setting it versus using + (similar to what's done for react-native)?

Looking at the most popular react-native libraries that use Google Play Services the rest of them use +, which makes it difficult to use them with react-native-maps.

PR #737, opened 2016-10-27
Fix dynamic imageSrc removal

btoueg:
Prevent triggering irrelevant imageLoader call, thus avoiding an exception.


gilbox:
https://github.com/airbnb/react-native-maps/issues/738

PR #750, opened 2016-10-31
[ios][google] implement fitToSuppliedMarkers and fitToCoordinates

gilbox:
@lelandrichardson 

fit


cgpomeroy:
Is this still iOS only? If so, is there a recommended method to produce a similar effect on Android?

PR #752, opened 2016-10-31
Add latest patch releases to the changelog

felipecsl:


gilbox:
lgtm

stereodenis:
@gilbox  @spikebrehm why 0.10.4 goes after 0.11.0?

spikebrehm:
@stereodenis because it was published later, chronologically speaking.

But perhaps we should keep it ordered by version number?

PR #785, opened 2016-11-10
[RN][iOS][google] Set region only when view has width&height

gilbox:
- Fix type issue in AIRMapManager
- Setup Gemfile in example/ios dir to avoid problems with different versions of cocoapods
- Update examples-setup.md to use bundler
- Change MapView so that we only set the native region prop when there is a width and height. GoogleMaps iOS requires the width and height to properly calculate the map zoom level.

@spikebrehm @lelandrichardson


components/MapView.js
```diff
@@ -371,15 +371,6 @@ class MapView extends React.Component {
     return { provider: this.props.provider };
   }
 
-  componentDidMount() {
-    const { region, initialRegion } = this.props;
-    if (region && this.state.isReady) {
-      this.map.setNativeProps({ region });
-    } else if (initialRegion && this.state.isReady) {
```
spikebrehm: Why is this no longer needed?

spikebrehm: Because of the change to `_onLayout()`?

docs/examples-setup.md
```diff
@@ -2,17 +2,24 @@
...
+
+2. Install dependencies and open the workspace:
 
 ```
 cd example
 npm install
 cd ios
-pod install
+bundle install
+bundle exec pod install
```
spikebrehm: Do we really need both?  Can it just be the second?

components/MapView.js
gilbox: For google+ios, this will cause a region calculation that is incorrect. For mapkit+ios it will cause the map to redundantly set the region because it will happen in `_onLayout` anyway.

docs/examples-setup.md
gilbox: @spikebrehm unless there is a way to add a command to run `pod install` from the `Gemfile` I think we need both.

gilbox: oh are you saying that `bundle exec` runs `bundle install` automatically? I don't know much about bundler tbh

spikebrehm: oh no nevermind brainfart

PR #793, opened 2016-11-12
[iOS][Android] Add example and docs for polygon+polyline onPress events

frankrowe:
- Adds a polygon and polyline with `onPress` handlers to the "Events" example
- Fixes an issue where recording of map events was overriding recording of marker/overlay events. `setState` in `recordEvent` now takes a function with previous state to ensure an atomic update occurs.
- Adds docs for polygon + polyline onPress event

screen shot 2016-11-12 at 3 46 36 pm


gilbox:
👏 👏 👏

voidale:
@frankrowe can we get coordinates for this onPress event? it looks like mapPressed isn't being called when clicking on polygons now :(

PR #670, opened 2016-10-10
Added map styling

azt3k:

Issue #781, opened 2016-11-08
MapStyleOptions


kordeviant:
hi , there. as of [september_21_2016](https://developers.google.com/maps/documentation/android-api/releases#september_21_2016) google has made it possible to give styling to mobile maps. please make this possible !!

kordeviant:
can anybody help?

ali-alamine:
I'll be doing this (I need it for a project). But I'm not sure about the naming since I don't want to confuse it with the react 'style' option. Any suggestions?? I'm thinking about about tilesStyle.

gilbox:
This might be related

ali-alamine:
Yes it does solve this.

gilbox:
@ali-alamine if you'd like to assist by breaking up that code into smaller PRs, it would be very helpful!

ali-alamine:
@gilbox I already am, just need some time cz I have to finish a few things first. By monday it will be done.

PR #808, opened 2016-11-17
Map Styling for android

ali-alamine:
This pull request is to separate the styling into a separate pull request from #670.

I don't have access to a mac now so I'll create another pull request for ios by monday. But this code should compile normally on ios, but the map won't be styled.


ali-alamine:
ref #781

PR #814, opened 2016-11-21
[android] Create example for marker not updating bug

gilbox:
Created an example to expose this android bug.

rn-maps-heading-bug-example


spikebrehm:
Is this still relevant with latest changes?  If so let's get this rebased.

Saliheen:
I solved it like this  n                                                                                                                                                                                                                                            . const [loading, setLoading] = useState(false);

  const handleOnLayout = () => {
    setLoading(true);
  };

  return (
    <Marker
      tracksViewChanges={false}
      coordinate={coords}>
      <Image
        source={imageurl}
        onLayout={handleOnLayout}
        key={`key-${loading}`}
      />
    </Marker>
  );

PR #840, opened 2016-11-29
Fixes #470. Support legalLabelInsets on Apple Maps

### LINKED ISSUE #470: Is it possible to change  the position of legalLabelInsets on the bottom left corner of the map, and also changing the location of the button for centering the map to user current Location?

scarlac:
Fixes #470. This will add support for `legalLabelInsets` prop on iOS with Apple Maps. Code is adapted from RN core.

### DIFF
--- a/example/App.js
+++ b/example/App.js
@@ -32,6 +32,7 @@ import CustomTiles from './examples/CustomTiles';
 import ZIndexMarkers from './examples/ZIndexMarkers';
 import StaticMap from './examples/StaticMap';
 import MapStyle from './examples/MapStyle';
+import LegalLabel from './examples/LegalLabel';
 
 const IOS = Platform.OS === 'ios';
 const ANDROID = Platform.OS === 'android';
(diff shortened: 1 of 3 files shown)


gilbox:
Thanks! We should probably avoid `->` notation. Would you mind adding an example for this?

scarlac:
@gilbox Yeah, let's keep it consistent. I've removed it.

gilbox:
@scarlac will you be able to add an example?

scarlac:
@gilbox Ah! I am assuming you mean updating the documentation. Yes, I'll get right on it tomorrow and update this PR :)

gilbox:
@scarlac I was talking about adding an example in the `example/examples` dir. That's how I smoke-test PRs since we don't have specs.

scarlac:
@gilbox I've added an example to demonstrate the usefulness of the `legalLabelInsets`. I think it's fairly simple while showing a very realistic scenario of when it could be of use. Personally, I'm using it in an advanced way but opted for simplicity in the example, for the sake of understanding.

Let me know if there's more you need before it's ready for a merging.

gilbox:
lgtm but I'd like to get a second opinion re: the async op

gilbox:
thanks @scarlac ( ◠‿◠ )

scarlac:
My pleasure! 👍

PR #873, opened 2016-12-08
Allow legalLabelInsets to be changed and animated

scarlac:
Added docs for legalLabelInsets as well. Related to #840 

While property is structurally identical to EdgePadding, used for region fitting, they serve completely
different purposes. So a new type is defined.

Default left empty since it's an OS default (around {left: 10, bottom: 10} in my tests)


gilbox:
🏆

spikebrehm:
Published as [v0.12.2](https://github.com/airbnb/react-native-maps/releases/tag/v0.12.2)

scarlac:
@spikebrehm and @gilbox : Thanks for the quick turn-around on these fixes. Really appreciated.

Issue #874, opened 2016-12-08
Updating react-native-maps > 0.8.2 crashes Android


kevinnguy:
When I update `react-native-maps` to `0.9.0` and higher, my Android app crashes at app startup and it doesn't show an error screen. I've been having a difficult time trying to debug why my app on Android is crashing.

I've tested all versions `0.9.0` and higher and `0.8.2` was the version that didn't crash for me. 

My `package.json`:
```
  "dependencies": {
    "firebase": "3.4.0",
    "lodash": "^4.16.1",
    "react": "15.4.1",
    "react-native": "0.39.0",
    "react-native-maps": "0.8.2",
    "react-native-navigation": "2.0.0-experimental.141",
    "react-redux": "^4.4.5",
    "realm": "^0.14.3",
    "redux": "3.6.0",
    "redux-thunk": "^2.1.0"
  },
```

Any suggestions to debugging Android would be appreciated! Thanks!

gilbox:
try `react-native-maps@0.12.1`?

kevinnguy:
@gilbox I've tried `0.12.1` earlier today and it crashed on me :/

gilbox:
@kevinnguy check your `adb logcat`

kevinnguy:
This is the log I get back from `adb logcat` on `"react-native-maps": "0.12.1"`

```
12-08 14:57:18.336   799   817 I ActivityManager: Displayed com.hero/.MainActivity: +349ms
12-08 14:57:18.511  5809  5809 I art     : Thread[1,tid=5809,Native,Thread*=0xb4d36500,peer=0x748372a0,"main"] recursive attempt to load library "/data/app/com.hero-1/lib/arm/libfb.so"
12-08 14:57:18.512  5809  5809 W unknown:React: You seem to be running on device. Run 'adb reverse tcp:8081 tcp:8081' to forward the debug server's port to the device.
12-08 14:57:18.509  1191  1191 W Binder_6: type=1400 audit(0.0:4213): avc: denied { ioctl } for path="socket:[836123]" dev="sockfs" ino=836123 ioctlcmd=7704 scontext=u:r:system_server:s0 tcontext=u:r:system_server:s0 tclass=unix_stream_socket permissive=0
[...]

Issue #831, opened 2016-11-26
Not compatible with react-native-push-notification


pewh:
Hi, my project is use react-native-push-notification and react-native-maps. Previously my project already configured to use react-native-push-notification, and it's work. But when I try insert react-native-maps, I got an error. I have guess something is wrong with my configuration. So I have do some experiment.

First, I fork your repo, and I can run your Example project in my android. But when I try to add react-native-push-notification on your Example project, I got same error.

```
D/dalvikvm( 8529): DexOpt: unable to opt direct call 0x8dfc at 0x7e in Lcom/airbnb/android/react/maps/AirMapManager;.receiveCommand
W/dalvikvm( 8529): Unable to resolve superclass of Lcom/google/android/gms/maps/model/LatLng; (4130)
W/dalvikvm( 8529): Link of class 'Lcom/google/android/gms/maps/model/LatLng;' failed
D/dalvikvm( 8529): DexOpt: unable to opt direct call 0x8def at 0xbc in Lcom/airbnb/android/react/maps/AirMapManager;.receiveCommand
W/dalvikvm( 8529): threadid=12: thread exiting with uncaught exception (group=0xa614d908)
E/AndroidRuntime( 8529): FATAL EXCEPTION: AsyncTask #2
E/AndroidRuntime( 8529): java.lang.RuntimeException: An error occured while executing doInBackground()
E/AndroidRuntime( 8529):        at android.os.AsyncTask$3.done(AsyncTask.java:299)
E/AndroidRuntime( 8529):        at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:352)
E/AndroidRuntime( 8529):        at java.util.concurrent.FutureTask.setException(FutureTask.java:219)
E/AndroidRuntime( 8529):        at java.util.concurrent.FutureTask.run(FutureTask.java:239)
E/AndroidRuntime( 8529):        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1080)
E/AndroidRuntime( 8529):        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:573)
E/AndroidRuntime( 8529):        at java.lang.Thread.run(Thread.java:856)
E/AndroidRuntime( 8529): Caused by: java.lang.NoClassDefFoundError: com.google.android.gms.maps.GoogleMapOptions
[...]

PR #877, opened 2016-12-09
changed google play dependency from 9.8.0 to point to latest version

akshay2604:
I upgraded react native maps and app was crashing on android redmi 3s prime. I changed google play base and maps dependency to '+' and it worked. Also it fixes [#831](https://github.com/airbnb/react-native-maps/issues/831) and [#874](https://github.com/airbnb/react-native-maps/issues/874)

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -54,6 +54,6 @@ android {
 
 dependencies {
   compile "com.facebook.react:react-native:+"
-  compile "com.google.android.gms:play-services-base:9.8.0"
-  compile "com.google.android.gms:play-services-maps:9.8.0"
+  compile "com.google.android.gms:play-services-base:+"
+  compile "com.google.android.gms:play-services-maps:+"
 }


bfarrgaynor:
This fixes an issue I was having as well, see here: https://github.com/airbnb/react-native-maps/issues/894

jpgarcia:
please merge this one!

ali-alamine:
I'm against such a change. 
Having a precise version is really important for cd and ci because we need the stability and predicatbility that come with specifying a version. Having such a thing will create a layer of abstraction that will be a nightmare to debug if anything goes wrong on the ci server and not on the dev machine. This is especially true if the person is just beginning using the library and doesn't notice this thing. See my solution on #910 that can solve this issue, and still have a specific version.

kiranraj2k:
@akshay2604 solution worked for me too!

ali-alamine:
@kiranraj2k This solution does work, but here is where it might fail and be hard to find:

Consider that you have google-services v. 10.0.1 and your friend has google-services v. 9.8.0. If something works with you on 10.0.1 it might not work on 9.8.0, and since this solution just uses the latest version installed, it will be really hard to find out the reason if anything like this happens (you saw how ambiguous the error is with such things).

This is really important if you have a bot that is responsible of automatically building the app and publishing it for alpha testing, especially since these will not check for newer versions of play-services if you're using the '+' for performance reasons and will just use the installed one if available.


If you want to use a version other than the one included with the react-native-maps, you can override it in your gradle build (see #910 for how to do that)

lucasbento:
@akshay2604: thanks you for opening this PR, had the exact same issue!

kanekotic:
thanks for this PR. had also n issue related to this. Where if you add any  play services dependency. Maps would make my app not start.
My 2 cents on why  the why i prefer not force a specific version is because normally this module is not the only one added and if each starts requesting different random versions of play services it might get more difficult to debug (also normally new versions should be backwards compatible).
Also for what @alialamine said, i think its sensible concern, and should be added as a note in the README or documentation as the recommended version and help to troubleshoot.

sandman45:
Thanks! this worked for me as well.. we upgraded from 0.39 to 0.41 was getting the same error and made the change.. we are good to go!

antwan2093:
This is breaking builds again, is there a reason why you don't simply use '+' for the gms deps?

Issue #899, opened 2016-12-15
Option 2: CocoaPods -- Untested Way -- Works


vjeranc:
Hello, I've been trying to build stuff without manual intervention.

Using default `app.xcodeproj/project.pbxproj` file generated by `react-native upgrade`.

Adjusted Podfile from the AIR example:
```
# You Podfile should look similar to this file. Whether you use_frameworks! or not, the following configuration should work.
#
# However if you DO NOT use_frameworks! and you prefer to install pods instead of
# dragging the AirMaps directory into your project, refer to the comments below (steps 1~4)

source 'https://github.com/CocoaPods/Specs.git'

platform :ios, '8.0'
# use_frameworks!

# Change 'extraSpecialCoolAppName' to match the target in your Xcode project.
target 'extraSpecialCoolAppName' do
  pod 'React', path: '../node_modules/react-native', :subspecs => [
    'Core',
    'RCTActionSheet',
    'RCTGeolocation',
    'RCTImage',
    'RCTLinkingIOS',
    'RCTNetwork',
    'RCTSettings',
    'RCTText',
    'RCTVibration',
    'RCTWebSocket'
  ]

  # when not using frameworks  we can remove this line (2/4):
  # pod 'GoogleMaps', '~> 2.1.1'  # <~~ remove this line if you do not want to support GoogleMaps on iOS
  
  # when not using frameworks  we can do this instead of including the source files in our project (1/4):
  pod 'react-native-google-maps', path: '../node_modules/react-native-maps/react-native-google-maps.podspec'  # <~~ if you need GoogleMaps support on iOS
  pod 'react-native-maps', path: '../node_modules/react-native-maps/react-native-maps.podspec'
end

# when not using frameworks this might be necessary, but try without it first and see if `pod install` works (3/4):
# THIS IS ONLY NECESSARY IF YOU NEED GoogleMaps SUPPORT
pre_install do |installer|
  # copied from https://github.com/CocoaPods/CocoaPods/issues/2926
  # workaround for https://github.com/CocoaPods/CocoaPods/issues/3289
  def installer.verify_no_static_framework_transitive_dependencies; end
end

# when not using frameworks (4/4):
# THIS IS ONLY NECESSARY IF YOU NEED GoogleMaps SUPPORT
# #Crud, this shouldn't be necessary, but https://github.com/CocoaPods/CocoaPods/issues/5429
post_install do |installer|
  installer.pods_project.targets.each do |target|
[...]

PR #900, opened 2016-12-15
fixes Option 2 building

vjeranc:
- afaik `GMSMapStyle` is not present in 2.0.x

I've opened issue #899.


spikebrehm:
Released as v0.12.3

PR #901, opened 2016-12-15
[android crash]  Fix exception when animating region during initialization

mlanter:
Fixes a crash cause by calling animateToRegion on the map while it is still be initialized.

To reproduce:  
In the `componentDidMount` of the component that renders the `MapView` add a `    this._map.animateToRegion(region, 1500);`.

Basically on the native side we haven't initialized the `map` variable yet (which happens in `onMapReady`) so if that's the case we can't do an `animateToRegion` yet.

screenshot-2016-12-15_14 08 08 230


mlanter:
cc: @spikebrehm

spikebrehm:
Released as v0.12.3

PR #904, opened 2016-12-17
Updated install documentation after successfully getting iOS and Android working

dboydor:
- Updated descriptions of getting and using Android API Keys.  It works differently than when this was originally written.
- Added fix if DexException encountered
- Added "import" in map component code
- Other small changes


gilbox:
thanks ( ◠‿◠ )

PR #910, opened 2016-12-20
[doc, android] Adding section to fix gradle problems

ali-alamine:
This adds a solution to fix the mismatch between the play-services in case the user wants to use a different version

This fixes #877,  #831 and #874 for all versions while keeping compatibility with projects using older version of play-services

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -73,6 +73,21 @@ After your `Podfile` is setup properly, run `pod install`.
      compile project(':react-native-maps')
    }
    ```
+   
+   If you have a different play serivces than the one included in this library, use the following instead:
+   
+   ```groovy
+   ...
+   dependencies {
+     ...
+     compile project(':react-native-maps'){
+       exclude group: 'com.google.android.gms', module: 'play-services-base'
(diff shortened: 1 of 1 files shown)


Froelund:
I can't get this to work as is.
When I run with the change explained, I get this error.
```
* What went wrong:
A problem occurred evaluating project ':app'.
> Could not find method exclude() for arguments [{group=com.google.android.gms, module=play-services-base}] on project ':react-native-maps'.
```

I've change it to
```
    compile(project(':react-native-maps')){
        exclude group: 'com.google.android.gms', module: 'play-services-base'
        exclude group: 'com.google.android.gms', module: 'play-services-maps'
    }
    compile 'com.google.android.gms:play-services-base:10.0.1'
    compile 'com.google.android.gms:play-services-maps:10.0.1'
```
and it works

ali-alamine:
That's what i meant, to replace the <> with the version number that you need. Have any idea for how to make it less confusing??

ali-alamine:
Nevermind the above, I'll fix the syntax ( and weird cz the version that I have worked for me)

Issue #286, opened 2016-06-09
onPress and onCalloutPress doesn't trigger on markers in iOS


gmarinosyan:
Here is my code:

```
render() {
    return (
      <MapView style={styles.mapView} region={this.state.region}>
        <MapView.Marker
                coordinate={{latitude: 45, longitude: -75}}
                onSelect={this.markerPressed}
         />
      </MapView>
    )
}

markerPressed(){
    console.log("MarkerPressed");
}
```

The function is being called if I use onSelect or onDeselect, but not onPress or onCalloutPress. Is this an issue with something I am doing or a bug?

christopherdro:
@gmarinosyan Is this for iOS or Android. onSelect and onDeselect if specifically for iOS and is what triggers the callout.

gmarinosyan:
It is on iOS. Do the onPress and onCalloutPress functions not work on iOS?

DennisMG:
@I'm experiencing the same problem... Did you solve it @gmarinosyan ??

gmarinosyan:
No, was not able to solve.

christopherdro:
@gmarinosyan Did you guys already try the most recent release 0.6.0?

cjbutcher:
Think I'm having this issue on 0.6.0

klaste:
@christopherdro I have the same problem. It seems as if the _handleTap gesture recognizer delegate method of AIRMapMarkerManager is never called. I am on 0.6.0, React Native 0.29-rc, and iOS 10, so I don't know if this is of any relevance or if I am looking in the right place anyway.

klaste:
@christopherdro I will look into it. Do you have any idea what could be the cause?

nickw:
@christopherdro @lelandrichardson I'm experiencing this issue as well using v0.6.0 and `react-native` v0.28.0. I'm happy to open a PR if anyone can point me in the right direction.

KunwarSingh:
Hey man I'm also facing same issue, onSelect is working but not onCallbackPress. Please try to solve this issue.

DmsChrisPena:
Yup, onPress isn't working at all for me on IOS. Any update on a work around or PR?

KGALLET:
+1 onPress isn't working for me too on iOS.

zelll:
@DmsChrisPena @KGALLET , I found this issue. It's a better place than #403 to answer your question.

See patch below. I'm not a iOS developer. Can someone review it and create a PR? @nickw ?

[...]

PR #954, opened 2017-01-12
Fixed issue #286.

RajkumarPunchh:
#### Description:
The issue relates to [getAnnotationView](https://github.com/airbnb/react-native-maps/blob/master/ios/AirMaps/AIRMapMarker.m#L68) method.
A tap gesture is added to `AirMapMarker` view in `AirMapMarkerManager` class which listens to tap/press events on annotation view as well as callout view. But when `shouldUsePinView` returns `true`, the annotation view on map is a `MKPinAnnotationView` instead of `AirMapMarker` and there is no tap gesture listener on this MKPinAnnotationView.

#### Fix:
Now the Tap gesture is listened by the `AirMapMarker` class itself and gesture now added to MKPinAnnotationView as well.

### DIFF
--- a/ios/AirMaps/AIRMapMarker.h
+++ b/ios/AirMaps/AIRMapMarker.h
@@ -48,6 +48,7 @@
 - (BOOL)shouldShowCalloutView;
 - (void)showCalloutView;
 - (void)hideCalloutView;
+- (void)addTapGestureRecognizer;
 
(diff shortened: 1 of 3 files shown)


spikebrehm:
Thanks 🍻!

willyyang:
when will this be released?

Issue #991, opened 2017-01-23
updating urlTemplate for MapView.URLTile on iOS doesn't update the map


jschloer:
When updating the urlTemplate prop only, the MapView doesn't update. If I remove the child urlTile component and re-add with the new url, it displays fine. 

It appears the cause is that the _map variable is never set on the AIRMapURLTile class, so when the update is called, the _map is always nil and the update call in AIRMapURLTile bails out [here](https://github.com/airbnb/react-native-maps/blob/master/ios/AirMaps/AIRMapUrlTile.m#L36).

I was able to fix by setting the map value in AIRMap.m insertReactSubview similar to what's being done for AIRMapPolyline and AIRMapPolygon, [here](https://github.com/airbnb/react-native-maps/blob/master/ios/AirMaps/AIRMap.m#L99).

I'll be submitting a small PR for this in just a moment. Let me know if there are other implications of this that I am missing.

spikebrehm:
Fixed by #992

PR #992, opened 2017-01-23
Sets the map value for the AirMapUrlTile so that it can be updated properly

jschloer:
This addresses issue #991. This should allow for templateURL updates to display properly.


jschloer:
@spikebrehm, not to be a bother, but do you think this could be merged? The issue breaks our live updating map selector. Let me know if there's anything else needed. I appreciate the consideration.

spikebrehm:
Thanks for the fix! 🍻

PR #1008, opened 2017-02-01
Add id identifier to marker-press event on Android

stan229:
Currently, on iOS the onMarkerPress callback includes the identifier of the map marker as an `id` property. Google Map markers do not store their identifier. This change aligns both platforms.


spikebrehm:
Thanks! 🍻

PR #1013, opened 2017-02-02
Create `update-version` script to more easily update version

spikebrehm:
Similar to `npm version`, but updates the version number in all the
files we need to.

Usage:

    ./scripts/update-version.js [major|minor|patch]

It will:

* Get current npm version.
* Determine what type of version bump: major, minor, patch
* Update the files:
  * package.json
  * android/gradle.properties
  * react-native-maps.podspec
  * react-native-google-maps.podspec
* Make Git commit.
* Create Git tag.

to: @lelandrichardson @gilbox 
cc: @ljharb

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,15 +28,12 @@
   },
   "devDependencies": {
     "babel-eslint": "^6.1.2",
-    "commander": "^2.9.0",
-    "dedent": "^0.7.0",
     "eslint": "^3.3.1",
     "eslint-config-airbnb": "^10.0.1",
     "eslint-plugin-import": "^1.14.0",
     "eslint-plugin-jsx-a11y": "^2.1.0",
     "eslint-plugin-prefer-object-spread": "^1.1.0",
-    "eslint-plugin-react": "^6.1.2",
-    "semver": "^5.3.0"
+    "eslint-plugin-react": "^6.1.2"
   },
   "rnpm": {
     "android": {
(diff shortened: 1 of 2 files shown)


ljharb:
If you add this script under "preversion", "version", or "postversion" then `npm version` should invoke it for you.

scripts/update-version.js
```diff
@@ -0,0 +1,105 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+const semver = require('semver');
+const pkg = require('../package.json');
+
+const versionTypes = ['major', 'minor', 'patch'];
+const filesToUpdate = [
+  'package.json',
+  'android/gradle.properties',
+  'react-native-maps.podspec',
+  'react-native-google-maps.podspec',
```
ljharb: are all three of these always in sync? What happens when there's a bugfix that only applies to one of them?

gilbox:
very cool! 👍

spikebrehm:
ok PTAL Jordan

package.json
```diff
@@ -6,7 +6,8 @@
   "version": "0.13.0",
   "scripts": {
     "start": "react-native start",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "preversion": "./scripts/update-version.js"
```
ljharb: the `.js` isn't necessary since it has a shebang, fwiw

Issue #1003, opened 2017-01-31
fitToCoordinates at initialization of the map


MathieuMailhos:
The example provided in the examples shows a button to execute a function to set the appropriate zoom considering to a list of markers:

```
  fitAllMarkers() {
    this.map.fitToCoordinates(MARKERS, {
      edgePadding: DEFAULT_PADDING,
      animated: true,
    });
}

```

Would it be possible to initialize the map with the appropriate fit given an array of markers already initialized ?

When trying to set the right fit on the componentDidMount, I am receiving:
`
Error using new LatLntBounds(LatLngBounds, int): Map size can't be 0. Most likely, layout has not yet occured for the map view.`

It is definitively too early to call my` fitAllMarkers() `function. Is there a `onLoad` function that can be triggered right after the map was initialized ?

spikebrehm:
Interesting, so `componentDidMount()` is too early then.

Try adding an `onLayout` callback to your `<MapView>` instance and calling `fitAllMarkers()` in `onLayout`.

spikebrehm:
> Would it be possible to initialize the map with the appropriate fit given an array of markers already initialized ?

It is possible to pass in an `initialRegion` prop.  You could use that.

kanekotic:
uhmmm this works for me. I will leave the code i am using for future people that arrive here. I am not using markers as fitToCoordinates is for `latlong` object.

```js
class Map extends Component {

    constructor() {
        super()
        this.mapRef = null;
    }

    componentDidMount() {
        this.mapRef.fitToCoordinates(this.props.myLatLong, { edgePadding: { top: 10, right: 10, bottom: 10, left: 10 }, animated: false })
    }

    render() {
      return    <MapView style={{flex: 1}}
                    ref={(ref) => { this.mapRef = ref }}>
                    <Polyline coordinates={this.props.myLatLong} strokeWidth={4} strokeColor="#2962FF" /> 
                </MapView>
    }
}

```

kanekotic:
the previous code fails on android but not in IOS, so following the what @spikebrehm said, a better solution that work in both is
```js
class Map extends Component {

    constructor() {
        super()
        this.mapRef = null;
    }

    render() {
      return    <MapView style={{flex: 1}}
                    ref={(ref) => { this.mapRef = ref }}
[...]

PR #1037, opened 2017-02-11
Improve documentation for fitToCoordinates

kanekotic:
Added this related to #1003 and probably. It is just a comment on the notes of `fitToCoordinates` related to the when is recommended to call this method. Due a crash on android if called from `ComponentDidMount`. 
This could be a temporal solution, till someone (I will try if i find the time for it) investigate why is it crashing the call in android in this specific case.


christopherdro:
Does it make sense to make this recommendation for iOS as well?

kanekotic:
@christopherdro as per one of the testsi perform while I was working on this you can see it in #1003 . In IOS setting this in `componentDidMount` works fine. So this is actually a specifical note for android

PR #1042, opened 2017-02-15
Add `npm start` to iOS steps

dfjs:


christopherdro:
Hey there! Thanks for PR but I don't think this step really applies here.

Clicking `Run` as mentioned in step 3 will automatically start the packager as will building the app using `react-native run-ios`

Issue #934, opened 2017-01-04
Ability to Customize "My Location" Text


joncursi:
Would be great if this text were customizable via prop, especially for internationalized apps that use different languages based on user locale.

![screen shot 2017-01-04 at 3 19 35 pm](https://cloud.githubusercontent.com/assets/4974609/21657654/418e4c06-d291-11e6-8871-c109da1c3573.png)

americool:
I did a little triaging and I think I've found the relevant component(s) inside MApView that generating the AIRMap on the fly. https://github.com/airbnb/react-native-maps/blob/master/components/MapView.js#L646-L648 

I'm new to react-native and this project, but I'm eager to help out. Would this require extensive objective C/Java knowledge?  

I'll continue to read the code for know, but any direction would be greatly appreciated.

PR #1049, opened 2017-02-20
Issue #934 Add customized user location annotation text

pyliaorachel:
##### Description

On iOS, the user location annotation has a default and unsettable text `My Location`, as shown in issue #934.

##### Fix

Now you can add `userLocationAnnotationTitle` as a prop to `<MapView />` when `showsUserLocation` is true.

##### Note

I had no prior experience in Objective-C so do let me know if the codes modified are dirty. Any other suggestions on the modified code are also welcome.

### DIFF
--- a/components/MapView.js
+++ b/components/MapView.js
@@ -87,6 +87,15 @@ const propTypes = {
   showsUserLocation: PropTypes.bool,
 
   /**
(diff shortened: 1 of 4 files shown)


lelandrichardson:
Thanks!

PR #1060, opened 2017-02-24
Add one more tip to Android installation troubleshooting

evansiroky:
Add note about dealing with correcting the system image to use with the
emulator.


docs/installation.md
```diff
@@ -212,5 +212,7 @@ Enter the name of the API key and create it.
    gradlew clean
    cd ..
   ```
-  
+
 1. If you are using Android Virtual Devices (AVD), ensure that `Use Host GPU` is checked in the settings for your virtual device.
+
+1. If using an emulator and instead of the map, you see the message `[APPNAME] won't run without Google Play services which are not supported by your device.`, you need to change the emulator CPU/ABI setting to a system image that includes Google APIs.  These may need to be downloaded from the Android SDK Manager first.
```
christopherdro: Do you mean instead of a device?

evansiroky: No, on an emulator.  After re-reading this it didn't make sense to me either, so I updated the wording.

Issue #553, opened 2016-09-08
Custom callouts and markers not working correctly


nazywamsiepawel:
Hey all,

After upgrading to RN 0.32.0 and react-native-maps 0.8.0 weird things started happening with maps in my app. At first, I have experienced the callout cut-off issue (#427) which is "workaroundable". What puzzles me now is the issues that came after :

<img width="320" alt="screen shot 2016-09-08 at 15 34 15" src="https://cloud.githubusercontent.com/assets/334875/18353455/eef5197a-75d9-11e6-9f92-d00a314b6f01.png">
1. In order to trigger the callout I need to click a little bit over the custom marker (which in this case is the red square). Clicking on the red square itself doesn't do anything in this regard apart from triggering onSelect and onPress (callout should show at this point as well!).
2. After clicking a little bit over the marker the callout seems to show but under other markers. How to position it back on the top? It worked correctly in the previous versions.

Supplementary snippet :

```
 <MapView.Marker
            coordinate={marker.loc}
            onSelect={() => alert("test")}
            onPress={() => alert("test")}
          >
            <View style={{height:20, width:20, backgroundColor:'red'}} />
            <MapView.Callout tooltip={false}>
              <BCClickableCallout title={marker.title} bar={marker.bar} onSelected={() => { alert("go places")  </MapView.Callout>
          </MapView.Marker>
```

What happened in the new version that broke it all? Is it me doing something wrong or is it a bug?

latimesmike:
I'm also having the same problem, here is the code I'm using to test out the events:

```
import React, { Component } from 'react';
import MapView from 'react-native-maps';
import {
  StyleSheet,
  View,
} from 'react-native';

var styles = StyleSheet.create({
    container: {
    	backgroundColor: '#003D40',
    	flex: 1,
    	marginTop: 65
    },
    map: {
        position: 'absolute',
        top: 0,
        left: 0,
        right: 0,
        bottom: 0,
      },
})
class Main extends React.Component{
	constructor(props) {
		super(props);
		this.state = {
			region: {
		      latitude: 37.78825,
		      longitude: -122.4324,
		      latitudeDelta: 0.0922,
		      longitudeDelta: 0.0421,
		    }
		}
	}
	onRegionChange(region) {
	  this.setState({
	  	region: region
	  });
	}
	render() {
		return (
			<View style={styles.container}>
				<MapView.Animated
				      region={this.state.region}
				      onRegionChange={this.onRegionChange.bind(this)}
[...]

Issue #436, opened 2016-07-28
Markers buggy on 0.30.0


yonahforst:
Custom markers became buggy on iOS and Android when upgrading from 0.28.0 to 0.30.0,
### iOS
- marker sometimes stay in the same screen position
- custom marker onSelect only called when marker pressed just outside the view. It looks like onPress is consuming the click and not letting the callout show.
  ![srrgzinqgv](https://cloud.githubusercontent.com/assets/1440796/17217605/3579c10e-54e5-11e6-8794-aa85204bc5fc.gif)

![ezgif com-gif-maker 1](https://cloud.githubusercontent.com/assets/1440796/17217937/5ebcd8d4-54e6-11e6-92ff-1cadb4a6f656.gif)
### Android
- custom marker drawn slightly above coordinate
- custom callout clipped halfway
  ![pnso6eusy2](https://cloud.githubusercontent.com/assets/1440796/17218816/7935d474-54e9-11e6-90d8-2940dcdcec9c.gif)

LeBlaaanc:
I'm having a similar issue on 0.29.2.

carleryd:
I am also experiencing this with react-native@0.30.0 and react-native-maps@0.7.1 using <MapView.Marker> with <Image> inside on both iPad, iPhone and simulator. Works great on Android though!

monotv:
Markers with a custom images are broken too. Also on react-native@0.30.0 and react-native-maps@0.7.1

yonahforst:
has anyone tested this on 0.31.0?

nazywamsiepawel:
@yonahforst I'm using it with 0.31 now - callouts are cut off, markers' onSelect is giving some random clicks.

pellepersson:
Any luck with this? I am experiencing a similar issue in iOS with react-native@0.31 and react-native-maps@0.7.1.

Everything works as expected using the standard marker, but when using anything else (just a view with a height, width and background color set as a simple example), the "onSelect" event is triggered (or "showCallout" method is called) randomly. The "onPress" event is triggered at every click on the marker though, but it is just not passing the event to whatever should happend next...

alexHlebnikov:
I'm having slightly different issue with markers on Android. 
My markers have a custom image inside and a label. But they renders everytime in different way. 
[...]

PR #1079, opened 2017-03-03
fix onSelected event for markers with custom view

yonahforst:
addresses #436, parts of #553 

Problem: when using markers with a custom subview, `didSelectAnnotation` does not get triggered on tap events, unless you press on the outside edge of the marker (outside the subview but inside the marker)

Solution: 'marker-press' events manually call `selectAnnotation` on the marker's mapview


lelandrichardson:
Thank you!

leonardogutierrezh:
this creates an issue when you have multiple markers close together, the callout box jumps between the different markers

switz:
Can this be hidden behind a flag? I believe it's causing multiple callouts to open on iOS when they're nearby.

PR #1092, opened 2017-03-06
Update for dependency issues

ddipanfilo:
https://github.com/airbnb/react-native-maps/issues/1067 this is a common issue people are experiencing right now


christopherdro:
@ddipanfilo Thanks for the PR.

Should we maybe update be either specific with the version of React Native that required 0.13.0, or make it generic and recommend users to be on the latest version.

Users of React Native >= .40 should use 0.13.0 whereas older versions use 0.12.4

ddipanfilo:
I updated the commit to reflect this distinction!  Please let me know if there is anything else to consider.

anonrig:
Can we get this merged? Without this PR I couldn't have installed react-native-maps properly on RN 0.42.x

PR #1112, opened 2017-03-12
Implements animateToCoordinate to Google Maps iOS

peterept:
Implements [animateToCoordinate](https://github.com/airbnb/react-native-maps/blob/master/docs/mapview.md#methods) on Google Maps iOS

To see it working, use the follow:

`index.ios.js:`
```jsx
import React, { Component } from 'react';
import {
  AppRegistry,
  StyleSheet,
  Button,
} from 'react-native';
import MapView from 'react-native-maps'

export default class HelloMaps extends Component {
  render() {
    return (
      <MapView ref={component => this.map = component} style={styles.container}
          provider={MapView.PROVIDER_GOOGLE}
      >
        <Button 
          title="Go Sydney" 
          onPress={() => { this.map.animateToCoordinate({latitude: -33.852896, longitude: 151.210291})  }}
        />
      </MapView>
    );
  }
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'flex-end',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
});

AppRegistry.registerComponent('HelloMaps', () => HelloMaps);
```

Issue #955, opened 2017-01-12
animateToCoordinate in iOS with Google Maps


KarlosQ:
If I execute the method animateToCoordinate in iOS using the google maps the app crash and show me this stack trace:

![captura de pantalla 2017-01-12 a las 13 28 07](https://cloud.githubusercontent.com/assets/15233913/21889820/1a50cd86-d8cb-11e6-9c78-367ab820933f.png)

Can anybody tell me what I should do? Thanks!

KarlosQ:
The problem is easy resolved if instead animateToCoordinate I use animateToRegion with current region.longitudeDelta and region.latitudeDelta.

jpgarcia:
I'm facing the same issue

boytainnt:
i think because you use Google Map and rn-maps stil get AirMap 
<img width="673" alt="screen shot 2017-02-17 at 3 08 12 pm" src="https://cloud.githubusercontent.com/assets/15927583/23057472/f19f7968-f522-11e6-86e9-220691e659cb.png">

Issue #811, opened 2016-11-19
animateToCoordinate not implemented for Google Maps


therealgilles:
The name of the native function on Google Maps seems to be animateToLocation. The simplest workaround is to use animateToRegion instead.

PR #1115, opened 2017-03-12
[iOS - Google Maps] Fix animateToCoordinate and animateToRegion

ryankask:
This addresses #955 and #811 by doing the following:

- Add the missing `animateToCoordinate()` for Google Maps on iOS - I just noticed that @peterept recently create #1112 😄 but one issue I found when I was working on it was that UIView animation doesn't work. This means the `duration` argument is ignored. The solution was to use Core Animation, which I found here: http://stackoverflow.com/questions/15662148/controlling-animation-duration-in-google-maps-for-ios/15663039#15663039
- Add a new button to the "Tracking Position" example that uses `animateToCoordinate()`. It's very similar to the existing `animateToRegion()` button but is a useful test.
- Use Core Animation for `animateToRegion()`. This is backwards incompatible because the default duration of 500ms was being ignored before but now works.

Google Maps requires that the QuartzCore framework is linked against the project so I assume it's fine to use `CATransaction`.

### DIFF
--- a/example/examples/DisplayLatLng.js
+++ b/example/examples/DisplayLatLng.js
@@ -43,12 +43,22 @@ class DisplayLatLng extends React.Component {
     this.map.animateToRegion(this.randomRegion());
   }
 
-  randomRegion() {
-    const { region } = this.state;
+  animateRandomCoordinate() {
+    this.map.animateToCoordinate(this.randomCoordinate());
+  }
+
+  randomCoordinate() {
+    const region = this.state.region;
     return {
-      ...this.state.region,
       latitude: region.latitude + ((Math.random() - 0.5) * (region.latitudeDelta / 2)),
       longitude: region.longitude + ((Math.random() - 0.5) * (region.longitudeDelta / 2)),
+    };
+  }
+
+  randomRegion() {
+    return {
+      ...this.state.region,
+      ...this.randomCoordinate(),
     };
   }
 
@@ -74,13 +84,19 @@ class DisplayLatLng extends React.Component {
             onPress={() => this.jumpRandom()}
             style={[styles.bubble, styles.button]}
           >
-            <Text>Jump</Text>
+            <Text style={styles.buttonText}>Jump</Text>
           </TouchableOpacity>
           <TouchableOpacity
             onPress={() => this.animateRandom()}
             style={[styles.bubble, styles.button]}
           >
-            <Text>Animate</Text>
+            <Text style={styles.buttonText}>Animate (Region)</Text>
+          </TouchableOpacity>
+          <TouchableOpacity
+            onPress={() => this.animateRandomCoordinate()}
+            style={[styles.bubble, styles.button]}
+          >
+            <Text style={styles.buttonText}>Animate (Coordinate)</Text>
           </TouchableOpacity>
         </View>
       </View>
@@ -112,15 +128,19 @@ const styles = StyleSheet.create({
     alignItems: 'stretch',
   },
   button: {
-    width: 80,
-    paddingHorizontal: 12,
+    width: 100,
+    paddingHorizontal: 8,
     alignItems: 'center',
-    marginHorizontal: 10,
+    justifyContent: 'center',
+    marginHorizontal: 5,
   },
   buttonContainer: {
     flexDirection: 'row',
     marginVertical: 20,
     backgroundColor: 'transparent',
+  },
+  buttonText: {
+    textAlign: 'center',
   },
 });
 
--- a/lib/ios/AirGoogleMaps/AIRGoogleMapManager.m
+++ b/lib/ios/AirGoogleMaps/AIRGoogleMapManager.m
@@ -76,10 +76,14 @@ RCT_EXPORT_METHOD(animateToRegion:(nonnull NSNumber *)reactTag
     if (![view isKindOfClass:[AIRGoogleMap class]]) {
       RCTLogError(@"Invalid view returned from registry, expecting AIRGoogleMap, got: %@", view);
     } else {
-      [AIRGoogleMap animateWithDuration:duration/1000 animations:^{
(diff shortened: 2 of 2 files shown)


peterept:
Hi @ryankask brilliant mate - great catch on duration not working - Embarrassingly I assumed the animation code I copied from animateToRegion worked..  

So I pulled this PR down and ran it with the following test page with a duration as 1500ms:

```jsx
import React, { Component } from 'react';
import {
  AppRegistry,
  StyleSheet,
  Button,
  View,
  Text
} from 'react-native';
import MapView from 'react-native-maps'
import CustomCallout from './CustomCallout';

export default class HelloMaps extends Component {
  state = {
    markers: [
      { key: "test", coordinate: {latitude: -33.852896, longitude: 151.210291}}
    ]
  }
  render() {
    return (
      <View style={styles.container} >
      <MapView ref={component => this.map = component} style={styles.map} 
          provider={MapView.PROVIDER_GOOGLE}
      >
          {this.state.markers.map(marker => (
            <MapView.Marker
              title={marker.key}
              image={require('./assets/pin-comment.png')}
              key={marker.key}
              coordinate={marker.coordinate}
            >
              <MapView.Callout tooltip style={styles.customView}>
                <CustomCallout>
                  <Text>This is a custom callout bubble view</Text>
                </CustomCallout>
              </MapView.Callout>
            </MapView.Marker>            
          ))}
      </MapView>
      <Button 
        title="Go Sydney" 
        onPress={() => { this.map.animateToCoordinate({latitude: -33.852896, longitude: 151.210291}, 1500)  }}
      />
      </View>
    );
  }
}

  container: {
    flex: 1,
    justifyContent: 'flex-end',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
  map: {
    ...StyleSheet.absoluteFillObject,
  },  
  customView: {
    width: 140,
    height: 100,
  },  
});

AppRegistry.registerComponent('HelloMaps', () => HelloMaps);

```

Just to confirm with my fix it animates in about 1 sec (as you discovered):

[Imgur](http://i.imgur.com/kh1yKrN.gifv)

Now if I run with your fix:

[Imgur](http://i.imgur.com/CMWEJhu.gifv)

Expected: Animates for 15 seconds
Actual: Animated for ~4 seconds. The animation is 4 times faster then expected.

_Note: Same behaviour in both IOS Simulator (iPhone 6) and real device (iPhone 5)._
Xcode version 8.1

On playing around if I use the duration 1500*4, then it runs at 15 seconds.

I have searched around and I can't see why this is happening! :-( 

Small nit, a bit cleaner to read rather then:
` [CATransaction setValue:[NSNumber numberWithFloat: duration/1000] forKey:kCATransactionAnimationDuration];` 
use: `[CATransaction setAnimationDuration:duration/1000];`

ryankask:
Hi @peterept. Thanks for pointing out the `setAnimationDuration` method. It's much nicer and I've updated the PR.

In your example, can you change the duration to `15000`  (note the extra 0) to check if it animates for 15 seconds? We're going from millisecond durations in the JS world to durations in seconds. The video seems like 1.5 seconds to me although it is bit tricky with tiles loading after.

peterept:
Ahh yes that was it @ryankask ! !

I was staring at the 1500 for an hour and didn't notice it should have been 15000. Thank you for spotting that.

Confirmed, 15000 animates correctly for 15 seconds:

[animateToCoordinate imgur](http://i.imgur.com/LGm6rVf.gifv)

Also checked animateToRegion with `this.map.animateToRegion({latitude: -33.852896, longitude: 151.210291, latitudeDelta: 0.0025, longitudeDelta: 0.0025}, 15000)`, and it correctly works:

[animateToRegion imgur](http://i.imgur.com/ivUzaZA.gifv)

Both work and smooth as silk on simulator and device. 

Your PR lgtm. 

Nice one, this will improve my app!

grundmanise:
Can  `fitToElements`, `fitToSuppliedMarkers`, `fitToCoordinates`,  also have a duration prop?

ryankask:
@grundmanise They might be able to but at the moment the only two methods on `MapView` that take a `duration` argument or the ones addressed in this PR.

@peterept I'm not sure. Perhaps next time the project maintainers comb through the PRs, they can comment.

At the moment I just created a new category in `AIRGoogleMapManager+Animation.m` to add the `animateToCoordinate` method.

grundmanise:
Thanks for the answer.

> Use Core Animation for animateToRegion(). This is backwards incompatible because the default duration of 500ms was being ignored before but now works.

You mean that now `animateToRegion` ignores duration prop when `PROVIDER_GOOGLE`?

ryankask:
If I've observed correctly, it doesn't ignore it but it tries to use it with an iOS animation method that doesn't work. To the end user the effect is that it's ignored.

seppemarotta:
I'm so exited to add this !

lelandrichardson:
@ryankask i'd like to get this in... do you think you can rebase? I've changed some of the project structure around and it's causing conflicts

ryankask:
@lelandrichardson I've rebased the branch. There weren't any conflicts but I wasn't able to run the example project.

There are some instructions for how to run the example project here: https://github.com/airbnb/react-native-maps/blob/master/docs/examples-setup.md. It looks like the `npm install` part needs to be tweaked because of the reorganisation but what about the Cocoapods part?

I'm seeing the below but I haven't had a chance to look into it:

2017-03-27 at 09 31

guilhermepontes:
any updates on this?

mbroadst:
@lelandrichardson just ran into this problem myself, is there anything that can be done to expedite merging this?

sonaye:
Been waiting for this one to get merged for a while.

terribleben:
Just chiming in that we've verified this fix on Expo and are pulling it into our next SDK version.

peterept:
Thanks @terribleben! (and of course @ryankask for the solution) 😀

ngandhy:
@ryankask - Thanks!! Works great :)

Issue #414, opened 2016-07-22
Crashes when clicking back button to exit on Android


DVassilev:
Hi, I use MapView as the main screen of the app and when I try exiting using the back button it crashes with the following error
```java.lang.NullPointerException: Attempt to invoke interface method 'void maps.af.y.q()' on a null object reference
	at maps.af.Q.d(Unknown Source)```
Anyone know how I can fix it?
Thanks

**Comment by DavidYKay** [comment]
My users are getting this error left and right as well. I haven't gotten to the bottom of it yet, but would love to figure this out.

**Comment by DVassilev** [comment]
@DavidYKay I can't seem to get to the bottom of it either. Putting logs on this here too.

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'void maps.ad.z.q()' on a null object reference
       at maps.ad.R.d(Unknown Source)
       at wc.onTransact(:com.google.android.gms.DynamiteModulesB:87)
       at android.os.Binder.transact(Binder.java:387)
       at com.google.android.gms.maps.internal.IMapViewDelegate$zza$zza.onDestroy(Unknown Source)
       at com.google.android.gms.maps.MapView$zza.onDestroy(Unknown Source)
       at com.google.android.gms.dynamic.zza.onDestroy(Unknown Source)
       at com.google.android.gms.maps.MapView.onDestroy(Unknown Source)
       at com.airbnb.android.react.maps.AirMapView.doDestroy(AirMapView.java:289)
       at com.airbnb.android.react.maps.AirMapManager.onDropViewInstance(AirMapManager.java:70)
       at com.airbnb.android.react.maps.AirMapManager.onDropViewInstance(AirMapManager.java:27)
       at com.facebook.react.uimanager.NativeViewHierarchyManager.dropView(NativeViewHierarchyManager.java:514)
       at com.facebook.react.uimanager.NativeViewHierarchyManager.dropView(NativeViewHierarchyManager.java:523)
       at com.facebook.react.uimanager.NativeViewHierarchyManager.dropView(NativeViewHierarchyManager.java:523)
       at com.facebook.react.uimanager.NativeViewHierarchyManager.dropView(NativeViewHierarchyManager.java:523)
       at com.facebook.react.uimanager.NativeViewHierarchyManager.dropView(NativeViewHierarchyManager.java:523)
       at com.facebook.react.uimanager.NativeViewHierarchyManager.manageChildren(NativeViewHierarchyManager.java:414)
[...]

PR #1130, opened 2017-03-15
issue#939, fix multiple-instance memory leak

mattshen:
This is to revert part of PR#694 to fix the memory leak problem on android. 

It also has the patch from https://github.com/IjzerenHein/react-native-maps/commit/14ca6f73078a2e8dedccfeefe8a037d55f03c59c to avoid the crash on pause -- which is mentioned in #414

### DIFF
--- a/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
+++ b/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
@@ -21,6 +21,7 @@ import android.widget.ProgressBar;
 import android.widget.RelativeLayout;
 
 import com.facebook.react.bridge.LifecycleEventListener;
+import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
@@ -79,7 +80,9 @@ public class AirMapView extends MapView implements GoogleMap.InfoWindowAdapter,
     private final ScaleGestureDetector scaleDetector;
     private final GestureDetectorCompat gestureDetector;
     private final AirMapManager manager;
+    private LifecycleEventListener lifecycleListener;
     private boolean paused = false;
+    private boolean destroyed = false;
     private final ThemedReactContext context;
     private final EventDispatcher eventDispatcher;
 
@@ -254,7 +257,7 @@ public class AirMapView extends MapView implements GoogleMap.InfoWindowAdapter,
         // updating location constantly, killing the battery, even though some other location-mgmt
         // module may
         // desire to shut-down location-services.
-      LifecycleEventListener lifecycleListener = new LifecycleEventListener() {
+    lifecycleListener = new LifecycleEventListener() {
         @Override
         public void onHostResume() {
           if (hasPermissions()) {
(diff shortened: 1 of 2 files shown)


lelandrichardson:
@gpeal when you get a chance, can you review this?

android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
```diff
@@ -273,11 +276,15 @@ public void onHostPause() {
             //noinspection MissingPermission
             map.setMyLocationEnabled(false);
           }
-          paused = true;
+            synchronized (AirMapView.this) {
+                AirMapView.this.onPause();
+                paused = true;
```
gpeal: Can you move setting paused to true and false into onPause and onResume? The boolean logic being here breaks now that you call onPause separately.

android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
```diff
@@ -289,6 +296,24 @@ private boolean hasPermissions() {
...
 
+    /*
+    onDestroy is final method so I can't override it.
+     */
+    public synchronized  void doDestroy() {
+        if (lifecycleListener != null && context != null) {
+            context.removeLifecycleEventListener(lifecycleListener);
+            lifecycleListener = null;
+        }
+        if(!paused) {
```
gpeal: How would destroy get called without pause?

android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
```diff
@@ -289,6 +296,24 @@ private boolean hasPermissions() {
...
+            onPause();
+        }
+        if (!destroyed) {
+            onDestroy();
+            destroyed = true;
```
gpeal: You can just set destroyed to true at the top of the method. This will prevent future logic errors if you ever return early or anything

gpeal: You can also start the method with:
```
if (destroyed) {
    return;
}
destroyed = true;
```

android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
```diff
@@ -289,6 +296,24 @@ private boolean hasPermissions() {
                 checkSelfPermission(getContext(), PERMISSIONS[1]) == PackageManager.PERMISSION_GRANTED;
     }
 
+    /*
+    onDestroy is final method so I can't override it.
+     */
+    public synchronized  void doDestroy() {
+        if (lifecycleListener != null && context != null) {
```
gpeal: What happens if this gets called while the map is initializing before onMapReady is called? Can you cancel the initialization? Maybe return early from onMapReady if destroyed?

lelandrichardson:
@gpeal @mattshen I'm going to merge this and add @gpeal's requested changes in a followup PR

mattshen:
Hi @lelandrichardson, thanks. @gpeal 's comments make sense to me (but I haven't got a chance to change  :( ). If your followup PR contains some changes requiring some real case testing, please let me know, I will apply it to our app and let the testers to have a try.

lelandrichardson:
@mattshen check out #1130 ^

PR #1134, opened 2017-03-16
Add iOS tip

evansiroky:
Add clarification to iOS options section that the advanced options may not be necessary.


spikebrehm:
Thanks 🍻

PR #1138, opened 2017-03-17
Use local RCTConvert+MapKit instead of the one in React Native

janicduplessis:
Since https://github.com/facebook/react-native/commit/48f30eca7e3d4c239501de515a7cc35615ed6bd1 RCTConvert+MapKit was removed from React Native so this pulls in the relevant methods and renames `RCTConvert+MoreMapKit` to `RCTConvert+MapKit`. I checked that having these extension defined multiple times doesn't cause issues so this is fully backwards compatible.


lelandrichardson:
Thanks!

PR #1144, opened 2017-03-20
Fix UIPanGestureRecognizer not being triggered

justinhouse:
Suggested fix for https://github.com/airbnb/react-native-maps/issues/1142

### DIFF
--- a/lib/ios/AirMaps/AIRMapManager.h
+++ b/lib/ios/AirMaps/AIRMapManager.h
@@ -9,6 +9,6 @@
 
 #import <React/RCTViewManager.h>
 
-@interface AIRMapManager : RCTViewManager, UIGestureRecognizerDelegate
+@interface AIRMapManager : RCTViewManager
 
 @end
--- a/lib/ios/AirMaps/AIRMapManager.h
+++ b/lib/ios/AirMaps/AIRMapManager.h
@@ -9,6 +9,6 @@
 
 #import <React/RCTViewManager.h>
 
-@interface AIRMapManager : RCTViewManager
+@interface AIRMapManager : RCTViewManager, UIGestureRecognizerDelegate
 
 @end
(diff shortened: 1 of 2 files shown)


lelandrichardson:
@justinhouse can you rebase this? I'd like to get this in. I've restructured the project a bit and it's causing some conflicts.

I'm also a little bit worried about this change breaking other things. Will need to pull it down and test some things out.

Thanks for helping with this!

justinhouse:
@lelandrichardson It's rebased. I've worked with it for a bit already and so far no problems, but granted I don't have any functionality that drags markers. Testing that seems worth it.

I quickly did a test in Swift (works faster for me) and it did trigger the pan, while also dragging Markers, which we wouldn't want since the map stays in place during that interaction. I tried checking whether the otherRecognizer was a UIPanGestureRecognizer as well and in that case not recognize our map-dragger. For some reason I don't see right now it works the other way around. Whereas the suggested fix returns true to get the pan to always trigger, returning true only when the other recognizer is a pan as well makes it work as supposed to (when also dragging markers) where I would expect it to work the other way around. Anyway, that's my thoughts and findings on the matter. I didn't update the PR yet as I'm not fully confident about the addition and would appreciate your thoughts on it.

```
func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
        if(otherGestureRecognizer.isKind(of: UIPanGestureRecognizer.self)) {
            return true;
        }
        
        return false;
    }
```

lelandrichardson:
@justinhouse thanks for the response. that's helpful! I'm going to pull this down and play around a little bit before merging, but if all looks well... 👍

ccreonopoulos:
@lelandrichardson  Hello, this has showed up on a couple of projects as an issue and I was wondering what the state of the PR is?

rborn:

PR #1151, opened 2017-03-21
Use provided as the gradle dependency

lelandrichardson:
to: @gpeal 
cc: @GantMan 

This allows the library to compile but will use whichever RN the
app is using and also allows it to pass POM verification on sonatype
which doesn't allow non-explicit dependencies.

I believe this can actually finally settle the gradle `:+` vs `:0.x` debate! This seems to work how we want it.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -53,7 +53,7 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


spikebrehm:
🍻

PR #1152, opened 2017-03-21
Update the android buildToolsVersion to 25.0.0

markusguenther:
When using the current android studio with the current SDKs you have the problem that your app is on version 25 and you can not create a APK without adjusting the buildToolsVersion in the build.gradle

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -40,7 +40,7 @@ allprojects {
 
 android {
   compileSdkVersion 25
-  buildToolsVersion "23.0.3"
+  buildToolsVersion "25.0.0"
 
   defaultConfig {
     minSdkVersion 16
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -69,8 +69,8 @@ def enableSeparateBuildPerCPUArchitecture = false
 def enableProguardInReleaseBuilds = false
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion "23.0.3"
+    compileSdkVersion 25
+    buildToolsVersion "25.0.0"
 
     defaultConfig {
         applicationId "com.airbnb.android.react.maps.example"


tlvenn:
Are we sure this is not creating issue given RN and pretty much all the libs I know depends on SDK 23 ?

tlvenn:
The `targetSdkVersion` is still targeting 23 but I am just wondering if it will not force people to move to use android SDK 25 and update their own gradle file to use `buildToolsVersion 25` as well.

markusguenther:
But isn`t it good to use the newer SDK. The Android developer documentation page recommend to use the newer versions. The app itself can still support older versions. You configure always minimum and target version.

```
defaultConfig {
  minSdkVersion 16
  targetSdkVersion 25
}
```

Thought that is better. But if leads to problems we can revert it and I just patch this in my project.

tlvenn:
I totally agree it's better to use newest SDK when you can. I am just saying that we have been bitten by gradle before in some unexpected ways with regards to sdkVersion and buildTools where for example, our CI builds would not work anymore because the needed build tools required by a dep was not availaible out of the box.

For example, if this indeed force you to update your own gradle file, then people will not be able to use create-react-native-app with react native maps without ejecting.

If this change has absolutely no side effect on your own app and simply requires you to have the latest buildTools installed, this is probably safe then.

ryankask:
The default React Native project templates specify `buildToolsVersion` 23.0.1: https://github.com/facebook/react-native/blob/master/local-cli/templates/HelloWorld/android/app/build.gradle#L87

If you're using a new build tools version, you can override the `buildToolsVersion` for each subproject using something like this: https://gist.github.com/AndrewJack/10dc22708582b009fb74f07904f3bfb1

tlvenn:
That's a neat trick @ryankask, I did not know you could do that. Just pointing out that this is not an option if you are using the new create-react-native-app and expo.

lelandrichardson:
@tlvenn my understanding was that create-react-native-app and expo both don't let you install any extra native modules other than what's supported in their SDK. Is that not correct (totally possible i'm not understanding CRNA or expo properly)?

react-native-maps is one of the supported native modules for CRNA and expo and I don't think this should change that in any way?

tlvenn:
Hi @lelandrichardson, your understanding is correct, Expo is bundling react-native-maps and therefore they handle the dependencies on their side. CRNA does not let you install any native deps at the moment, should they decide to include react-native-maps, it will be the same story as Expo.

So looks like I was worried for nothing but learnt an interesting Gradle tip in the process ;)

PR #1166, opened 2017-03-27
Address PR feedback from #1130

lelandrichardson:
to: @mattshen @gpeal 

I addressed some of @gpeal's feedback from #1130 

So it looks like `onDestroy`, `onPause` and `onResume` are final methods, and so are not able to be overridden... which makes some of gabe's feedback impossible to implement.  As a result, this PR doesn't include that many changes... but LMK what ya'll think.

### DIFF
--- a/lib/android/src/main/AndroidManifest.xml
+++ b/lib/android/src/main/AndroidManifest.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.airbnb.android.react.maps" >
+    <uses-sdk android:minSdkVersion="16" />
     <uses-permission android:name="android.permission.INTERNET" />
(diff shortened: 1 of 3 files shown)


lib/android/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.airbnb.android.react.maps" >
+    <uses-sdk android:minSdkVersion="16" />
```
mattshen: Not sure if this is necessary, because build.gradle defines the minSdkVersion and would override the one defined in AndroidManifest.xml. Please correct me if I'm wrong.

lelandrichardson: It's necessary for intelliJ / Android Studio to give you the proper warnings...

...it's not related to this PR, however :P

mattshen: TIL :)

Issue #1104, opened 2017-03-10
TestID's for e2e automation


nischayv:
I am trying to add testID's to the map components, similar to Views in react native. The goal is to make react-native-maps testable using automation tools like Appium. This would allow end to end testing in react-native-maps. I started off with the Google Maps MapMarker. But I am running into challenges and Appium is still not able to find the marker. Here is my change - https://github.com/nischayv/react-native-maps/commit/c37548d18b25fbe1521e8263ff63577931b00215

Any help would be appreciated! 

cc: @unboundfire  @lelandrichardson

kidmillions:
+1 to this. Blocking automation efforts for us...we'd love to be able to automate map interactions with Appium and XCUITest.

lelandrichardson:
@nischayv this looks great. I would love it if you opened this up as a PR and we can discuss it there?

tiagofs:
Hi,
Have you managed to find a map marker with Appium?

PR #1170, opened 2017-03-27
TestID's for e2e automation

nischayv:
#1104

### DIFF
--- a/ios/AirGoogleMaps/AIRGoogleMapMarkerManager.m
+++ b/ios/AirGoogleMaps/AIRGoogleMapMarkerManager.m
@@ -23,6 +23,8 @@ RCT_EXPORT_MODULE()
 //  tapGestureRecognizer.cancelsTouchesInView = NO;
 //  [marker addGestureRecognizer:tapGestureRecognizer];
   marker.bridge = self.bridge;
+  marker.isAccessibilityElement = "YES";
+  marker.accessibilityElementsHidden = "NO";
   return marker;
 }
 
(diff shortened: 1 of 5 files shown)


ios/AirGoogleMaps/AIRGoogleMapMarkerManager.m
```diff
@@ -23,6 +23,8 @@ - (UIView *)view
 //  tapGestureRecognizer.cancelsTouchesInView = NO;
 //  [marker addGestureRecognizer:tapGestureRecognizer];
   marker.bridge = self.bridge;
+  marker.isAccessibilityElement = "YES";
+  marker.accessibilityElementsHidden = "NO";
```
lelandrichardson: 1. Are these supposed to be strings?
2. It seems like we shouldn't set these here, but rather set them when/if people set the `testID` prop. (or other accessibility props if we add them, which would be great if we did)

lelandrichardson:
@nischayv thanks for putting up this PR. Would you mind rebasing? Seems like there are several conflicts.

unboundfire:
@lelandrichardson   

@nischayv and I will be updating this PR soon that will add support for testIDs and accessibilityLabels.  We have gotten accessibilityLabels and testIDs to work on iOS(mapkit)/Android. We haven't figured out how to get testIDs to work with iOS google maps since there overlay objects do not seem to support the `accessibilityIdentifier` property and it seems like google maps is generating there own ids when we set `map.accessibilityElementsHidden = NO`.

kirakik:
@unboundfire any news on this?

christopherdro:
Superseded by https://github.com/react-native-community/react-native-maps/pull/2253/

PR #1178, opened 2017-03-30
Update docs to specify how to access event data

### LINKED ISSUE #1136: Confusion on Readme documentation
The document is a bit misleading, especially for the part where it mentioned onPress/onLongPress events (for MapView or MapView.Marker for example) where the stated return would be "{ coordinate: LatLng, position: Point }".

Source: https://github.com/airbnb/react-native-maps/blob/master/docs/mapview.md

After scrutinizing the examples, I only realized that it should be something like "e.nativeEvent.coordinate" instead, where "e" is the direct return for onPress/onLongPress. For example:

//map view:
<MapView
     //truncated for clarity
                    showsUserLocation={true}
                    showsMyLocationButton={true}
                    onLongPress={this.onMapPress.bind(this)}
                ></MapView>

//event handler:
onMapPress(e) {
        alert(e.nativeEvent.coordinate);//returns object
        alert(e.nativeEvent.coordinate.latitude);//returns a number
//confusion happens when readers will assume "e" as parameter that has "coordinate" and "position" objects, which in turn returns null for both
    }

Appreciate some updates on the readme to cause future confusion for beginners like me.

Thanks and regards.

samcorcos:
resolves #1136


christopherdro:
Thanks!

PR #1184, opened 2017-04-03
Feature/always usage permissions

simonmitchell:
Adds support for requesting always permissions on iOS when NSLocationAlwaysUsageDescription is provided in the app's info.plist. This is required as apps cannot change their requested permissions type after launch and so if this library is used before other similar libraries which need always usage then permissions are broken by the time they request always usage permissions.


rborn:
@alvelig 🐽 LGTM

alvelig:
👍 LGTM

alvelig:
Oh, wait. But why do we need it in maps?

simonmitchell:
Because when I PR’d this, if you showed a map before requesting ‘always location permissions’ (in iOS 10) you can’t then request the user for always usage permissions... you’re stuck with ‘when in use’. As of iOS 11 I believe this has changed so you can ask the user to upgrade to always usage at a later point in the app flow, but this still breaks it in iOS 10! 

Hope that explains 👍🏻

alvelig:
Ah ok... Need to document that. Kinda:

### NSLocationAlwaysUsageDescription

If you need to use `requestAlwaysAuthorization`, then define `NSLocationAlwaysUsageDescription` in `Info.plist` and when maps are shown it will ask you `requestAlwaysAuthorization`

rborn:
@simonmitchell can you add the docs please? then we merge 🙌

simonmitchell:
@rborn Will endeavour too when I get a free second! 👍

simonmitchell:
Thinking about this now, it might be better to only call `requestAlwaysAuthorization` on iOS 10 and lower, because as of iOS 11 Apple recommend only calling `requestWhenInUseAuthorization` until you actually need always authorisation and provide a mechanism for "Upgrading" the permissions now. This will both fix the issue where you couldn't ever get always use permissions on iOS 10, and allow you to choose which permission to request on iOS 11. Thoughts @rborn @alvelig?

simonmitchell:
See article here: https://m.rover.io/wwdc-2017-update-significant-updates-to-location-permissions-coming-with-ios-11-41f96001f87f

rborn:
@simonmitchell I somehow missed your message 😔

I totally agree we should ask for the user's permission only when we really need it (good UX, not only what Apple says) so I'm curious how you want to tackle this?

Thanks for all the effort 🤗

rborn:

simonmitchell:
Apologies @rborn seems I missed yours too! Thinking I need to re-enable GitHub emails as I never think to read notifications on the website :/

PR #1194, opened 2017-04-05
[iOS - Google Maps] Add onPress support for Polyline

guilhermepontes:
This PR fixes the problem of not having the `onPress` event on Google Maps in iOS.


IgorVanian:
Any news about the merge?

christopherdro:
@guilhermepontes This looks good!

Do you think you can resolve the conflicts so we can get this merged in.

Conflict is from the library trying to use RCTConvert+MapKit instead of RCTConvert+AirMap

guilhermepontes:
@christopherdro done 👍

PR #1199, opened 2017-04-05
Update installation.md

moronicgeek:
Changed to #import <GoogleMaps/GoogleMaps.h> instead of the @import GoogleMaps; to reflect the example application


christopherdro:
Thanks for the PR but I will be closing this out since another PR that was recently merged addressed this.

Issue #1219, opened 2017-04-08
Do not install v0.14.0 on Android


hemedani:
In install instruction you say should add this line in `android/settings.gradle` :
```
include ':react-native-maps'
project(':react-native-maps').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-maps/android')
```

But in the `../node_modules/react-native-maps` we don't have `android` folder I think it's move to `lib` folder, so I change the path to :
```
include ':react-native-maps'
project(':react-native-maps').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-maps/lib/android')
```

The path error for dependency is fixed but we have `Unknown plugin "module-resolver"` problem :
```
TransformError: /home/syd/Work/ignite/TaxiBS/node_modules/react-native-maps/index.js: Unknown plugin "module-resolver" specified in "/home/syd/Work/ignite/TaxiBS/node_modules/react-native-maps/.babelrc" at 0, attempted to resolve relative to "/home/syd/Work/ignite/TaxiBS/node_modules/react-native-maps"
```

hemedani:
The version 0.13.0 working fine

henrikra:
I have exactly same problem after updating from 0.13.1 to 0.14 :/

chris-at-translate:
Yup I encountered the same problem too, upgrading the same as @henrikra above. I was also looking around and found that it's related to babelrc. For one reason or another. Let me find the issue again.

chris-at-translate:
Sorry if this is messy I'm on my phone. This issue here #1206 references this issue here https://github.com/airbnb/react-native-maps/issues/795#issuecomment-289828439 which makes my project work again but breaks other libraries I'm using.

zengyanzi:
is there any solution towards this problem please?

iztokf:
i can confirm that version 0.14 is buggy on android:

1. `react-native link react-native-maps` not working
2. if i manually set up dependencies (settings.gradle, build.gradle, MainApplication.java) i get this during build:
```
A problem occurred configuring project ':app'.
> Could not resolve all dependencies for configuration ':app:_debugApk'.
   > Configuration with name 'default' not found.
```
3. after downgrading to version 0.13, works as expected

zengyanzi:
![image](https://cloud.githubusercontent.com/assets/21092607/24954792/ebdb5e20-1fd4-11e7-9396-33440b3b9f0d.png)
here is my error

karpoff:
[...]

PR #1226, opened 2017-04-10
Update installation.md

aaronofleonard:
Update the paths to the `android` and `ios` folders to match the new structure

Addresses first half of #1219


javiercr:
This worked for me too. Thanks!

PR #1230, opened 2017-04-11
Fixed path in `android/settings.gradle`

ndbroadbent:
I think the android folder moved:

`../node_modules/react-native-maps/android` => `../node_modules/react-native-maps/lib/android`


Nkzn:
same as #1226 ?

Issue #1206, opened 2017-04-07
.babelrc in git repo


davidhellsing:
https://github.com/airbnb/react-native-maps/blob/master/.babelrc

This file should be removed, or you will get `Unknown plugin: "module-resolver"` when compiling.

ZeroCho:
+1

tugorez:
+1 Same issue here

rigdern:
There's some discussion about this starting here: https://github.com/airbnb/react-native-maps/issues/795#issuecomment-289828439

khalid32:
I had to remove .babelrc file. Now works fine.

anydasa:
+1 Same issue here

anandbhardwaj:
+1 same issue. I am not able to use the map. getting error - Unknown plugin "Module-Resolver" specified in the .../.babelrc.

ZeroCho:
Temporay solution for this is to npm-install **module-resolver**

tugorez:
@ZeroCho or install the latest-working version.

sangagung:
@tugorez it seems that the developers pulled the earlier version from npm registry. 
Running `npm install react-native--maps@0.13.1` presented me with error 404 messages.

I've tried removing .babelrc and followed the suggestion from #1219, I still got the `Unknown plugin: "module-resolver"` error

tugorez:
@sangagung try with this version 0.12.4

anydasa:
I have installed 0.13.0 version.

quickskape:
@sangagung 

> Running npm install react-native-maps@0.13.1 presented me with error 404 messages.

You have two dashes in your package name it should be `npm install react-native-maps@0.13.1`

That should at least get rid of those 404 messages on install and targeting that version has also fixed my compile issue too.

EmielM:
Afaict this really does nothing, just adds another bunch of useless (dev)dependencies to my project. Aliassing the module is not necessary as it already exists in `node_modules/` 99% of the time. An idea to just remove the file for the next release?

agentilela:
Wrote a simple shell script to fix react-native-maps build errors that gets run whenever i rebuild my project. Maybe it'll be helpful to someone:

`sed -ie 's/RCTConvert\+MapKit\.m/RCTConvert\+AirMap\.m/g' node_modules/react-native-maps/lib/ios/AirMaps.xcodeproj/project.pbxproj
[...]

PR #1246, opened 2017-04-16
Add babelrc to npmignore

yangli1990:
I get errors saying that the plugin "module-resolver" does not exist because the .babelrc is published with the react-native-maps package. This PR will ensure this is not published in the future and as to avoid this error.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,1 +1,2 @@
 example
+.babelrc


javiercr:
This should fix #1206

f0rr0:
Merge this?

PR #1249, opened 2017-04-17
Update path in android installation

toddw:
I was having trouble compiling the android version but updating the project directory resolved it for me.


Nkzn:
same as #1226 ?

PR #1277, opened 2017-04-27
Fix for RN >= 0.40

scerelli:
This fix errors in `react-native`>= 0.40.

### DIFF
--- a/lib/ios/AirGoogleMaps/AIRGMSPolygon.h
+++ b/lib/ios/AirGoogleMaps/AIRGMSPolygon.h
@@ -6,7 +6,7 @@
 //
 
 #import <GoogleMaps/GoogleMaps.h>
-#import "UIView+React.h"
+#import "React/UIView+React.h"
 
 @class AIRGoogleMapPolygon;
 


lib/ios/AirGoogleMaps/AIRGMSPolygon.h
```diff
@@ -6,7 +6,7 @@
 //
 
 #import <GoogleMaps/GoogleMaps.h>
-#import "UIView+React.h"
+#import "React/UIView+React.h"
```
christopherdro: Should this be `<React/UIView+React.h>`?

scerelli:
Ye, let me fix that, I've some other changes to merge into this repo so I think you can close this PR and I'll open a new one

christopherdro:
Sounds good. Thanks for the update!

atom-r:
I just had a failure to compile on iOS because Polyline does the wrong import, just like Polygon as shown above in this PR.

I can provide more information about my setup if needed.

aaronbuchanan:
same as @atom-r 

As a workaround, I've added this to my package.json scripts for now:
`"postinstall": "sed -i '' 's/#import \"UIView+React.h\"/#import <React\\/UIView+React.h>/' ./node_modules/react-native-maps/lib/ios/AirGoogleMaps/AIRGMSPolyline.h"`

I feel dirty, lmk when a proper fix is up?

aaronbuchanan:
closing my own loop here, the fix is in [e085e2](https://github.com/airbnb/react-native-maps/commit/d62f5ef5ae4edb065643dc1e675753b134e085e2), but hasn't yet made it to npm.

Issue #1241, opened 2017-04-13
"'default' not found" issue after update to RN 0.43.3


MichelDiz:
It says:
Could not resolve all dependencies for configuration ':app:_debugApk'.
   > Configuration with name 'default' not found.

when I comment the React-native-maps lines it compiles normally.

It has too:
warning "react-native-maps@0.14.0" has incorrect peer dependency "react@>=15.4.0".

follow:
    "react-native-maps": "^0.14.0",
  //  compile project(':react-native-maps')

//include ':react-native-maps'
//project(':react-native-maps').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-maps/android/')

//import com.airbnb.android.react.maps.MapsPackage;

//       new MapsPackage()

Thanks - Best Regards xD

iiitmahesh:
it is bug in react-native-maps @14 

@MichelDiz  android folder not available in ../node_modules/react-native-maps/android. by copying android you will get it.

--> create android folder move . move all below folders to android.
 
![screenshot from 2017-04-14 15-22-56](https://cloud.githubusercontent.com/assets/13597781/25040338/5254153c-2126-11e7-9061-6ee4c486f965.png)

MichelDiz:
I downgrade to RN 0.43.0 and Works oO

Seprum:
Have wasted mine evening trying to make latest version build and I managed to do it.
Basically you need to manually link react-native-maps to your app using instruction, but in `settings.gradle`
you should add
`project(':react-native-maps').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-maps/lib/android')`
instead of
`project(':react-native-maps').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-maps/android')`
(Notice **lib** directory before **android**)
Note: I guess, auto linking with `react-native link react-native-maps` doesn't work because of no `android` directory located in `node_modules/react-native-maps`. Hope it will be fixed soon.

MichelDiz:
Many thanks! I gonna try out soon!

robgev:
@Seprum I got the same error after doing what you said. Linking is not working as you mentioned.

andrewkvavlewhite:
@robgev What he's saying is that because of the path not conforming to react-native standards, the command "react-native link" will not link anything for android. Have you gone through the manual installation process provided [here](https://github.com/airbnb/react-native-maps/blob/master/docs/installation.md#android)?

robgev:
[...]

PR #1257, opened 2017-04-20
Fix android path in settings.gradle configuration

rodolfobarretoweb:

PR #1290, opened 2017-05-02
Fix to #1241

yskoverride:
Corrections made to the path in "android/settings.gradle"


tonyhb:
Dupe of #1257, #1249, please close.

PR #1293, opened 2017-05-02
Add missing the `lib` folder in the docs

sroze:
The `lib` folder is missing from the documentation, adding it.


tonyhb:
Yup, this caught me too

christopherdro:
Thanks!

Issue #1225, opened 2017-04-10
Map from other tab on top of each other [Android]


Deevent:
I have two maps on different tabs in my app and when I go from a tab to another the previous map is still rendered on top of the new one. I've never had an issue like that. I tried to play with zIndex but no luck.

![ezgif-2-214d5f1d98](https://cloud.githubusercontent.com/assets/16573898/24863950/08d6f652-1e03-11e7-84ce-d7c1b673b2f4.gif)

react-native-maps : 0.13.0
react-native : 0.41.2

KarlosQ:
I have the same issue but only with react-native-maps 0.13.1 and 0.14.0. Some one know's way occurs?

Deevent:
Actually I'm using the last version `compile 'com.google.android.gms:play-services-maps:+'` but I've tried with other ones and also with 9.8.0 and my app crashed at startup. Same story if I exclude play-services from react-native-maps and compile them in my main build.gradle

KarlosQ:
Maybe the app crash because you have a conflict with other component (maybe react-native-google-places, or react-native-device-info) that use play-services and this component have different version. Put 9.8.0 in all component that use play-services. 

If you do like see why crash use bugsnag-react-native
The docs are here https://docs.bugsnag.com/platforms/react-native/

This service show you the crash error and is useful for production too because save the error in a server.

thylo:
I'm experiencing the same issue with the following stack :
- react-native 0.43.3
- react-native-maps 0.14
Every screen with a <MapView /> overlaps the other maps as if the zIndex was not right
@Deevent Did you manage to find a solution ?

Deevent:
@thylo Nope, I didn't find any solution. I ended up using images on my first screen because I had a short deadline.

aserrano1971:
I've used liteMode={true} and now I've a beatiful static map in Android but It has solved the problem for me.

ijharulislam:
I am facing the same problem and didn't find any solution.

Deevent:
@ijharulislam updating to 0.13.1 fixed the issue by my side.

cubbuk:
Having the same issue:
react-native 0.43.4
react-native-maps 0.14.0

Using `liteMode` on previous screen solves it, but I need the both map to be interactive, any solution to that?

samthui:
I have same problem.
Let say I have 2 scenes: A & B, both contain Map. Pushing B from A will display A's map on B's.
[...]

Issue #1161, opened 2017-03-24
[Android] Overlapping maps regression in v0.13.1


ryankask:
@lelandrichardson recently merged #1150 to deal with crashes and a memory leak on Android.

The problematic code changed in that PR was added to deal with a longstanding issues (#207 and #453) for Android users rendering more than one `MapView`, one of which isn't supposed to be visible on the screen and is hidden in a `View` with `opacity: 0` (for example).

@fdnhkj submitted #589 to fix the issue but now that the code has been removed.

Is there any way to fix the overlapping issue without causing memory leaks and/or crashes?

lelandrichardson:
@ryankask interesting. Thank you for this context. I'm going to be going through a lot of stuff this weekend for this repo and will try to make sure we get this fixed in a dot release.

Can you help me understand better how #589 fixed issue #453 ?  If you have any context that would be helpful.

We can add back the MapView.initialize(..) call if that was it. That code was not causing any issues, but per google's docs is "unnecessary" since we are cooling getMapAsync in the MapView itself. Perhaps it is what was fixing this bug and we can add it back in.

ryankask:
@lelandrichardson Unfortunatley I lack the Android knowledge to understand @fdnhkj's change. I couldn't find an explanation.

The heart of the issue is which context `MapView` is instantiated with. If I tweak the code to use the activity again, it works. The `initialize()` call doesn't appear to be be related.

Could the activity be used and if it's null, use `reactContext`?

#449 also relates to using the activity albeit in a different way. This PR was mentioned in #589.

Issue #453, opened 2016-08-09
Android: first instance of the map overlaps everything when have another map in a different scene.


kakbeya:
Hi guys, 

I know you have lots of issues already to work on, but here is another one.

I'm using RN 0.31 with latest maps and was finally able to get my map to show up on Android with a Browser Key.

I also use react-native-router-flux and I have two scenes with maps. When I switch to the second scene, the map from the first one is frozen on top of it. I can tap/drag the map behind it and everything, but can't see what I am tapping because of this lag. If I remove the map from the second screen - everything works fine.

Any ideas or workarounds?

Thanks, 
Alex.

jeveloper:
I filed a ticket for this , both here and in RNRF git, folks are taking it easy ,  i rolled back to rn 28 
tried many ways
if you do find a work around (basically demounting mapview so its completely gone when navigation state changes , please share.

thanks

kakbeya:
@jeveloper thanks,

I just was able to do it with changing my second scene type to:

**type="replace"**

Not the best solution as it removes the smooth transition, but works!

jeveloper:
Yeah i did observe that , but obviously when you want to stack navigation and want to have to a back arrow , it won't work.
Yes it also completely unmounts the view and mapview is actually a heavy component ( takes a bit of time to render on older devices)

morrigan:
+1, same with NavigationExperimental and RN 0.28

jeveloper:
@morrigan i believe 29 or 30 has a fix for it. 
By unresponsive do you mean that its basically appearing on top of a previous view?

danscan:
I'm having this same issue on a project that uses `react-native-router-flux`.  Adding `type="replace"` to the affected scenes works as a workaround.

jeveloper:
@danscan  can you confirm which version of rnrf you're using , latest is 34 and which RN 

thanks

danscan:
@jeveloper Sorry for the delay.  I'm using  `react-native-router-flux@3.30.9` and `react-native@0.29.2`

jeveloper:
No problem @danscan  , let me know if latest RNRF will work

danscan:
@jeveloper Updating to `react-native-router-flux@3.34.0` doesn't affect this issue.

olegpesok:
[...]

PR #1311, opened 2017-05-09
Fix overlay issues in Android introduced in 0.13.1

frmdstryr:
For #1225, #1161, and #453.


delvedor:
I tryied your solution and seems to fix the issue :)

PR #1323, opened 2017-05-11
Update from View.propTypes to ViewPropTypes to match RN v0.44.0

TBouder:
Update the ViewPropTypes to match the last revision of React Native V0.44.0

> View.propTypes to ViewPropTypes (53905a5) - @bvaughn

https://github.com/facebook/react-native/releases


RWOverdijk:
I think this may have broken the example. 

I'm getting an `undefined is not an object (evaluating '_reactNative.ViewPropTypes.style')

PR #1326, opened 2017-05-12
Add Marker rotation for Google Maps on iOS

Kerumen:
I added the rotation for the Google Maps provider on iOS. It was already working on Android.
I can't figure out how to do it with `MKAnnotationView` for the MapKit provider..

The rotation is in degrees, starting from 0 to 360.


henrikra:
Wow very nice! Is it possible to animate rotation?

Kerumen:
I didn't test it but it should work yes.

juergengunz:
@Kerumen it works fine for Mapview.Marker but when i use an animated Mapview.Marker.Animated, the rotation is not animated and it throws a warning to use an image with style transform instead...

PR #1340, opened 2017-05-17
Prevent Android crashes

dabit1:


christopherdro:
@dabit1 What sort of crashes does this prevent?

dabit1:
I don't know why but sometimes my app crashed when I loaded maps. Reading logcat I found that `map` var was null.

dabit1:
Besides, other methods in the same class checks if map var is null before do an actions, likely for the same reason.

christopherdro:
I don't see this check happening anywhere else in the class. Can you show me where you see this?

You can do a check for `ref` in JS before calling any of these methods.

liyuanhust:
Why this pr was not merged? This crash happens many times on my HTC devices.
Now I have to do some tricky method to prevent the crash.
Pls merge the pr !

Fatal Exception: java.lang.NullPointerExceptionAttempt to invoke virtual method 'void com.google.android.gms.maps.GoogleMap.setPadding(int, int, int, int)' on a null object reference Raw Text
--
  | com.airbnb.android.react.maps.AirMapView.fitToCoordinates (AirMapView.java:655)
  | com.airbnb.android.react.maps.AirMapManager.receiveCommand (AirMapManager.java:236)
  | com.airbnb.android.react.maps.AirMapManager.receiveCommand (AirMapManager.java:27)

christopherdro:
Thanks for submitting this and sorry for the late response. An updated PR has been submitted here. 

https://github.com/airbnb/react-native-maps/pull/1835

PR #1341, opened 2017-05-18
fix compile error in rn version >= 0.40

yosimasu:
Fix compile error in rn version >= 0.40


itsmepetrov:
@christopherdro could you please publish patch with this fix to npm?

christopherdro:
@itsmepetrov Just published 0.15.1

aaronbuchanan:
AIRGMSPolyline needs the same fix, right?

PR #1360, opened 2017-05-25
Add minZoom and maxZoom properties for android and ios

foyarash:
Adding ability to set minZoomLevel and maxZoomLevel for both iOS and Android (GMaps included for iOS).

Need maybe some review for iOS, as i did a dumb implementation of https://github.com/johndpope/MKMapViewZoom .


foyarash:
Well yeah which is quite logic. What will happen is that, if the required zoom to fit your markers is more than the maxZoomLevel, you will have your map at this maxZoomLevel (same case for minZoomLevel).

henrikra:
Then this PR is very good. What if I want to use `fitToMarkers` with different min zoom level? I guess this setting is global 🤔

foyarash:
This is a global setting for the map yeah, but something which has to be tested is to set the max/min possible zoom values in case you do a fitMarkers by settings the props with a state value.

foyarash:
PR updated, i forgot to add some things to the AIRMap.h, and GMaps for iOS was not working correctly

lib/components/MapMarker.js
```diff
@@ -257,7 +257,7 @@ class MapMarker extends React.Component {
     let image;
     if (this.props.image) {
       image = resolveAssetSource(this.props.image) || {};
-      image = image.uri;
+      image = image.uri || this.props.image;
```
christopherdro: What warrants this change?

lib/ios/AirGoogleMaps/GlobalVars.m
```diff
@@ -0,0 +1,54 @@
+//
+//  GlobalVars.m
+//
+//  Created by Eric Kim on 2017-04-04.
+//  Copyright © 2017 Apply Digital. All rights reserved.
```
christopherdro: not a legal expert but i'm guessing this would have to change

foyarash:
Corrected that, i got confused when merging some branches, but it's ok now

samthui:
I'm using react-native-maps 0.17.1. I set maxZoomLevel=10 and minZoomLevel=8, but I still can zoom in/out like nothing is limited. Does anyone have same problem?

marengga:
Strange behavior on Android.
You should place maxZoomLevel props in the end of MapView declaration to make it works.

```
<MapView
     style={{ flex: 1 }}
     showsUserLocation={true}
     showsMyLocationButton={true}
     mapType={this.state.mapType}
     ref={(ref) => { this.mapRef = ref; }}
     onLayout={() => this.mapRef.fitToCoordinates(markers)}
     maxZoomLevel={17}>
```
Maybe conflicts with onLayout props?

PR #1369, opened 2017-05-30
Add onMapReady callback

### LINKED ISSUE #246: On Map Ready
It would be really helpful to be able to specify an onMapReady callback, or listen for an event.

Use case:
I need to be able to `showCallout` on the Marker as soon as the map appears.

Workaround:
I'm currently using `setInterval` to check `this.refs.marker` until it's available, then calling `showCallout` and clearing the interval.  This is pretty nasty.

Kerumen:
Fix #246

### DIFF
--- a/lib/components/MapView.js
+++ b/lib/components/MapView.js
@@ -442,8 +442,9 @@ class MapView extends React.Component {
       this.map.setNativeProps({ region: initialRegion });
     }
     this._updateStyle();
-    if (onMapReady) onMapReady();
-    this.setState({ isReady: true });
+    this.setState({ isReady: true }, () => {
+      if (onMapReady) onMapReady();
+    });
   }
 
   _onLayout(e) {
--- a/lib/components/MapView.js
+++ b/lib/components/MapView.js
@@ -318,6 +318,11 @@ const propTypes = {
   legalLabelInsets: EdgeInsetsPropType,
 
   /**
+   * Callback that is called once the map is fully loaded.
+   */
+  onMapReady: PropTypes.func,
+
+  /**
    * Callback that is called continuously when the user is dragging the map.
    */
   onRegionChange: PropTypes.func,
@@ -430,13 +435,14 @@ class MapView extends React.Component {
   }
 
   _onMapReady() {
-    const { region, initialRegion } = this.props;
+    const { region, initialRegion, onMapReady } = this.props;
     if (region) {
       this.map.setNativeProps({ region });
     } else if (initialRegion) {
       this.map.setNativeProps({ region: initialRegion });
     }
     this._updateStyle();
+    if (onMapReady) onMapReady();
     this.setState({ isReady: true });
   }
 
--- a/docs/mapview.md
+++ b/docs/mapview.md
@@ -40,6 +40,7 @@ To access event data, you will need to use `e.nativeEvent`. For example, `onPres
 
 | Event Name | Returns | Notes
 |---|---|---|
+| `onMapReady` |  | Callback that is called once the map is fully loaded.
 | `onRegionChange` | `Region` | Callback that is called continuously when the region changes, such as when a user is dragging the map.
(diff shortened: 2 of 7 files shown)


lib/components/MapView.js
```diff
@@ -430,13 +435,14 @@ class MapView extends React.Component {
...
-    const { region, initialRegion } = this.props;
+    const { region, initialRegion, onMapReady } = this.props;
     if (region) {
       this.map.setNativeProps({ region });
     } else if (initialRegion) {
       this.map.setNativeProps({ region: initialRegion });
     }
     this._updateStyle();
+    if (onMapReady) onMapReady();
     this.setState({ isReady: true });
```
henrikra: I think you should call `onMapReady` when state has changed. So like this:
```
this.setState({ isReady: true }, () => {
  if (onMapReady) onMapReady();
});
```

cyakimov:
Seems that `onMapReady` never fires.

freshnewegg:
onMapReady doesn't seem to fire here.

I have a one that works for RN 0.40 (the only difference is new RN's have ViewPropTypes and b4 we had View.propTypes)

https://github.com/freshnewegg/react-native-maps/commit/8e079552e2d588270aa012b6e9a882abfefa3a9b

Cheers,

Kerumen:
It was only for the Android side. I will update the PR with the iOS part. Thanks @freshnewegg!

viktorsec:
@Kerumen do you have an idea of how to fix this on iOS?

Kerumen:
@viktorsec I've just updated the PR with iOS support! 🎉

christopherdro:
Thanks for the PR!
This looks specific to use google maps on iOS. What is the behavior when using the default?

Kerumen:
The callback will be called in the `mapViewDidFinishRenderingMap` delegate ([source](https://github.com/Kerumen/react-native-maps/blob/978b08dd188e64bbc6eea5cdc5a85fbd8a8dc87f/lib/ios/AirMaps/AIRMapManager.m#L656)).

lib/components/MapView.js
betoharres: is there a reason to why `this` is not being passed as an argument to `onMapReady()`? 
refs are set in a arbitrary moment during react renders, and devs does not have control when they are set, thus outputting different results in some of initial renders. By passing the `this` as an argument that would be fixed and we would have the map ref exactly when the map is ready every time. 

May I create a PR for this?

rborn: @betoharres I don't think you need to do it, the function is bound already - https://github.com/react-native-community/react-native-maps/blob/master/lib/components/MapView.js#L539

betoharres: @rborn thanks for the reply. Unfortunately no arguments are being passed on that callback, I edited the file just by modifying that line to `if (onMapReady) onMapReady(this);` and I got the result as I expected.

I've ran into this problem before and tried to workaround, but the results, as I said in my previous comment, are unstable; sometimes the ref is set, sometimes not.
My main goal is to run `mapRef.animateToRegion(userLocation)` when the map is ready, so that's why I'm interested on that change. 

I have a PR ready to send if you want(besides, I'm just adding a word :laughing: )

rborn: ref should be set, if is not set you are doing something funny :)

betoharres: Thanks for the reply again! I did some testing and you're right! If you only render the MapView component, it does have the ref set on every `onMapReady()` call, but if I add the [useNetIfno()](https://github.com/react-native-community/react-native-netinfo) hook call to my component, the ref starts to return `null` sometimes when `onMapReady` fires. 
If you are interested, I can open a issue and try to push a example project, because maybe something else could be interfering(even though I'm only rendering the Map with `react-navigation`).
I still like the idea of having an event available to set the ref, I'm not sure how it would affect the current code, but it doesn't seems to hurt I guess :sweat_smile:

Issue #418, opened 2016-07-24
Markers with custom Views are positioned above Callouts


deckchairlabs:
When rendering a relatively dense cluster of markers, tapping one to reveal the callout causes the custom markers to be displayed above the callout and it's contents.

Is there no way to change the zIndex of the markers?

1ne8ight7even:
Same problem here, but only on ios. Android is working fine.

attila135786:
Is there any update on this? I'm seeing the same problem.

chadwilken:
I am also seeing this issue on iOS. Running React Native 0.32.1 and react-native-maps 0.8.0.

MadeInMoon:
Did someone succeeded to handle this problem?

sandman45:
I see the same.. pretty annoying 
"react-native": "^0.41.2",
 "react-native-maps": "^0.13.0",

maggialejandro:
This should solve the issue https://github.com/airbnb/react-native-maps/pull/1405

alvelig:
#1845 should solve this problem. Please comment if you tested and if there are any bugs detected, as collaborators ask me to have positive replies on my pull request to merge it to master.

sircEUNHO:
@alvelig This doesn't work for me. My react native maps version is 0.21.0 and the codes on the attempts to fixes are on the package. But callouts are still behind markers and weird thing is that when I click outside the callout, the callout goes in front of the marker. By the way, my markers are custom image markers.

christopherdro:
I’m closing this issue because it has been inactive for a while. This probably means that it is not reproducible or it has been fixed in a newer version. If it’s an enhancement and hasn’t been taken on for so long, then it seems no one has the time to implement this.

Please reopen if you still encounter this issue with the [latest stable version](https://github.com/react-native-community/react-native-maps/releases).

Thank you!

aitorct:
This is still happening. I've tried upgrading to the latest version of the library (0.26.1), no luck. The issue has not been solved yet.

PR #1405, opened 2017-06-14
Fixed issue where callout is presented behind markers on iOS

madapper:
(https://github.com/airbnb/react-native-maps/issues/1393)

### DIFF
--- a/lib/ios/AirMaps/AIRMapMarker.m
+++ b/lib/ios/AirMaps/AIRMapMarker.m
@@ -157,6 +157,8 @@
         // no callout to show
         return;
     }
+    
+    [self.superview bringSubviewToFront:self];
 
     [self fillCalloutView:self.map.calloutView];
 


tpisto:
Unfortunately this is not fixing the issue where the markers are react views with numbers. Would be wonderful to solve this, since now it's kind of trouble for us. Can anyone give any pointers where to start to look at this - I can also spend some time to figure this out.

calloutunder

madapper:
@tpisto Thank you for your feedback. I have looked further into this with the fix added and have replicated the error by implementing a clustering system that re adds new markers based on a bias. This doesn't always happen, but it happens in the same place each time. (Unfortunately, I got too excited that my fix fixed it!) 

The issue seems to be where the annotations are being added as subviews. See example below:

**Working:**
airmapmarker-good-map
airmapmarker-good-list

**Not Working:**
airmapmarker-bad-map
airmapmarker-bad-list

In the working example, the annotations are being added directly to the AirMap as annotation subviews. However, in the not working example, the annotation that is sitting over the top seems to have been added to an MKNewAnnotationContainerView, which is then added to the _MKMapContentView. There are two potential solutions to this that I can see on first glance:

1) Review the process for how the annotations are added. This could be a longer process in determining why the makers are being added in this way, but the general fix will be better in the long term.
2) Implement a workaround that will find the map view as a recursive function and shift the hierarchy around to suit the needs of the markers. This solution seems to be hacky and I am not sure it will work consistently. However, it is a quicker solution to implement if it works. 

I am happy to take a look at either option for its efficacy. But also throwing it out to the community if they have any thoughts that might improve the solutions.

kirakik:
Any news on this being merged?

xiaoTuiMao:
+1

alizahid:
@madapper Another thing I noticed; if the callout is already open, and new markers are added and overlap, they will show above the callout. So maybe call `bringSubviewToFront` again after new makers are rendered?

nbonamy:
+1

markusguenther:
Please merge it. It is more or less a no brainer, so why not merge it?

wwwemree:
+1

cinder92:
this merge does not fix the issue

captura de pantalla 2017-10-23 a la s 18 06 24

martsie:
+1 please merge this

anandkumar144:
Still not working

alvelig:
Does not fix. But I managed to find a workaround: https://github.com/airbnb/react-native-maps/issues/865#issuecomment-346980723

olofd:
Did not work for me. Using this instead, works well : #1843

alvelig:
@olofd Unfortunately, it did not work for me, as I am using supercluster and have markers that appear after callout is shown.  :(  And these appear above the callout. I'm struggling with the issue for a week, but I'm very weak at Objective C and MapKit :( Looks like these is due to the order of reactElements :(

olofd:
@alvelig. Ah. Ok. I’ll see if I can make it work for markers that appear when a callout is open.

alvelig:
Look, my approach was to add callout as a new annotation to the end of the stack. Calling `mapview:addAnnotation:` for a newly allocated MKPointAnnotation with self coordinates and initializing PinView resulted in the pinView appear behind the current marker.

olofd:
@alvelig send a small code example of what you’re doing or that demonstrates your issue!

alvelig:
@olofd need your help. It looks like I found the way to solution. 

I noticed that the callout is in MKNewAnnotationContainer, but the markers above it are direct children of AirMaps, added in AirMaps.m:
```
- (void)insertReactSubview:(id<RCTComponent>)subview atIndex:(NSInteger)atIndex {
...
    [_reactSubviews insertObject:(UIView *)subview atIndex:(NSUInteger) atIndex];
}
```
Commenting that line all annotations appear in MKNewAnnotationContainer and the callout is before all the markers.

Of course, this is only a way to go, as `_reactSubviews` array is used to detect removed elements if markers change in React. But it looks like I found a way to go and correctly deal with this undocumented feature 🤣

alvelig:
Eureka: https://github.com/airbnb/react-native-maps/issues/1844:

Just override `- (void)addSubview:(UIView *)view` in AirMap.m:

```
- (void)addSubview:(UIView *)view {
    if([view isKindOfClass:[AIRMapMarker class]]) {
        [self addAnnotation:(id <MKAnnotation>)view];
    } else {
        [super addSubview:view];
    }
}
```

alvelig:
#1845 should solve this problem. Please comment if you tested and if there are any bugs detected, as collaborators ask me to have positive replies on my pull request to merge it to master.

alvelig:
This does not solve the problem. Follow #418

PR #1440, opened 2017-06-27
Upgrade Android and iOS examples to RN 0.45.1

gpeal:
@lelandrichardson


example/examples/EventListener.js
```diff
@@ -7,7 +7,7 @@ import {
   ScrollView,
 } from 'react-native';
 // eslint-disable-next-line max-len
-import SyntheticEvent from 'react-native/Libraries/Renderer/src/renderers/shared/stack/event/SyntheticEvent';
+import SyntheticEvent from 'react-native/Libraries/Renderer/src/renderers/shared/shared/event/SyntheticEvent';
```
lelandrichardson: this will make it so the example can only be run with RN 0.45+

PR #1443, opened 2017-06-30
[Android Crash] skip region monitoring if map object is null

mattshen:
This happens if google play services version is lower than required. It is to fix issues#618

There are also other places checking if `map == null` before calling `startMonitoringRegion` or `stopMonitoringRegion` usually followed with another call, e.g.
```java
public void animateToRegion(LatLngBounds bounds, int duration) {
    if (map != null) {
      startMonitoringRegion();
      map.animateCamera(CameraUpdateFactory.newLatLngBounds(bounds, 0), duration, null);
    }
  }
```

However, both ways of check map's nullability are lack of elegance.


dramaking:
Other region related method like "fitToCoordinates" need to do a null check as well. Otherwise the following error will come out:
"CameraUpdateFactory is not initialized"

Issue #718, opened 2016-10-21
duplicate symbols for architecture x86_64


skylarmb:
After dragging the `AirMaps` and `AirGoogleMaps` folders into my xcode project and running `react-native run-ios`, I get 

```
ld: 175 duplicate symbols for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

** BUILD FAILED **
```

Full output here: http://pastebin.com/td7gRy9g

Anyone else have this issue or just me?

jojonarte:
Have you managed to solve this? I also have this issue when trying to run it for adhoc on a device.

palmsey:
I'm running into this too. Were either of you able to resolve it yet?

jojonarte:
@palmsey nope it wasn't resolved yet. I'm still waiting for other people who might have encountered this too and resolved. And I've been trying things but to no success.

skylarmb:
No, I haven't resolved it. I am just using Apple maps on iOS for now. feelsbadman.jpg

jojonarte:
I have resolved it, this is my pod file below
```

# Uncomment this line to define a global platform for your project
# platform :ios, '9.0'

target 'App' do
  # Uncomment this line if you're using Swift or would like to use dynamic frameworks
  # use_frameworks!

  # Pods for NearBy
  
  
  pod 'React', path: '../node_modules/react-native', :subspecs => [
  'Core',
  'RCTActionSheet',
  'RCTGeolocation',
  'RCTImage',
  'RCTLinkingIOS',
  'RCTNetwork',
  'RCTSettings',
  'RCTText',
  'RCTVibration',
  'RCTWebSocket'
  ]

 
  pod "HockeySDK", :subspecs => ['AllFeaturesLib']
  pod 'GoogleMaps'
  


end
source 'https://github.com/CocoaPods/Specs.git'
platform :ios, '8.0'
use_frameworks!


```
- First I made sure that react-native link goes on without an error.
- Do the pod install with the above podfile edited with right project
- Moved AirGoogleMaps and AirMaps to ios project by dragging.
- Removed Airmaps.xcodeproj and libairmaps if it exist in Libraries folder and Build Settings respectively.

mersay:
i have the exact same error too, also on xcode 8

mersay:
ok i successfully installed the whole thing by react-native init a new project. I use option 1, I'm on xCode 8, RN 0.36 and node v7, you have to be very careful with each step. 
[...]

PR #1448, opened 2017-07-03
Reference install solution in issue #718 in install docs

se1exin:


christopherdro:
Closes https://github.com/airbnb/react-native-maps/issues/718

PR #1463, opened 2017-07-11
Fix pointer-indirection.

mikelambert:
min/maxZoomLevel are floats, not pointers. I think this is what was semantically intended here (to check if the value is intentionally set)


christopherdro:
Thank you for submitting this PR but it appears similiar changes were merged with this PR. https://github.com/airbnb/react-native-maps/pull/1485

Issue #1356, opened 2017-05-21
setPadding support is missing


zefj:
https://developers.google.com/maps/documentation/android-api/map#map_padding

Related issue: https://github.com/airbnb/react-native-maps/pull/239

This has been mentioned (and implemented) before, yet closed and left unmerged due to the examples project being broken. I haven't been able to run this library examples either, however I have tested the setPadding implementation provided in the aforementioned issue, and it works flawlessly. This is an important feature for every project that uses MapView with overlays, how (or why) has this been overlooked?

Mrachko:
Absolutely agree. We have a panel that slides from the bottom and we can't implement such an easy functionality without violating terms of use (which is not ok for us). Please, have a look at this problem as soon as it's possible. Thanks in advance.

benvium:
I've added a new PR for this #1467

juergengunz:
@zefj what did you end up doing? did you manage to make it work on android and ios?

zefj:
@juergengunz I just used the changes proposed in #239. It seemed to work on Android, don't know about iOS. It was just a small weekend project to see what react-native was about, I have long abandoned it now and haven't been using this library for a while so I'm out of the loop, sorry.

PR #239, opened 2016-05-18
Adding support for innerPadding prop (GoogleMap.setPadding())

julianocomg:
Sometimes we need a padding inside the map container.
http://stackoverflow.com/questions/6923472/relocate-google-logo-in-mapview

PR #1467, opened 2017-07-11
Fix android edgePadding

benvium:
We found that the edgePadding value while using map.fitToCoordinates was not being taken into account on Android.

Sample:

```js
this.map.fitToCoordinates(coordinates, {
      animated: animate,
      edgePadding: {
        top: Math.ceil(props.markerPadding.top + this.mapLayout.height * props.viewport.top),
        right: Math.ceil(props.markerPadding.right + this.mapLayout.width * props.viewport.right),
        bottom: Math.ceil(props.markerPadding.bottom + this.mapLayout.height * props.viewport.bottom),
        left: Math.ceil(props.markerPadding.left + this.mapLayout.width * props.viewport.left),
      },
    });
```

On iOS this was working as expected, but on Android the edgePadding values made no difference.

Looking at the source, the user's edgePadding value was immediately overwritten with 0,0,0,0.  
The intention of this seems to be to prevent the Google Logo from moving upwards based on the bottom padding, however (a) this made edgePadding pointless and (b) the Google Logo moving when padding is set is an intentional part of the Google Maps API (see https://developers.google.com/maps/documentation/android-api/map#map_padding)

Additionally the screen's density was not taken into account (thanks to #239 for inspiration on the second point). 

iOS and Android edgePadding now works the same.

Related to issue #1356,  and PR #239

### DIFF
--- a/docs/mapview.md
+++ b/docs/mapview.md
@@ -67,6 +67,7 @@ To access event data, you will need to use `e.nativeEvent`. For example, `onPres
 | `fitToElements` | `animated: Boolean` |
 | `fitToSuppliedMarkers` | `markerIDs: String[]`, `animated: Boolean` | If you need to use this in `ComponentDidMount`, make sure you put it in a timeout or it will cause performance problems.
 | `fitToCoordinates` | `coordinates: Array<LatLng>, options: { edgePadding: EdgePadding, animated: Boolean }` | If called in `ComponentDidMount` in android, it will cause an exception. It is recommended to call it from the MapView `onLayout` event.
+| `setPadding` | `edgePadding: EdgePadding` | **Note**: Android only.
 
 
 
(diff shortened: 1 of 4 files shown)


gpeal:
@benvium Thanks for doing this. I'll merge this if you resolve the conflicts.

juergengunz:
this is not working when you set edgePadding on the MapView on iOS. Would be useful to have in on iOS Google Maps too. .fitToCoordinates works fine but you should also be able to set it as a prop to the map.

peterjuras:
I've opened a new PR for to add setPadding support via the "mapPadding" prop:

https://github.com/airbnb/react-native-maps/pull/1750

christopherdro:
Hey thanks for submitting this. Just merged https://github.com/airbnb/react-native-maps/pull/1750 that addresses the issue across both platforms.

Issue #1473, opened 2017-07-13
Current version not compatible with RN < 0.44 (ViewPropTypes)


paulmasters:
ViewPropTypes replaced View.PropTypes in (I think) RN 0.44 
The peerDependency in the package.json says it's compatible with >=0.40 which is wrong.
Would it be possible to test for the existence of ViewPropTypes, and fall back to View.PropTypes to still support older versions?

alvelig:
I recently made a fresh example build.
https://github.com/alvelig/react-native-maps-example.
Try to check it out or build through the README.MD. If it does not help, let me know.

PR #1474, opened 2017-07-13
Fall back to View.propTypes if ViewPropTypes is not available (#1473)

paulmasters:
This is to provide support for RN <0.44 where ViewPropTypes was introduced


CristianPQ:
That has to be added on each component, if not it's going to fail on using any component

PR #1481, opened 2017-07-16
Handle Android RN 0.47 breaking change

Benjamin-Dobell:
https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/MapsPackage.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/MapsPackage.java
@@ -24,7 +24,7 @@ public class MapsPackage implements ReactPackage {
     return Arrays.<NativeModule>asList(new AirMapModule(reactContext));
   }
 
-  @Override
+  // Deprecated RN 0.47
   public List<Class<? extends JavaScriptModule>> createJSModules() {
     return Collections.emptyList();
   }


lib/android/src/main/java/com/airbnb/android/react/maps/MapsPackage.java
```diff
@@ -24,7 +24,7 @@ public MapsPackage() {
     return Arrays.<NativeModule>asList(new AirMapModule(reactContext));
   }
 
-  @Override
+  // Deprecated RN 0.47
```
henrikra: I would remove the `@Override` part instead of the comment. Since comments get old very fast

Benjamin-Dobell: @henrikra I believe the comment is important as it explains why this code is here at all. It's actually 100% dead code in RN >= 0.47.

So what this comment means is:

> Feel free to use this in your own projects, but the method is deprecated as of React Native 0.47.

Now I could have used a Java `@Deprecated` annotation, but I believe that's actually misleading in this circumstance. `@Deprecated` is designed to provide warnings to end-users that they're calling/using code that they shouldn't be using anymore. However, if you're still using RN < 0.47 this method is _required_, there is no alternative. You don't want compiler warnings complaining you're using a deprecated methods.

The real alternative to this pull request is not just to delete `@Override`, as that's confusing and leaves around dead code without an explanation; the alternative is to simply delete the method entirely. However, that's problematic as it would means user's of `react-native-maps` can't update (obtain fixes/patches) _unless_ they also upgrade React Native to >= 0.47, which depending on the project may be a large undertaking.

This I believe the succinct comment is the best option.

henrikra: I agree on this! :)

christopherdro:
@lelandrichardson @felipecsl @spikebrehm  Thoughts on the best way to move forward?

Issue #1031, opened 2017-02-07
Custom marker in google map causes high CPU load on iOS


vinceyuan:
Even if there is only one custom marker in google map, the app uses a LOT of CPU on iOS. It happens on both simulator and device with debug/release builds.

You can see in the screenshot of the [demo project](https://github.com/vinceyuan/CustomMarkerIssueDemo) I made, when we use custom marker, the CPU usage jumps from 0.8% to 62%. I also tried creating a custom marker in a native iOS app using Google Maps SDK for iOS. It does not use much CPU.

The following two ways of defining a custom marker do not make any difference. Both of them use a lot of CPU.
```
<MapView.Marker coordinate={coordinate} anchor={{x: 0.65, y: 1}}>
  <Image source={imageSrc}/>
</MapView.Marker>
```

```
<MapView.Marker coordinate={coordinate} anchor={{x: 0.65, y: 1}} image={imageSrc}/>
```

![Screenshot](https://github.com/vinceyuan/CustomMarkerIssueDemo/raw/master/CustomMarkerIssueDemo.gif)

I tried to use Instruments to profile the app. For unknown reason, something (the custom marker i guess) **keeps redrawing**.

![Screenshot](https://github.com/vinceyuan/CustomMarkerIssueDemo/raw/master/Profile.png)

Environment:
react-native-maps: 0.13.0
react-native: ^0.40.0
map provider: google
OS: iOS

Another person encountered the same issue. https://github.com/airbnb/react-native-maps/issues/741#issuecomment-263400988

vinceyuan:
Can anybody take a look at this issue? It happens in React Native project only. Creating a custom marker using native google map sdk does not result in high CPU load.

tema87:
I have the same problem with google&ios

ryancoughlin:
Same issue on the latest build.

grundmanise:
Same issue even with a super 'light' custom marker that contains empty `View`. iOS FPS drastically drops and other native UI elements (ScrollView) are lagging.

Marquis-H:
Same issue on the latest build.

mtt87:
Same here, did anyone find a workaround other than using Apple Maps?
Unfortunately I need custom styles 😢

MathieuDebit:
Still an issue here... Agree with Grundmanise, seems to be an important one.

zheng-liu-seattle:
still have this issue, any update on it?

compojoom:
[...]

PR #1491, opened 2017-07-19
Expose tracksViewChange and icon property to improve some performance issues

PhamBaTho:
Expose [tracksViewChanges](https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_marker.html#abaca10e1f2ba33aa1a10d7014269e780) property of GMSMarker. By default, this property is YES. 

Setting `tracksViewChange` to NO appropriately can solve **some performance issues** causing high CPU load on iOS (e.g. #1031 , #1176).

The followings are some guidelines relating to `tracksViewChange` I extracted from [Google Maps SDK documentation](https://developers.google.com/maps/documentation/ios-sdk/marker):

> - The UIView can be demanding on resources when tracksViewChanges is set to YES, which may result in **increased battery usage**. In comparison, a single frame UIImage is static and does not need to be re-rendered.
> - **Some devices may struggle to render the map if you have many markers on screen**, and each marker has its own UIView, and all markers are tracking changes at the same time.
> - To decide when to set the tracksViewChanges property, you **should weigh up performance** considerations against the advantages of having the marker redrawn automatically. For example:
>   * If you have a series of changes to make, you can change the property to YES then back to NO.
>   * When an animation is running or the contents are being loaded asynchronously, you should keep the property set to YES until the actions are complete.
> 

Please note that this property has no effect if iconView of marker is nil. In other words, I don't think we have any issue with performance if we use default marker.

I have tested on iOS simulator with many use cases like adding many markers on maps, drawing polyline between markers, animating custom markers... and the CPU remains in good health when the map is idle.

Sample usage:
```
<MapView.Marker
   tracksViewChanges={ false }
   anchor={ { x: 1, y: 1 } }
   coordinate={ {
      longitude: location.longitude,
      latitude: location.latitude,
   } }>
   { this.renderMarker() }
</MapView.Marker>
```


PhamBaTho:
UPDATE:
I have just updated this PR a bit. I'm exposing additionally [icon](https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_marker.html#af709ab9605990b2a1c616be8c88eddbc) property. It's equivalent to `icon` property of GMSMarker Class. 

The `image` property is great, it can cover most of the use case. Though, I believe that using `icon` property still has an advantage over `image` property in term of performance, battery usage... because it doesn't trigger tracksViewChanges which [has no effect if iconView is nil](https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_marker.html#abaca10e1f2ba33aa1a10d7014269e780).

We still need to have a better naming between `icon` and existing `image` to make them less confusing.

lib/components/MapMarker.js
```diff
@@ -260,13 +273,20 @@ class MapMarker extends React.Component {
       image = image.uri;
     }
 
+    let icon;
+    if (this.props.icon) {
```
henrikra: I would refactor this to ternary if statement :)

heysailor:
Thanks for this. Could you give a full example of use with the icon property?

juergengunz:
i discovered that if i use <image> the images are not being displayed on google maps (ios)... when i use the image property it works fine but i am not able to resize the images

dragonfire1119:
Seems like the icon prop doesn't do anything? If I try to use a regular image in icon prop it just shows the regular red icon from google maps. Is it suppose to work with a Custom Image like a .png? Also I set tracksViewChanges={ false } & it didn't change anything. The CPU is still at 90%+

krabbypattified:
+1 would really like to see these optimizations in my app!

compojoom:
@PhamBaTho - how do I add your changes to an existing project? I tried directly modifying the AirGoogleMaps in the project in xCode, but then I get a compile error:
"Linker command failed with exit code 1"

docs/marker.md
```diff
@@ -17,6 +18,7 @@
 | `identifier` | `String` |  | An identifier used to reference this marker at a later date.
 | `rotation` | `Float` |  | A float number indicating marker's rotation angle, in degrees.
 | `draggable` | `<null>` |  | This is a non-value based prop. Adding this allows the marker to be draggable (re-positioned).
+| `tracksViewChanges` | `Boolean` |  | Controls whether the icon for this marker should be redrawn every frame.
```
gpeal: @PhamBaTho Is this prop iOS only?

lib/components/MapMarker.js
```diff
@@ -50,6 +50,14 @@ const propTypes = {
   image: PropTypes.any,
 
   /**
+   * Marker icon to render (equivalent to `icon` property of GMSMarker Class).
+   * Using this property has an advantage over `image` in term of performance, battery usage...
+   * because it doesn't trigger tracksViewChanges.
+   * (tracksViewChanges is set to YES by default if iconView is not nil).
+   */
+  icon: PropTypes.any,
```
gpeal: @PhamBaTho Can you give this a more specific prop type?

lib/components/MapMarker.js
gpeal: +1

docs/marker.md
satori-ytolstoguzov: Only applicable for iOS. Android don't have issues with that.

satori-ytolstoguzov:
At what stage this PR? I use a fork with `tracksViewChanges` in my project for at least half a year while waiting to be merged.
@PhamBaTho Are you going to fix comments?

rborn:
@satori-ytolstoguzov if you could address the comments (if @PhamBaTho won't) I can have a look

compojoom:
@rborn @satori-ytolstoguzov @PhamBaTho - we've merged this back in November: https://github.com/react-community/react-native-maps/pull/1705 and it exposes tracksViewChanges. Shouldn't we close this pull request? Or what is the difference here with pull request 1705?

satori-ytolstoguzov:
@compojoom It seems merged PR forgot to add that to the docs. Other than that I don't see any differences

rborn:
@satori-ytolstoguzov could you test if it's working and make a PR for documentation? I'll be here to merge it ❤️

satori-ytolstoguzov:
@rborn It's a busy day, will look into that on Monday. Also would you consider useful if I add one more bridged property: `tracksInfoWindowChanges`. It's only for iOS and without it Google Maps SDK ignores all changes in callout after first render.

pqminh:
i'm testing on maps provider google and default.
rn 0.47.2
react-native-maps 0.17.1 | 0.19.0. 
All of them don't working. I'm render maps about 700 markers. When use provider google, it's so slow and lag. And use provider default. It does not slow, but when press on random marker, callout is show behind the markers around...

christopherdro:
trackViewChange prop has been added here: 78a36ac3727d9781187b008d255fc365b7873f94

PR #1521, opened 2017-08-01
fix rare android crashes when map size is 0

mattshen:
This is to fix https://github.com/airbnb/react-native-maps/issues/245

This cannot be easily reproduced, and only happens in our released app from time to time (not often). 

The safest way is to check width and height before calling `newLatLngBounds(boundsToMove, width, height, 0)`

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
@@ -541,16 +541,17 @@ public class AirMapView extends MapView implements GoogleMap.InfoWindowAdapter,
     // a proper camera move
     if (boundsToMove != null) {
       HashMap<String, Float> data = (HashMap<String, Float>) extraData;
-      float width = data.get("width");
-      float height = data.get("height");
-      map.moveCamera(
-          CameraUpdateFactory.newLatLngBounds(
-              boundsToMove,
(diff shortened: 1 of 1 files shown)


jeveloper:
I too have seen this in production, i think it might be related to google play service sdk bug , but thanks for doing this . Hope AirBNB will finally start pulling in code and making builds

sytolk:
Put Map in https://github.com/jaysoo/react-native-parallax-scroll-view and issue https://github.com/airbnb/react-native-maps/issues/842 will not so rare :)

samthui:
Hi, this issue re-appears after upgrading to RN 0.43.3.

Previously I used RN 0.42.3 & react-native-maps 0.13.1. My map is included in a View with flex=1.
Before upgrading to RN 0.43.3, this code worked fine:
```
<MapView
        provider={provider}
        ref="map"
         rotateEnabled={false}
         maxZoomLevel={18}
        style={{ flex: 1, width: "100%" }}
         initialRegion={this.state._region}
         onRegionChange={region => this.onRegionChange(region)}
/>
```

But after upgrading, same error occurs, only changing `style={{ flex: 1, width: "100%" }}` to `style={{ width: some-hard-code-value, height: "100%" }}` can remove the error, but that's not what I want.

kozillla:
crash in RN 0.48

PR #1527, opened 2017-08-04
Added [iOS / Google Maps] support for showsIndoorLevelPicker.

jessym:


christopherdro:
Thanks for the PR. Could you add this to the docs as well?

jessym:
Documentation updated as well now.

PR #1546, opened 2017-08-11
Fix initial region native prop

foyarash:
Fixes https://github.com/airbnb/react-native-maps/issues/1540 and https://github.com/airbnb/react-native-maps/issues/1507.

Sometimes it could happen that, when using initialRegion prop and moving on the map, we could just come back to our initial region for no reason. That's because we never set the initialRegion prop in the native code, but we set the region prop, while our component has the initialRegion prop.

### DIFF
--- a/lib/components/MapView.js
+++ b/lib/components/MapView.js
@@ -450,7 +450,7 @@ class MapView extends React.Component {
     if (region) {
       this.map.setNativeProps({ region });
     } else if (initialRegion) {
-      this.map.setNativeProps({ region: initialRegion });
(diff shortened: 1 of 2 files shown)


christopherdro:
@foyarash Might be forced to revert his commit. It appears it's causing issues on Android.

foyarash:
@christopherdro yup go on, I'll work on a fix for Android tomorrow

Donclion911:
still cannot fix. Jump back to region issue. just wont jump back to initialRegion.

PR #1550, opened 2017-08-12
Fixed `Archive` configuration for iOS

eugenehp:
Remove `React` target from `Pods` xcproject.

As seen in https://github.com/facebook/react-native/issues/12814


eugenehp:
Thanks @christophedro 👍

PR #1563, opened 2017-08-17
Fix initial region android

foyarash:
Fix https://github.com/airbnb/react-native-maps/pull/1546 which was not working on Android


ngandhy:
The lines in MapView.js cause my map upon re-render to get reset to 0,0.  After switching back to "const region = this.props.region || this.props.initialRegion;" I no longer have the 0,0 reset issue. 

#1640

PR #1359, opened 2017-05-23
add ios default-map support overlay images (merge chocky335's branch and add iOS GoogleMaps support)

smellman:
I open new PR with bug fix https://github.com/airbnb/react-native-maps/pull/1279 .
Original PR is https://github.com/airbnb/react-native-maps/pull/1140 .

- supports overlay images both iOS MapKit and GoogleMaps.
- fix mirrored presentation of image bug in MapKit.
- ready to merge current master

Also it have this bug in MapKit:

- Show image just after tap screen, swipe, scale...

PR #1585, opened 2017-08-23
Android Ground Overlay support

isener:
Android native ground overlays support implemented. OverlayManager has
support also to transparency, but it's closed now.


isener:
Guy's for the iOS version; look here #1359 
I tried and both environments works right now. 

**NOTE: On iOS -> AirMaps does not support ground overlay fully. Use Google Maps on iOS.**

markmssd:
I've been trying to implement this for a while with no success. 🥇@isener, thanks a lot! Can't wait to try it out.

PR #1597, opened 2017-08-26
fix_fire_twice_onRegionChangeComplete

kesha-antonov:
Hello!

I guess I've found answer for fix these: https://github.com/airbnb/react-native-maps/issues/1596 & https://github.com/airbnb/react-native-maps/issues/1577

```onRegionChangeComplete``` fired again when next tiles are loaded.

According to these comments:
screen shot 2017-08-26 at 14 50 02
in this commit https://github.com/airbnb/react-native-maps/commit/8a86e9899dc5674516e5fea374ae307d59732bdc

### DIFF
--- a/lib/components/MapView.js
+++ b/lib/components/MapView.js
@@ -446,7 +446,7 @@ class MapView extends React.Component {
   }
 
   _onMapReady() {
-    if (this.state.isReady) return
+    if (this.state.isReady) return;
 
     const { region, initialRegion, onMapReady } = this.props;
     if (region) {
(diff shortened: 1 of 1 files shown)


draperunner:
This is good, and we use this in my team's project. But it introduces a problem when using `region` and moving the map programmatically. These changes https://github.com/airbnb/react-native-maps/issues/1577#issuecomment-324704828 seems to fix it for iOS, but not on Android.

rborn:

PR #1613, opened 2017-09-01
[MapView] revert initialRegion change

computerjazz:
Reverts JS changes made in #1563 which broke initialRegion on iOS,
causing the map to display a region centered around 0, 0.

### DIFF
--- a/lib/components/MapView.js
+++ b/lib/components/MapView.js
@@ -466,13 +466,10 @@ class MapView extends React.Component {
     const { layout } = e.nativeEvent;
     if (!layout.width || !layout.height) return;
     if (this.state.isReady && !this.__layoutCalled) {
-      const { region, initialRegion } = this.props;
+      const region = this.props.region || this.props.initialRegion;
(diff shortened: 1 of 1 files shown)


christopherdro:
cc @foyarash

foyarash:
initialRegion still works with this? If so i'm ok with this

heysailor:
Still buggy with GoogleMaps on android and iOS: #1617

ngandhy:
@christopherdro it looks like this "revert" was merged in with https://github.com/airbnb/react-native-maps/commit/e6a3f7aedc12cbb5584cb7df638e8ee0c8847aef and then the merge was reverted minutes later with https://github.com/airbnb/react-native-maps/commit/0916cdff2764a29d0817d3ccc45541ecf411fa1c . Why was the fix #1613 reverted from master? It's still breaking iOS and causing the map to reset to 0,0.

This issue is still causing the 0,0 issue. See #1640

PR #1612, opened 2017-09-01
Remove legalNotice from android AirMapModule

TikiTDO:
As per https://developers.google.com/android/reference/com/google/android/gms/common/GooglePlayServicesUtil#getOpenSourceSoftwareLicenseInfo(android.content.Context) this information does not need to be displayed in app.

PR #1627, opened 2017-09-08
[BUGFIX] Add condition to use getOpenSourceSoftwareLicenseInfo

### LINKED ISSUE #1625: getOpenSourceSoftwareLicenseInfo is deprecated since GPS 11
Google deprecated getOpenSourceSoftwareLicenseInfo() that is used since the following commit.

[https://github.com/airbnb/react-native-maps/commit/75a38eab271b83d37a7f4089de806acac1a4865d](https://github.com/airbnb/react-native-maps/commit/75a38eab271b83d37a7f4089de806acac1a4865d
)
 
So if you are using play services in version 11 and more because you are maybe using also firebase services, you get a problem. I actually fixed it for by not using this constant.

And will discuss here maybe if this constant is relevant? Maybe we just check before using if this API call is available.

markusguenther:
The getOpenSourceSoftwareLicenseInfo function has been deprecated
for Google Play-Services since version 11 and leads to problems.

> getOpenSourceSoftwareLicenseInfo(Context context)
> This method was deprecated. This license information is displayed in Settings > Google > Open
Source on any device running Google Play services. Applications do not need to display this license text, and this method will be removed in a future version of Google Play services.

Therefore we should check if the GoogleApiAvailability let us get the
license information.

Fixes #1625

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
@@ -47,7 +47,12 @@ public class AirMapModule extends ReactContextBaseJavaModule {
   @Override
   public Map<String, Object> getConstants() {
     final Map<String, Object> constants = new HashMap<>();
-    constants.put("legalNotice", GoogleApiAvailability.getInstance().getOpenSourceSoftwareLicenseInfo(getReactApplicationContext()));
+    try {
+      constants.put("legalNotice", GoogleApiAvailability.getInstance().getOpenSourceSoftwareLicenseInfo(getReactApplicationContext()));
+    } catch (Exception e) {
+      constants.put("legalNotice", "");
+      Log.i("ReactMaps", "getOpenSourceSoftwareLicenseInfo is not available: " + e.getMessage());
+    }
     return constants;
   }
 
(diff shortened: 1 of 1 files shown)


kelset:
I've tried your solution but it crashes anyway :/

markusguenther:
Sry had no time in the last days ... already recognized that the try catch is not enough. At the moment I did not start to find out how to check the PlayServices version or the existence of the method. But will tackle that soon.

But have seen that someone also opened a PR and remove this constant.

markusguenther:
My workaround was actually to comment out the constants.put ... that worked for me while my current project.

markusguenther:
Ok will give this a try and then it hopefully is fixed für version greater 11 and also works for older versions.

lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
```diff
@@ -47,7 +48,17 @@ public String getName() {
...
+
+    // The function getOpenSourceSoftwareLicenseInfo has been removed in GooglePlayServices version 11
+    if (GoogleApiAvailability.GOOGLE_PLAY_SERVICES_VERSION_CODE < GOOGLE_PLAY_SERVICES_VERSION_CODE_11) {
+      constants.put(
+        "legalNotice",
+        GoogleApiAvailability
+          .getInstance()
+          .getOpenSourceSoftwareLicenseInfo(getReactApplicationContext())
+      );
+    }
```
kelset: what if here you add an `else` that handles it the way PR #1612 works?

Meaning doing something like:

```js
} else {
    constants.put("legalNotice", "This license information is displayed in Settings > Google > Open Source on any device running Google Play services.");
}
```

gpeal:
Closing in favor of #1612 
Google Play Services auto updates on all devices so it's not necessary to support legacy versions.

PR #1669, opened 2017-09-26
Add customPadding prop

kyledecot:
We need to be able to set the `padding` attribute of the native maps for both iOS and Android however `padding` is a reserved prop name so I've added this as `customPadding`.


juergengunz:
hey @kyledecot  is this working? did you mange do make it work on android too?

peterjuras:
I've opened a new PR for to add setPadding support via the "mapPadding" prop and adds Android support:

https://github.com/airbnb/react-native-maps/pull/1750

PR #1694, opened 2017-10-10
Remove createJSModules(depricated) method from MapsPackage.java file

VahidBo:
# Motivation 
According to [this document ](https://facebook.github.io/react-native/docs/native-modules-ios.html) of `react-native`  documentation, the createJSModules() method is deprecated and it is not needed. I removed this method from MapsPackage.java file.

# Test Plan
I removed this method from mentioned file and ran my app. There was no error and problem and my test app is working properly.

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/MapsPackage.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/MapsPackage.java
@@ -24,11 +24,6 @@ public class MapsPackage implements ReactPackage {
     return Arrays.<NativeModule>asList(new AirMapModule(reactContext));
   }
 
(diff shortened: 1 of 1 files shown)


rborn:
@christopherdro 🐽
I think it's ok to merge this one, right?

christopherdro:
Merging this will exclude older versions of react native from using newer versions of this library.
I forgot exactly in what version this was deprecated but I think we could still wait a bit longer before completely removing it.

This shouldn't be causing any build issues. @VahidBo Are you coming across any issue like that?

rborn:
@VahidBo thanks a lot for your PR, but we'll have to close without merging for the reasons @christopherdro mentioned 🤗

PR #1699, opened 2017-10-13
applied fix for pinch/zoom

jesdaigle:
This was originally done by sjmueller. I'm not trying to take credit for his work. :) Rather I had to fork master and apply his fix to that fork to fix an issue and would love to see it merged into master so we collectively don't have to keep doing this. Thank you @sjmueller for the fix. I hope it's merged soon.


chrusart:
Using classes private structures and some subviews positions is not a good idea. I know it was said as a temporary solution but I wouldn't do it.
I think you can accomplish the same thing setting:
```pointerEvents="none"``` on the marker.
Of course you will stop getting touch events for custom views and if you use onPress, it will not trigger as it's a gesture recognizer event which will not get touch events anymore, but onSelect will still work.
But letting the MKMapView decide what marker to select (on user tap) is a good idea cause it has good algorithm i guess to choose proper marker. I'm working on a PR to propose something and with some fixes for issues I found.

PR #1700, opened 2017-10-13
Redirect to a more up to date walkthrough article for google maps on ios

sraikimaxime:
Hello, I just set up google maps on ios and went through several issues with the current documentation whereas this one is perfectly up to date for react-native above 0.46 :) thought it might help !

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,15 @@ Then add the AirGoogleMaps directory:
 
 https://github.com/airbnb/react-native-maps/blob/1e71a21f39e7b88554852951f773c731c94680c9/docs/installation.md#ios
 
-An unofficial step-by-step guide is also available at https://gist.github.com/heron2014/e60fa003e9b117ce80d56bb1d5bfe9e0
+Unofficial guides are also available:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -207,7 +207,9 @@ Then add the AirGoogleMaps directory:
 
 https://github.com/airbnb/react-native-maps/blob/1e71a21f39e7b88554852951f773c731c94680c9/docs/installation.md#ios
 
-An unofficial step-by-step guide is also available at https://gist.github.com/heron2014/e60fa003e9b117ce80d56bb1d5bfe9e0
```
christopherdro: It appears the discussion thread around this gist is still active. I think it's a good idea to leave this here and add your article.

sraikimaxime: Thank you for your return. True, I left it and add there was an active discussion thread on it + add the step by step article and some details. Let me know if you have any other recommandation ! :)

rborn:
@sraikimaxime Thanks for the PR. I'll close it for now as we updated the docs. But if you still think they are not complete and we should amend them, please comment.
Thanks again 🤗

PR #1702, opened 2017-10-13
Update to IOS example Podfile 'Yoga' pod

adamwade2384:
Error: [!] The name of the given podspec `yoga` doesn't match the expected one `Yoga`


christopherdro:
Thanks for making this PR! Looks like this change got merged in with https://github.com/airbnb/react-native-maps/pull/1816

PR #1706, opened 2017-10-16
Blank map issue: Google Maps API not enabled?

abiduzz420:
I faced a similar problem of blank map but when I looked into Developer console I had not enabled the Google Maps API. After enabling the API, I ran the application on my android device and it worked. 
So, you need to enable Google maps API in the google developer console for android/ios. I would be happy if you could assign `hacktoberfest` label to this PR :)


hedgepigdaniel:
Thanks so much :+1:

Issue #1591, opened 2017-08-24
Is there a way to get the current region after fitToCoordinates?


kenma9123:
I'm having a difficulty getting the right deltas on two points(marker). Everytime I use `fitToCoordinates` on certain points and fire a `setState` to rerender. The old region is getting rerendered on the `mapView`.

I want to able to update the `region` state of my app everytime I call the `fitToCoordinates`.

Is there a way to get the current region progmatically? I know there's a prop on the `mapview` about `onRegionChange` but this only happens when you interact on the map not when you call a method like the `fitToCoordinates`

IjzerenHein:
This should be fixed by PR #1710

supryan:
+1

christopherdro:
I’m closing this issue because it has been inactive for a while. This probably means that it is not reproducible or it has been fixed in a newer version. If it’s an enhancement and hasn’t been taken on for so long, then it seems no one has the time to implement this.

Please reopen if you still encounter this issue with the [latest stable version](https://github.com/react-native-community/react-native-maps/releases).

Thank you!

Issue #1379, opened 2017-06-02
onRegionChangeComplete will not always fire on android


avdept:
```
<MapView loadingEnabled
                     showsMyLocationButton={false}
                     toolbarEnabled={false}
                     provider="google"
                     showsUserLocation
                     ref={ref => { this.map = ref; }}
                     style={[styles.map]}
                     initialRegion={this.state.coords.usingCoordinates}
                     onRegionChangeComplete={region => this.onRegionChangeComplete(region)}
                     onRegionChange={region => this.onRegionChange(region)}
                >


onRegionChangeComplete(region) {
        console.log("REGION HAS CHANGED:");
        if (Platform.OS == 'android' && region.latitude == 0 && region.longitude == 0) {
            return true; // This is fix for strange bug, when android throws 0 lat and 0 lng coordinates when initializing map
        }
    }
```

I didn't find any reason of why sometimes callback fires and sometimes not.

chapeljuice:
Just curious do you get any errors when you run `react-native log-android` or `adb logcat`?

Radu-Filip:
I'm seeing this issue when zooming. 
onRegionChange fires properly while onRegionChangeComplete doesn't ever fire when finishing zoom. 
Dragging the map seems to work as expected.

No errors or warnings in the adb logcat

Radu-Filip:
I was looking that for Android GoogleMap.OnCameraChangeListener is deprecated. 
Maybe is worth looking into adapting the implementation using GoogleMap.OnCameraMoveListener  and GoogleMap.OnCameraIdleListener?

sandropoluan:
+1

chadzheng:
+1

vinceyuan:
Same problem here.

cbartondock:
Same problem. Any fix for this without editing the native code? Seems like it should be fixed within react-native-maps

cbartondock:
Speaking of, I get the same issue on ios. onRegionChangeCompleted does not fire when zoom out finishes

chadzheng:
[...]

Issue #846, opened 2016-12-01
onRegionChangeComplete props Callback triggering infinitely without any user input on Android.


akshay2604:
i am not using fitToSuppliedMarkers as another issue mentioned, only plain MapComponent . For some reason on Android onRegionChangeComplete is being called infinitely many times. However it is working correctly on IOS.

React native - 0.35
react-native-maps - 0.10.4

akshay2604:
if you use region prop along with onRegionChangeComplete , the onRegionChangeComplete is called as a callback. onRegionChangeComplete is called whenever onChange event occurs natively. it could be due to user dragging map, animation on map or even setting up of gps location. If possible another prop should be added like onPanDragComplete so that we get callback only when user drags map and finishes dragging it.

M3po:
Did you make it work and how? i am also having problem with infinite loop

akshay2604:
`<MapView
          ref={(ref) => { this.map = ref; }}
          style={styles.map}
          initialRegion={this.props.initialRegion}
          loadingEnabled
          onRegionChangeComplete={(region) => {
            // do something
            }
          }}
        >
</MapView>
`
This is my new mapComponent. It turns out i was using region prop and was only setting state of region. removing it also works.
I don't know how to use both together. If you look at components/MapView.js there is a particular function
`_onChange(event) {
    this.__lastRegion = event.nativeEvent.region;
    if (event.nativeEvent.continuous) {
      if (this.props.onRegionChange) {
        this.props.onRegionChange(event.nativeEvent.region);
      }
    } else if (this.props.onRegionChangeComplete) {
      this.props.onRegionChangeComplete(event.nativeEvent.region);
    }
  }`

The _onChange method is called whenever any event happens.If whatever you write in onRegionChangeComplete **triggers a call to region prop** , region prop in itself triggers an _onChange event which causes onRegionChangeComplete to call. This quickly spirals to an infinite loop each callback trigering each other. You can put logs in MapView.js file to see it for yourself.
 
I think something like a panDragComplete in this case would do the trick. I am not sure if mapkit or google maps provider provide a native event for this or not. But until then make sure onRegionChangeComplete has code without any side effects. Or if you can't do that, remove region prop, it worked for me

M3po:
Thanks for the input i will try to edit the component so i might get something (or will at the do without region change until they make something like you suggested). You should leave this issue open.

meddy:
I have an app flow like so:
1. user enters map scene
[...]

Issue #459, opened 2016-08-10
OnCameraChangeListener() is deprecated


alexHlebnikov:
[In 0.9.4 version of the API](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap), Google replaced _GoogleMap.OnCameraChangeListener()_ with three camera listeners :

GoogleMap.OnCameraMoveStartedListener
GoogleMap.OnCameraMoveListener
GoogleMap.OnCameraIdleListener

IjzerenHein:
Fixed in PR #1710

PR #1710, opened 2017-10-17
Added support for new Android camera movement APIs

### LINKED ISSUE #1709: onRegionChange is constantly called on Android after touching map
This happens, when the following props are set:

```js
zoomEnabled={false}
rotateEnabled={false}
scrollEnabled={false}
pitchEnabled={false}
```

After inspecting of the Java code I found that `startMonitoringRegion` was called, but because the camera never changed, `stopMonitoringRegion` never got called. This caused `onRegionChange` to be called continuously until the camera was changed. But because no input was possible because of the props, the camera never changed. Even after the component was unmounted, that particular java code was still being executed.

IjzerenHein:
`OnCameraChangeListener` is deprecated and should no longer be used. This PR adds support for the new more powerful APIs and removes the old (quite hacky..) region detection code. This fixes various issues and paves the way for distinguishing between gestures and manual/programatic animations at the same time.

image
https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.OnCameraChangeListener

This PR fixes #1709 and I'm pretty sure it also fixes:
- #349
- #1379 
- #1591 
- #134 
- #846 
- #459

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
@@ -13,7 +13,6 @@ import android.support.v4.view.GestureDetectorCompat;
 import android.support.v4.view.MotionEventCompat;
 import android.view.GestureDetector;
 import android.view.MotionEvent;
-import android.view.ScaleGestureDetector;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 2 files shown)


IjzerenHein:
Any chance we can get this pulled?

marano:
Thank you this is great! It's a shame its not merged yet.

PR #1715, opened 2017-10-19
fixed region/initialRegion null overrides of this.props

atkit:
I noticed that 
props.region and props.initialRegion are always come as null. (even if set for `<MapView />`)

This change gives ability to set `region` and `initialRegion` to something that come from the top. 

Also it causes big performance issue on RN side when these props are null, (I'll try to mitigate that with different PR)

### DIFF
--- a/lib/components/MapView.js
+++ b/lib/components/MapView.js
@@ -659,10 +659,10 @@ class MapView extends React.Component {
         region: null,
         initialRegion: null,
         onMarkerPress: this._onMarkerPress,
-        ...this.props,
         onChange: this._onChange,
         onMapReady: this._onMapReady,
         onLayout: this._onLayout,
+        ...this.props,
       };
       if (Platform.OS === 'ios' && props.provider === ProviderConstants.PROVIDER_DEFAULT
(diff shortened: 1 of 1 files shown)


larryranches:
Can this get merged soon? This bug fix would help a lot as i'm also having issues with #1742

atkit:
probably someone from AirBnb can merge it

larryranches:
@atkit I did test your fixes and have confirmed that this fixes #1742. Great Work!

@gpeal I'm new to this library, how often is the release cycle or PRs get approved and merged?

aaronbuchanan:
@larryranches in my experience it can take quite a while for approved changes to make it in, I had to maintain my own fork to speed things up

rborn:
@atkit could you please update the PR (so the conflict goes away) maybe we can merge this? 💪
Thanks.

rborn:
LGTM @alvelig

rborn:
@christopherdro

christopherdro:
Should be just move the spread down to the bottom?

rborn:
@christopherdro I think yes, because otherwise the default will overwrite the spread.

PR #1716, opened 2017-10-19
Add missing flag at react-native-google-maps.podspec

alexeypodolian:
Changes: 
 - [RN] Update TypeScript typings
 - [iOS] Add missing flag at google-maps.podspec

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,156 +1,193 @@
 import * as React from 'react';
 
 interface MapViewProps {
-    provider?: 'google';
-    style: any;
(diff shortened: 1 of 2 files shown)


rborn:
@alexeypodolian thanks for the PR, few comments:

- the flag it's already there (from other merge I believe) 
- would you mind rebase with the current master and see if the TS typings are still needed?

As an aside, would be good to make PRs that handle one thing a a time , so in this case I would have done 2 PRs 🤗

PR #1746, opened 2017-10-31
Fix React Native 0.49 requiresMainQueueSetup

Titozzz:
Fixes Warning for react-native > 0.49

> Module AIRGoogleMapManager requires main queue setup since it overrides `constantsToExport` but doesn't implement `requiresMainQueueSetup. In a future release React Native will default to initializing all native modules on a background thread unless explicitly opted-out of.

### DIFF
--- a/lib/ios/AirGoogleMaps/AIRGoogleMapManager.m
+++ b/lib/ios/AirGoogleMaps/AIRGoogleMapManager.m
@@ -263,6 +263,10 @@ RCT_EXPORT_METHOD(takeSnapshot:(nonnull NSNumber *)reactTag
   return @{ @"legalNotice": [GMSServices openSourceLicenseInfo] };
 }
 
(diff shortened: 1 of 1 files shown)


kesha-antonov:
Please merge 👍   @gpeal

kesha-antonov:
I think now... Do we really need to setup this module in the main thread? Cause it's just static constants that will be used later.

I think we should init RN-maps in background:
```
(BOOL)requiresMainQueueSetup {
  return YES;
}
```

Or remove exportsConstants since legalNotice is deprecated. See my PR here: https://github.com/airbnb/react-native-maps/pull/1757

Titozzz:
Hey, I just used a non-breaking strategy since it was on the main thread before. Maybe we can effectively set NO but I'm not sure 😊

kesha-antonov:
I think if more libs will be initialized in background the faster your app will start 👍 @Titozzz

PR #1687, opened 2017-10-09
Fixed a TypeScript Error

PaitoAnderson:
Fix a bug I found after submitting PR #1601.

I was getting this error:
`[TS] Exports and export assignments are not permitted in module augmentations.`

Thanks!

PR #1759, opened 2017-11-05
Added style prop to MapCalloutProps in Typings

PaitoAnderson:
Fixes @marcmo issue in #1687


haxxpop:
I also need it in `MarkerProps` as well

PaitoAnderson:
@haxxpop Just curious, what kind of styles can you do directly on `<MapView.Marker>`?

Issue #1765, opened 2017-11-07
Animated Markers with useNativeDriver: true


kalininilya:
Due to low framerate (5-10fps) on android while using `MapView.Marker.Animated` I've tried setting `useNativeDriver` to `true` and got a following error: `This JS animated node type cannot be used as native animated node`

Is it possible to animate coordinates of `MapView.Marker.Animated` with `useNativeDriver` set to `true`? If not, Is there other ways to improve framerate without native animations?

Dependencies:
```
"react-native": "^0.50",
"react-native-maps": "^0.17.1"
```
Code:
```
import React, { Component } from 'react'
import {
  StyleSheet,
  View,
  Text,
  Dimensions,
  TouchableOpacity,
  Animated,
} from 'react-native'

import MapView from 'react-native-maps'

const screen = Dimensions.get('window')

const ASPECT_RATIO = screen.width / screen.height
const LATITUDE = 37.78825
const LONGITUDE = -122.4324
const LATITUDE_DELTA = 0.0922
const LONGITUDE_DELTA = LATITUDE_DELTA * ASPECT_RATIO

class Map extends Component {
  constructor(props) {
    super(props)

    this.state = {
      coordinate: new MapView.AnimatedRegion({
        latitude: LATITUDE,
        longitude: LONGITUDE,
      }),
    }
  }

  render() {
    return (
      <View style={styles.container}>
        <MapView
          style={styles.map}
          initialRegion={{
            latitude: LATITUDE,
            longitude: LONGITUDE,
            latitudeDelta: LATITUDE_DELTA,
            longitudeDelta: LONGITUDE_DELTA,
          }}
        >
          <MapView.Marker.Animated
            coordinate={this.state.coordinate}
          />
        </MapView>
        <View>
          <View pointerEvents="box-none" style={styles.buttonContainer}>
            <TouchableOpacity
              onPress={() => {
                this.state.coordinate.timing({
                  latitude: LATITUDE + ((Math.random() - 0.5) * (LATITUDE_DELTA / 2)),
                  longitude: LONGITUDE + ((Math.random() - 0.5) * (LONGITUDE_DELTA / 2)),
                  useNativeDriver: true,
                }).start()
              }}
              style={[styles.bubble, styles.button]}
            >
              <Text>Animate</Text>
            </TouchableOpacity>
          </View>
        </View>
      </View>
    )
  }
}

const styles = StyleSheet.create({
  container: {
    ...StyleSheet.absoluteFillObject,
[...]

PR #1782, opened 2017-11-13
Add native animation for Markers on Android

kalininilya:
Adds support for native marker animations on android. As of right now JS animations for multiple markers result in a poor performance as described at [#1765](https://github.com/airbnb/react-native-maps/issues/1765). This PR allows to animate up to 50 markers (depending on the device) with 60 fps on android.


christopherdro:
Thanks for submitting this. Can you also update the docs with this addition?

alvelig:
#1976 merged

TitanKing:
Would be awesome to have a rotation option too, most of the times with latlng animation one would need rotation too. Atm you still have to animate this with js.

PR #1828, opened 2017-11-29
[AirMapManager] Update MapBuilder for getCommandsMap to support all entries

christopherdro:
Related to https://github.com/airbnb/react-native-maps/pull/1587


christopherdro:
Max number of entires is 7 https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/common/MapBuilder.java#L107 Therefore we're required to split things up.

@codedre You didn't run into this issue?

PR #1832, opened 2017-11-29
Add CoordinateForPoint (iOS/Google)

brandon-hass:
Adds the ability to get a map coordinate from a point on the screen.


christopherdro:
Is it possible to add this for native iOS maps and Android as well?

danielgindi:
I just saw this... I created a full implementation and sent a PR a moment ago (#1851)

PR #1839, opened 2017-11-30
[AirGoogleMapManager] Use RCTDirectEventBlock for onMarkerPress

androidseb:
Closes https://github.com/airbnb/react-native-maps/issues/1829


kytwb:
@androidseb @christopherdro I think this is breaking MapView onMarkerPress?

PR #1843, opened 2017-12-02
Fix callouts appearing behind markers

olofd:
A bit ugly, but works well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## 0.18.3 (November 30, 2017)
+* Android: [#1839](https://github.com/airbnb/react-native-maps/pull/1839) [AirGoogleMapManager] Use RCTDirectEventBlock for onMarkerPress
+
 ## 0.18.2 (November 29, 2017)
 * Android: [#1835](https://github.com/airbnb/react-native-maps/pull/1835)  [AirMapView] Null check map instance on view methods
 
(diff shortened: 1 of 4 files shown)


olofd:
@alvelig Can you please supply a small js-example exposing the problems that you noticed with this approach? I tried adding in markes late, but it still worked here

alvelig:
I could not 100% reproduce that behavior. It just random with many markers, especially using `react-native-super-cluster`. I have about 4000 markers, so I use supercluster to show clusters of them. Some of clusters and markers appear before the callout, and then just move the viewport a little and it comes to front.

So I started with a MapKit tutorial and noticed that annotations are not added to MapView, but to another layer, but in react-native-maps AirMap extends RCTUIView+React, which makes it have views as reactSubviews. So I came to conclusion I should override addView, so that the views are stored within that class but not added as views.

GarimaMathur07:
@alvelig I have about 600 markers. So, i am using https://github.com/novalabio/react-native-maps-super-cluster uses MapBox's SuperCluster as clustering engine. I am facing **Callout behind markers** issue. I am using `renderMarker` as follows:

```
renderMarker (pin) {
	const styles = this.styleSheet();
	return (
		<Marker
			key={pin.id || Math.random()}
			coordinate={pin.location}
			ref={o => (this.marker = o)}
			calloutOffset={{ x: -8, y: 28 }}
			calloutAnchor={{ x: 0.5, y: 0.4 }}
			onCalloutPress={() => this.onPressCallout(pin)}
			image={map_pin.source}
		>
			<Callout tooltip style={styles.customCalloutView}>
				<LocationMarkerCallout title={pin.storeloc.name} description={pin.storeloc.address} />
			</Callout>
		</Marker>
	);
};
```
Can you please suggest how can i resolve this issue.

PR #1847, opened 2017-12-03
Added option for calloutContentViewInset to customize padding of callout

olofd:
Adds ability to set a or override the default (12, 12, 12, 12) calloutContentViewInset.
This way you can have a bubble without padding.


olofd:
By enabling this, markers can look like this:
[marker]: https://imgur.com/a/jPUfu "Marker with negative contentInset"

alvelig:
@rborn 🐽 
@allthetime thanks for the reminder. Worth mentioning that it's only for iOS

rborn:
@olofd or @allthetime could you please add  the needed docs? 
Thank you :)

rborn:
Closing as being an old issue and not a clear reply. Thanks for all the effort 😻

PR #1853, opened 2017-12-05
Fixed onMapReady no longer getting called on iOS

IjzerenHein:
Hi, this fixes onMapReady no longer getting called correctly in v0.18.x.

Due to this commit (https://github.com/airbnb/react-native-maps/pull/1797/files#diff-8a72baa08460d22c71989ddcbf76d49f), `onMapReady` was only called when the very first MapView was rendered for the very first time. After unmounting/mounting, or rendering any other MapView, onMapReady would be never called.

### DIFF
--- a/lib/ios/AirMaps/AIRMapManager.m
+++ b/lib/ios/AirMaps/AIRMapManager.m
@@ -36,9 +36,6 @@ static NSString *const RCTMapViewKey = @"MapView";
 @end
 
 @implementation AIRMapManager
-{
-    BOOL didCallOnMapReady;
-}
   
 RCT_EXPORT_MODULE()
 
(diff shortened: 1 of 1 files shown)


rborn:
@christopherdro ?

vendramini:
I'm using the latest version and `onMapReady` still not being called. Is this PR released?

Thanks!

emilefleming:
Same here, running 0.19.0 with Google Maps on iOS and it only fires on the first map load. InitialRegion is working now party but doesn't seem to respect the delta and is very zoomed out. @vendramini are you seeing that as well?

alvelig:
@vendramini are you using GoogleMaps iOS?

@emilefleming Could you please open an issue for that?

emilefleming:
@alvelig sure thing.

vendramini:
@alvelig no, I'm using apple maps.
@emilefleming no, didn't happens here.

alvelig:
@vendramini Sure, with npm you just get the same but through npm and released under some tag (version). Of course, it's better to keep npm version in package.json, because it gives you version control (exact version, not higher than, etc).

kytwb:
@IjzerenHein @alvelig this doesn't seem to work in the latest release? I still experience the following:

> onMapReady was only called when the very first MapView was rendered for the very first time. After unmounting/mounting, or rendering any other MapView, onMapReady would be never called.

IjzerenHein:
Seems this hasn't been released yet to npm, you should use the master branch from github until that is done

rborn:
@christopherdro do we need to push a new version to npm ? thnx cc: @alvelig

Issue #1859, opened 2017-12-07
onPress event for Polygon doesn't work on iOS


yatagan:
react-native-maps v0.18.3 (GoogleMaps v2.5)
On Android with the same project onPress works fine.

yatagan:
The reason is: there is an undocumented props tappable :(

PR #1861, opened 2017-12-07
#1859 onPress event for Polygon doesn't work on iOS

yatagan:
Because there is an undocumented props `tappable`


yatagan:
It's related to #1859

PR #1879, opened 2017-12-13
Fix podfile, example and setup instructions so the example runs

rborn:
@alvelig @christopherdro ping 🐽


example/ios/Podfile.lock
```diff
@@ -58,25 +58,25 @@ DEPENDENCIES:
...
-  react-native-maps: ee0bb36520eb49dfc2bf11754a486a4fa5c5f883
+  react-native-google-maps: 57af8d96a008489af2657caaeb074e7448451ff8
+  react-native-maps: 016cff46e0ae7ad4c1b1143a1aa2044e9983a544
   Yoga: 89c8738d42a0b46a113acb4e574336d61cba2985
 
-PODFILE CHECKSUM: c069397afd4c9ecf6886b7c00e7d7264dced43a4
+PODFILE CHECKSUM: d8c939699ce1f7d87205849a333b7e27fb1c4725
 
-COCOAPODS: 1.3.1
+COCOAPODS: 1.2.0
```
christopherdro: Is there a reason we're downgrading cocoapods?

example/ios/Podfile
```diff
@@ -7,7 +7,7 @@ platform :ios, '8.0'
 target 'AirMapsExplorer' do
   rn_path = '../../node_modules/react-native'
 
-  pod 'yoga', path: "#{rn_path}/ReactCommon/yoga/yoga.podspec"
+  pod 'Yoga', path: "#{rn_path}/ReactCommon/yoga/yoga.podspec"
```
christopherdro: This was recently changed to lower case.
https://github.com/react-community/react-native-maps/pull/1816

rborn: it won't build with uppercase

example/ios/Podfile.lock
rborn: I think that's what I have on my system, I can try to update cocoapods

alvelig:
@christopherdro Do you think we should create a separate example project?
My project was just a clone-n-go bare example.

christopherdro:
@alvelig I see. Let's just stick with the one we have.

example/ios/Podfile
christopherdro: I'm guessing this is because the example runs on an older version of react native.
Would be great to bump it up to the latest but will likely require a bit more work.

rborn: I'll try 😊

rborn:
@christopherdro @alvelig  what about now 🐽

christopherdro:
Awesome work!

tmaly1980:
Sorry, y'all, but this is totally broken. There's no example folder, only examples. And even when I symlink the two, I get this error:

[!] `/Users/tomas/sites/<project>/node_modules/react-native-maps/example/ios` is not a valid directory.

Issue #24, opened 2016-01-22
Animated.Region undefined constructor


KelseyRegan:
I may be missing something.  How is Animated.Region supposed to be referenced?  I can't seem to get it bundled at all.  It isn't one of the module.exports.  I see it in the examples, but can't use it in my project.

Any help is appreciated!

lelandrichardson:
See instructions here: https://github.com/lelandrichardson/react-native-maps#using-with-the-animated-api

You need to modify your AnimatedImplementation.js file. Let me know if you need any more help.

erickreutz:
@lelandrichardson  Is there a place with an example of modifying the `AnimatedImplementation.js`? This step doesn't seem to be in the README.

erickreutz:
Never mind. Found it!

agumack:
@erickreutz how fix it? thanks!

seasick:
@erickreutz, use `MapView.AnimatedRegion` instead of `Animated.Region`

PR #1889, opened 2017-12-14
Update for Animated.Region #24 Fix on the AnimatedMarkers Example 

timotew:
Fix for 'Animated.Region undefined constructor' in recent react-native version. 
Solved by using MapView.AnimatedRegion instead of Animated.Region


timotew:
@christopherdro Hi there! How should we proceed on this PR?

PR #1898, opened 2017-12-18
corrected default value to ``round``

JonasWho:
``lineJoin`` and ``lineCap`` should have default value ``round``, according to docs.

### DIFF
--- a/lib/components/MapPolyline.js
+++ b/lib/components/MapPolyline.js
@@ -137,6 +137,8 @@ const propTypes = {
 const defaultProps = {
   strokeColor: '#000',
   strokeWidth: 1,
+  lineJoin: 'round',
+  lineCap: 'round',
 };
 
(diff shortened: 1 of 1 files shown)


lib/components/MapPolyline.js
```diff
@@ -137,6 +137,8 @@ const propTypes = {
 const defaultProps = {
   strokeColor: '#000',
   strokeWidth: 1,
+  lineJoin: 'round',
```
rborn: In the docs there is no default for `lineJoin` - could you also update the docs?
tnnx

JonasWho: already did, I'm just committing my PR's through the web interface one at a time, sorry about that :)

rborn: Missed the #1899 PR, could you please make one PR only?  🤗

PR #1904, opened 2017-12-20
default export is now MapView

ivpusic:
With old ts file based on namespaces there was no default export, and therefore `import MapView from 'react-native-maps';` was not valid. I removed namespace, introduced module and exported all members.


rborn:
@ivpusic do you mind fixing conflicts and push again? Thanks.

ivpusic:
conflicts resolved

PR #1911, opened 2017-12-21
Add gradient/multi-color polyline support for iOS (MapKit)

IjzerenHein:
Hi guys, it's Christmas so it's time for presents  🎁
This PR adds support for gradient/multi-color Polylines for iOS/MapKit.

It adds a prop called `strokeColors` to `<Polyline>` which should be an array of colors, and which should have the same length as the `coordinates` prop. If two colors are different, it will then draw a gradient between those two coordinates. In order to create "smoother" gradients, you can also create "holes" in the color-array, be specifying `#00000000`. This will then draw a path using a linear gradient between the coordinates for which a color was defined.

MapExplorer
Arion

Example:

```js
<MapView>
	<MapView.Polyline
		coordinates={[
			{ latitude: 37.8025259, longitude: -122.4351431 },
			{ latitude: 37.7896386, longitude: -122.421646 },
			{ latitude: 37.7665248, longitude: -122.4161628 },
			{ latitude: 37.7734153, longitude: -122.4577787 },
			{ latitude: 37.7948605, longitude: -122.4596065 },
			{ latitude: 37.8025259, longitude: -122.4351431 }
		])
		strokeColor="#000" // fallback for when `strokeColors` is not supported by the map-provider
		strokeColors={[
			'#7F0000',
			'#00000000', // no color, creates a "long" gradient between the previous and next coordinate
			'#B24112',
			'#E5845C',
			'#238C23',
			'#7F0000'
		]}
		strokeWidth={6}
	/>
</MapView>
```

I've updated the docs and the MapExplorer as well.

Hopefully I can Android support for it too soon.

cheers,
Hein


alvelig:
LGTM. As far as I can see there is not any drawback in replacing MKPolygonRenderer with MKOverlayPathRenderer, is it?

IjzerenHein:
That’s right. And to be on the safe side, I enabled the new polyline renderer only when ‘strokeLines’ is used for now.

lib/ios/AirMaps/AIRMapPolylineRenderer.m
```diff
@@ -0,0 +1,252 @@
...
+//
+
+#import "AIRMapPolylineRenderer.h"
+
+@interface AIRMapPolylineRendererSegment : NSObject
+- (id)initWithPoint:(CGPoint)point color:(UIColor*)color;
+- (void) addPoint:(CGPoint)point color:(UIColor*)color;
+@property UIColor *color;
+@property CGMutablePathRef path;
+@property UIColor *color2;
```
rborn: can we find a better name for `color2` ?

IjzerenHein: Good catch. You're right, it would have been better to name these `startColor` and `endColor`, because this is exactly what they are. I'll make the changes.

IjzerenHein:
Please wait a bit with merging this, I'm seeing an occasional rendering issue with certain zoom scales

IjzerenHein:
Alright, rendering issues seem resolved now 👍

alvelig:
LGTM

PR #1917, opened 2017-12-22
Update issues template with no_template warning

rborn:
It seems that some people still don't use the template, so moved the warning on top and made it a little more aggressive :)

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,5 +1,22 @@
 <!--
   PLEASE DON'T DELETE THIS TEMPLATE UNTIL YOU HAVE READ THE FIRST SECTION.
+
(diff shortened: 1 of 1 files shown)


rborn:
@alvelig @christopherdro 🐽

christopherdro:
We could just make it a hard requirement and close all new issues that don't follow the template.
That Might be a bit too aggressive :)

rborn:
For a while I'd close it manually, with a small message. 
But I thought the template should have the warning there.

alvelig:
I don't feel like being much aggressive. The warning low priority without filled template nor reproducible example is fair enough.

alvelig:
https://github.com/probot/stale?

rborn:
@alvelig @christopherdro the stale bot looks good, I think we could give it a shot.

PR #1928, opened 2017-12-30
Update settings.gradle

njho:
Name change required in order to compile project

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -3,5 +3,5 @@ rootProject.name = 'react-native-maps'
 include ":example-android"
 project(":example-android").projectDir = file("./example/android/app")
 
-include ":react-native-maps-lib"
-project(":react-native-maps-lib").projectDir = file("./lib/android")
+include ":react-native-maps"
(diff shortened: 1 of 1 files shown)


rborn:
@njho thanks for the PR, could you please explain why we need this change?  What's the case when it doesn't compile?

alvelig:
Neither am I sure about why we need that.

rborn:
@njho please let us know why is this needed, what made you create this PR. Thank you.

rborn:
@christocracy any thoughts about this ? (I think you have way better knowledge of gradle than me 🤗 ) 🐽

christopherdro:
This settings.gradle is specific to the example project in the repo. With these changes, we would also need to following: https://github.com/react-native-community/react-native-maps/blob/00552eaa37b8b0fa459dba7c84aa5c92effa17c6/example/android/app/build.gradle#L130

We could clean things up with a bit with the example project and move the .gradle files out of the root directory.

PR #1950, opened 2018-01-08
Allow initialRegion to be set when first onLayout was called, and avoid flicker

danielgindi:
The boolean flag that was there before failed to allow initialRegion be set at all,
as GMS has a bug where the Camera created has a zoom level of 2 - and this shows the whole world.
The Camera is only created correctly when the mapview has layout.

But then if doing this from the JS side - it causes a flicker, it first shows the whole world and then zooms in. So I've replace the boolean flag with storage of the initial region to be set automatically when the view is added to the window.

### DIFF
--- a/lib/ios/AirGoogleMaps/AIRGoogleMap.m
+++ b/lib/ios/AirGoogleMaps/AIRGoogleMap.m
@@ -34,7 +34,7 @@ id regionAsJSON(MKCoordinateRegion region) {
 @implementation AIRGoogleMap
 {
   NSMutableArray<UIView *> *_reactSubviews;
-  BOOL _initialRegionSet;
+  MKCoordinateRegion _initialRegion;
 }
 
 - (instancetype)init
@@ -46,7 +46,7 @@ id regionAsJSON(MKCoordinateRegion region) {
     _polylines = [NSMutableArray array];
     _circles = [NSMutableArray array];
     _tiles = [NSMutableArray array];
-    _initialRegionSet = false;
+    _initialRegion = MKCoordinateRegionMake(CLLocationCoordinate2DMake(0.0, 0.0), MKCoordinateSpanMake(0.0, 0.0));
   }
   return self;
 }
@@ -145,8 +145,7 @@ id regionAsJSON(MKCoordinateRegion region) {
 #pragma clang diagnostic pop
 
 - (void)setInitialRegion:(MKCoordinateRegion)initialRegion {
-  if (_initialRegionSet) return;
-  _initialRegionSet = true;
+  _initialRegion = initialRegion;
   self.camera = [AIRGoogleMap makeGMSCameraPositionFromMap:self andMKCoordinateRegion:initialRegion];
 }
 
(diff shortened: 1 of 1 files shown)


rborn:
LGTM
@alvelig @christopherdro 🐽

danielgindi:
Tested locally, stress testing it with react-native-navigation going back and forth - seems good!
This solved the issue that appeared when upgrading to Google Maps 2.5.0

alvelig:
LGTM. I don't know if it's the genuine way to show initialRegion in GoogleMaps iOS, but seems to be good.

Agontuk:
@danielgindi, can you check this PR #1953 and see if it solves the flicker issue? About that PR, when `onLayout` is called, map won't be ready on native side. So `this.state.isReady` will be false and the block won't be executed. When map is finally ready,` _onMapReady` will be called which'll set the initialRegion or region. That way region will be set correctly. Hope I clearly explained this.

rborn:
@danielgindi @Agontuk hey guys, could you agree on these two PRs so maybe we can merge them and fix the initialRegion issue on iOS? 🤗

danielgindi:
@Agontuk Your PR does a good job handling camera events, which I didn't think about.
The problem is that the flickr still exists, as there will be an initial camera set, and `initialRegion` will be re-set from the JS side - which is async

danielgindi:
By the way, your new representation of "isReady" is now different from the Android version.
On the Android side, onMapReady means you have a MapView. Before that - you don't even have an instance to work with or to call functions on.
On the iOS side - you have a MapView from the beginning. It's only that some functions do not behave before being added to the window, but the mapview is there.
So I personally don't like the change to isReady, because the flag comes from the necessity in Android, and is actually always true for iOS.

Agontuk:
The reason I changed the `isReady` logic is that `setInitialRegion` gets called first, even before the `onMapReady` is fired from the native side. It's because we're calling `setNativeProps` from `onLayout` method which triggers before `onMapReady`. IMO that's the main issue with initialRegion not setting correctly. Avoiding setting props during onLayout event should fix this, but I don't what'll be the impact on apple map or on android.

I haven't faced any flickering issue on my end, but I'll do some tests just to be sure. One thing about setting initial region to 0 (on my end) is that when map first loads, I'm reverse geocoding to fetch the address. So it'll probably return an error I think, did not test it yet.

rborn:
@danielgindi @Agontuk hey guys, any advance on this one? 🤗

Agontuk:
@rborn, currently busy with project deadline so can't focus on this :(
These two PR conflicts with one another, so we've to come up with something that resolves all the issues. I'll try and see if flicker issue can be fixed with my changes.

danielgindi:
@rborn It's okay the initialRegion is called before onMapReady, that's why I put the guard in place.
The flicker does happen on a real device, it's just a matter of timing and caching...

If you don't mind, I'll merge your changes into my PR and make them work together, the best parts of both....

Agontuk:
@danielgindi, sure. I'll also take a look how can I fix this with my changes.

I'd like to point out that you removed the `_initialRegion` check in the `setInitialRegion` method. Now it'll be possible to set initialRegion multiple times, which I think contradicts how react works. Initial region should only be set once during initialization to provide a default region to work with.

In react, we've concepts of controlled & uncontrolled components. In controlled components, user keeps track of the value in a state & update the state when it changes. But if user doesn't want that, s/he can only pass a `defaultValue` instead, which is how uncontrolled components work. `defaultValue` is only set once, so even if user starts to change it, it'll not update the defaultValue. Hopefully I explained this properly, please let me know if my understanding is wrong.

Agontuk:
@danielgindi, I took a shot at fixing the flickering and pushed a new commit in the PR with a possible fix. Can you try this out and see if it actually solves the issue ?

danielgindi:
@Agontuk So far so good!
I'm closing this, @rborn you can go ahead with his PR

PR #1957, opened 2018-01-10
Fix Google map podspec

Titozzz:
For use in cocoapods only projects
(Requires cocoapod 1.4)


rborn:
@Titozzz thanks for the PR, could you please explain why we need this?

Titozzz:
Hello, sorry for the big delay, this ensure compatibility with cocoapods based project such as mine, feel free if you have any other question

rborn:
@Titozzz how is this influencing the "normal" installation ? I'm just afraid it would break the already fragile installation process.

PR #1960, opened 2018-01-12
Fixed gradient polyline not always fully drawn + stability issues

IjzerenHein:
The new gradient polyline was not always completely visible, because the Matrix transform in the paths was not applied correctly. This happened because the segments were cached and could have the wrong matrix transform applied to them (this happens implicitely during calls to CGPathMoveToPoint & CGPathAddLineToPoint). This could cause the line to not be complete visible, because it was drawn behind the map-tile. (see screenshots below)
Also, I’ve witnessed some rare crashes due to the caching of the segments. Apparently, the class was accessed from multiple threads which led to threading issues. Removing the segments cache fixed this problem as well.

**Before fix:**
image

**After fix:**
image

### DIFF
--- a/lib/ios/AirMaps/AIRMapPolylineRenderer.m
+++ b/lib/ios/AirMaps/AIRMapPolylineRenderer.m
@@ -39,7 +39,6 @@
 @end
 
 @implementation AIRMapPolylineRenderer {
-    NSMutableArray* _segments;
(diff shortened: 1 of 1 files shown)


rborn:
LGTM

@alvelig  🐽

alvelig:
Sorry, I missed this one. LGTM and thanks!

PR #1964, opened 2018-01-14
 Allow wider range of react and react-native versions

hawkrives:
I'm upgrading my project to react-native@0.52, I've got some peerDeps errors, so I thought I'd drop by and toss in a patch.

If you'd rather have `^0.51.0 || ^0.52.0`, I can do that too.

I did notice: The README states that the peerDep for react-native should be "*"…

> […] the peer dependency of this requirement is set to "react-native": "*" explicitly for this reason.

Would you like me to edit that paragraph as well?

The change from `>=` to `^0.51` seems to have happened in https://github.com/react-community/react-native-maps/pull/1879, fwiw.

Have a great week!

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,8 +33,8 @@
     "mapkit"
   ],
   "peerDependencies": {
-    "react": "16.0.0",
-    "react-native": "0.51.0",
+    "react": "^16.0.0",
+    "react-native": ">=0.51.0",
     "prop-types": "^15.5.10"
   },
   "devDependencies": {


rborn:
@hawkrives any issues with 0.52? especially with google provider?
Thanks a lot for the PR 🤗

hawkrives:
Sadly, I only use this on iOS with the default provider – does the example have a Google provider demo that I should try out?

rborn:
@hawkrives https://github.com/alvelig/react-native-maps-pods-example

hawkrives:
I'm going to mark this as unread and try to test that repository tonight…

rborn:
@hawkrives thank you so much for your PR, will close in favour of #2030 🤗

PR #1969, opened 2018-01-16
Added tracksInfoWindowChanges property to iOS Google Maps

satori-ytolstoguzov:
This property only affects iOS Google Maps (Android Google Maps simply don't need it). It allows to have an option to update callout content after callout was already displayed. At the same time it leaves this feature disabled by default that corresponds default behavior before.


rborn:
Travis please

alvelig:
LOL

Issue #1970, opened 2018-01-16
 "export 'default' (imported as 'MapView') was not found in './lib/components/MapView'


hedgepigdaniel:
### Is this a bug report?

Yes

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

Yes

### Environment

"react-native-maps": "0.19.0",
"react-native": "0.52.0",
"haul": "1.0.0-beta.12",
"webpack" : "3.8.1",
"babel-code": "6.26.0",
"babel-plugin-transform-class-properties": "6.24.1",
"babel-plugin-transform-flow-strip-types": "6.22.0",
"babel-plugin-transform-object-rest-spread": "6.26.0",
"babel-preset-env": "1.6.1",
"babel-preset-react": "6.24.1",

#### To start haul/webpack and compile the JS ####
1. Check out the test repository and switch to the `master` branch:
2. Run `yarn` to install dependencies
4. Run `yarn run haul` to run haul, which compiles the JS. Select Android as the platform.

#### To test the app on a device ####
1. Connect an android device via ADB
2. Run `yarn run run-android` to compile, install, and run the native app on the device
3. Shake the phone, go to "Dev Settings", and disable "Delta Bundles", then reload

#### To see the bug ####
1. Checkout the `master` branch. Notice that the app does not compile. This is the issue.
2. To explain my desire to reconfigure babel in a manner that causes this issue, checkout the `workaround` branch, where I patch `webhack.haul.js` to apply a different babel config just for react-native-maps so that I can avoid the issue. Notice that in the `yarn run haul` process which runs webpack, there is a big obvious warning about the fact that my code imports a named export which does not exist. In my experience this makes development much faster by catching mistakes earlier in development. In the `fix-import` branch there is a fix for the error in my app, which would otherwise have become an runtime error which may or may not have cause a crash or some other hard to trace, incorrect behaviour at some point during use.

The error I hope to see when I import an undefined export looks as follows (reproduce on the `workaround` branch):
```
 ERROR  Failed to compile.

./App.js
37:10-24 "export 'NONEXISTENT_EXPORT' (imported as 'IMPORTED_VALUE') was not found in './helper.js'
   at HarmonyImportSpecifierDependency._getErrors (/home/daniel/flattenedmonad/RNMExportBug/node_modules/webpack/lib/dependencies/HarmonyImportSpecifierDependency.js:65:15)
[...]

PR #1982, opened 2018-01-21
Replace CommonJS exports with ES6 exports

hedgepigdaniel:
This allows for the use of react-native-maps without transpilation of modules to CommonJS. That means that it can be imported in projects that don't do such transpilation, which has the advantage of detection of undefined imports, dead code elimination, and generally better static analysis. See #1970 for an example.

Also in the future, it allows for static dead code elimination of code in react-native-maps since it will not be necessary to keep references to components that are not used, e.g. `MapView.Marker`, unless the user explicitly imported that component with 
```
import { Marker } from 'react-native-maps';
```
All documentation and examples have been updated to use these imports but all the components are still attached to the base `MapView` component to avoid a breaking change.

I've checked all the examples - they work for me on Android. I can't see any reason why iOS would be different but I don't have an iOS device to check. I had to make one change to the Events example to prevent warnings relating to event pooling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,17 +26,17 @@ versions you should add `react` as a dependency in your `package.json`.
 
 [`<MapView />` Component API](docs/mapview.md)
 
-[`<MapView.Marker />` Component API](docs/marker.md)
+[`<Marker />` Component API](docs/marker.md)
 
-[`<MapView.Callout />` Component API](docs/callout.md)
+[`<Callout />` Component API](docs/callout.md)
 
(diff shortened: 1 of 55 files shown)


alvelig:
Ups, we have a little conflict here please :(

I will check it on the weekend on iOS (formally we need to do it) and merge then. Thank you very much.

hedgepigdaniel:
Merged, and the `AnimatedMarkers.js` example which was changed in master still works.

PR #2022, opened 2018-02-10
Support Map.Overlay for iOS

smellman:
Map.Overlay supports only Android so I did re-implement #1359 for iOS.
But I didn't write any code for Google Maps iOS because my simulator didn't run "Use GoogleMaps?" switch.


alvelig:
LGTM @rborn @christopherdro 🐽

g3r4n:
Thank you very much for the work !
I've test it with the default ImageOverlayWithURL example file and got the following output when zoming in : 
simulator screen shot - iphone 6 - 2018-02-11 at 15 05 48
Is it expected ?

smellman:
@g3r4n 

>  Is it expected ?

Maybe yes.
It seems `MKOverlay` actions.
When zoom up, image's coordinate will overlap so odd image will break.

I added offset in coordinates in example:

```
// 116423, 51614, 17
const OVERLAY_TOP_LEFT_COORDINATE2 = [35.679609609368576 + 0.000001, 139.76531982421875 + 0.000001];
const OVERLAY_BOTTOM_RIGHT_COORDINATE2 = [35.67737855391474 + 0.000001, 139.76806640625 + 0.000001];
const IMAGE_URL2 = 'https://maps.gsi.go.jp/xyz/std/17/116423/51614.png';
```

Results are following:





The image is break again but the breaking zoom level will increase by offset (0.000001).

rborn:
@g3r4n @smellman could it be something related to the way the overlay images are created? Don't they need to respect some strict rules?

smellman:
@rborn Yes, it is related. But in this case, making tile images using GeoServer or other programs is better. Also, react-native-maps supports tile images (UrlTile or LocalTile).

rborn:
@smellman so no fix for this?

smellman:
@rborn I change my example a bit.
Also, v0.20.1 enable to use GoogleMaps again, so I re-implement GoogleMaps API for iOS.

smellman:
I finished implement additional support for GoogleMaps and edit my example.
Please check it again.

g3r4n:
@smellman ok i understand the issue now 😄 I agree with you people should use UrlTile or LocalTile when image's coordinate will overlap.
Can you update the documentation to let people know about this limitation ?
Thank you for your hard work, i hope this PR will be merged soon !

smellman:
@g3r4n I pushed the documentation.
But I'm not native in English, if something wrong, please point out.

g3r4n:
@smellman Hmm i mange to reproduce the issue, but only on iOS device not in the simulator.
Before zooming : 
img_0023
After zooming : 
img_0024
Can you investigate ? Will try to investigate from my own

smellman:
@g3r4n I fix the bug, I forget to fix parameter order in MKMapRectMake.
It's completely my mistake.

g3r4n:
Works like a charm ! test are OK on iOS from my end.
cc @alvelig  @rborn @christopherdro

alvelig:
LGTM

aravi365:
still facing the issue on latest version

PR #2030, opened 2018-02-15
Broader peer dependency support

gabrielbull:
This will remove warnings like this when installing:

```
warning " > react-native-maps@0.20.1" has incorrect peer dependency "react@16.0.0".
warning " > react-native-maps@0.20.1" has incorrect peer dependency "react-native@0.51.0".
warning " > react-native-maps@0.20.1" has unmet peer dependency "prop-types@^15.5.10".
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,9 +32,9 @@
     "mapkit"
   ],
   "peerDependencies": {
-    "react": "16.0.0",
-    "react-native": "0.51.0",
-    "prop-types": "^15.5.10"
+    "react": "^16.0",
+    "react-native": "^0.51",
+    "prop-types": "^15.0 || ^16.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


rborn:
@gabrielbull thanks a lot for this, but could we use the correct semver syntax so `16.0.0` instead of `16.0` 🤗

gabrielbull:
It is the correct syntax: https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

But I don't have time to argue so I'm closing the issue.

rborn:
@gabrielbull Why so angry ? 😔 I was wrong and I accept this, so no arguing from my side.

gabrielbull:
Sorry, didn't mean to sound rude. I guess I'm just exhausted 😴. Thanks for being nice about this.

Issue #1489, opened 2017-07-18
Ask geo permission only when the map is mounted/rendered


mtt87:
RN 0.46.4, react-native-maps 0.15.3

As soon as I open the app it's asking for geo permission, is there a way to show it only when the component with the map is being mounted/rendered for the first time?

Thanks 😄

mtt87:
Any hint on this? 😄

kennyfox7:
I need this too :(

alvelig:
https://github.com/react-community/react-native-maps/issues/209#issuecomment-350907665

mtt87:
@alvelig why did you close this?
Your comment is about "showing the user location" which is different than the subject of this issue.

This issue is about asking the geo permission popup on request rather than automatically when the app is loadded and/or the map is mounted.

I'm talking about this, if it's not clear:
<img src="https://i.stack.imgur.com/XhAsU.png" />

sonaye:
Removing `showsUserLocation` prop did it for me, it asks for location permissions immediately it seems.

PR #2032, opened 2018-02-18
adding onMyLocationChange event

dabit1:
A fork from [https://github.com/react-community/react-native-maps/pull/1508](https://github.com/react-community/react-native-maps/pull/1508) with no conflicts. Thanks so much @sagi . He is the author of the pull request and the thanks must go to him but It has conflicts since July, 2017. I think it is a great PR and must be merged.
I also added more data to event: altitude, accuracy, altitude_accuracy and speed.

@alvelig Please could you merge this awesome PR?

Regards,
David Escalera


alvelig:
@dabit1 Thank you so much for such an important contribution! 
May I ask you to make a simple example on how to use it (or how you use it)?

rborn:
@dabit1 ping for docs/example 🤗

dabit1:
😂 I'm so busy sorry. Maximum Sunday I hope do that.

bramus:
Oh, this one would be awesome! Nice work!

dabit1:
@alvelig there is an example file changed with this new feature on this PR. Is it ok?

Bosseskompis:
Any update on this?

alvelig:
@dabit1 LGTM sorry for the delay, please resolve conflicts and we merge.

alvelig:
And also do you ming renaming `onMyLocationChange` to `onUserLocationChange` as it will be more consistent with `showsUserLocation`

rborn:
+1 for renaming 🤗

dabit1:
Ey guys tomorrow afternoon I will do it. 😎

dabit1:
@alvelig done!

zerazeru:
There's a `}` missing in AIRGoogleMap.m. PR's get merged without even trying to build before..? :/

Anyway, after fixing the event shows up, but there's no data (on ios)

bramus:
Just tested out this PR using the example provided (up to the point before the rename) in the Android Emulator.

I had to disable the `coordinate.putFloat()` calls because it would not compile otherwise. Once I got it running it's not not working:
1. when playing back a `.gpx` file via the Emulator
2. when manually sending a new location to the Emulator

In both cases I don't see any location change events :(

Is it possible to revert this PR merge, awaiting a proper/better implementation? It'd also unbreak the master (see https://github.com/react-community/react-native-maps/issues/2058) along with that.

bramus:
With the hotfix above applied I've done some more testing in the Simulators:
- iOS: builds + works
- Android: builds + doesn't work (neither when playing back a `.gpx` file via the Emulator nor when sending over new location to the Emulator)

zerazeru:
On iOS it builds for me but there is no event data (no coordinates), only a Proxy object. Does it work for you?

I noticed that it's the same for onPress, but for example for onRegionChange the proper event object is received.
Could it be related to RCTBubblingEventBlock / RCTDirectEventBlock ?

rborn:
@bramus on android same issues with you no location is triggered, not even on android side - I've put a `Log.v` in the event  (tested on a genymotion). Waiting for my android device to charge and  I'll test for real too.

alvelig:
I've tested in the Emulator with example. Does it work for you, guys?

rborn:
@alvelig as I said, no

bramus:
Same here @rborn (added some log statements in the handling Java code … no result).

On iOS I see the events come in @zerazeru (Tested with Debug > Location > Bicycle Ride):

screen shot 2018-03-06 at 13 15 46

rborn:
@bramus  @alvelig  Ok, figured out, the example app doesn't ask for permissions on start 😳
You need to go and enable location manually in the Settings then it will work

bramus:
Weird, as [the `AndroidManifest.xml` file](https://github.com/react-community/react-native-maps/blob/master/example/android/app/src/main/AndroidManifest.xml) does require 'm:

```
    <uses-permission android:name="android.permission.ACCESS_COURSE_LOCATION"/>
    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
```

🤔

rborn:
@bramus @alvelig this? https://stackoverflow.com/a/32636039/8018651

bramus:
Aha, that could be it indeed … should we update the code to include something like https://github.com/yonahforst/react-native-permissions to request for location permissions first, in order to prevent confusion like we're having now? _(I'm using this library in an already deployed app, works fine :))_

rborn:
I like that library too but I'm not sure if we should suggest using it in the docs or try to do it inside this module so we don't force the user to use an extra module.

On the other hand that library gives way more flexibility and would allow a better UX

🤔

alvelig:
Consider #1489

bramus:
I wouldn't include `react-native-permissions` in the core itself, as it would only broaden the surface `react-native-maps` is covering. Including in the examples _(to make sure they work)_ and mentioning it in the documentation _(to not confuse those checking out the examples, like I was now)_ should do the job.

For an app I built I created this “Permissions Required” component which uses `react-native-permissions` under the hood. The text is in Dutch, but you get the idea: it asks for permission to use your location, and only after having granted the required permissions the `<MapView />` is being rendered.

screenshot

With some adjustments this could make it into the examples.

rborn:
@bramus I totally agree with adding some docs about this

aMarCruz:
@dabit1 please add typings for all of us using Typescript :)

```ts
onUserLocationChange?: (value: {
    coordinate: {
        latitude: number,
        longitude: number,
        altitude: number,
        accuracy: number,
        speed: number,
        isFromMockProvider: boolean
    }
}) => void;
```

dabit1:
@aMarCruz feel free to make a PR ;)

dabit1:
@aMarCruz well done!

AakashKB:
For some reason, when I use onUserLocationChange, I get an undefined latitude and longitude.

My code:
```
<MapView
          ref={map => (this._mapView = map)}
          style={styles.map}
          showsUserLocation={true}
          showsPointsOfInterest={false}
          followsUserLocation={false}
          provider="google"
          region={this.state.region}
          onRegionChange={region => this.state.region = region}
          onUserLocationChange={location => alert(location.latitude)}
          onLongPress={event =>
            this.createMapEvent(event.nativeEvent.coordinate)
          }
        >
```

dabit1:
@AakashKB are permissions enabled?

AakashKB:
@dabit1 yes, location permissions are enabled. showUserLocation works well.

AakashKB:
onuserlocationchange

I get this when I print out Object.keys(location). There is no key for latitude or longitude even returned.

PR #2035, opened 2018-02-20
fixed some typescript issues

AlmirKadric:
Fixed a few typescript issues I found when using the package within our project. Also exposed Animated versions of the components so that ES6 import statements can be used on them too.


rborn:
@AlmirKadric LGTM but there are conflicts due to the last merge I think. Could you please fix them 🤗

AlmirKadric:
@rborn sorry i didn't notice the lint error, should all be fine now

PR #2049, opened 2018-03-02
Added animateToNavigation method to MapView

GuilhermeEd:
animateToNavigation is a method to allow navigation-like moviment to a map that, for instance, follows the user position. (like google maps navigation or waze).

On android when trying to animateToRegion, animateToBearing and animateToViewingAngle simultaniously the methods would overwride each other's movement. animateToNavigation solves this problem.

CreateMap method is introduced to solve the react's MapBuilder limitation of key-value pairs.

This was already tested worked on android.


alvelig:
Great contribution! But could you add an example to the project please?

alvelig:
Please undo c819f45 and add an example [here](https://github.com/react-community/react-native-maps/blob/master/example/App.js) please.

GuilhermeEd:
Surely! [c819f45](https://github.com/react-community/react-native-maps/commit/c819f45e7468c3fe3a9b0990a9b49a4d067d2e50) undone and added example [here](https://github.com/GuilhermeEd/react-native-maps/commit/1c15fc3fdc0b4ff0762d4e8d3b742b05bdba7e42).

juergengunz:
@GuilhermeEd thanks for this contribution, I tested it on android - it works, but it behaves weird if you set mapPadding. If you want the marker a little bit closer to the bottom, like navigation apps do, for example.

alvelig:
LGTM @rborn @christopherdro 🐽

GuilhermeEd:
@juergengunz What a great idea! I'll see what i can do.

GuilhermeEd:
Hey guys. I'm having trouble calculating the offset at [AirMapView.java](https://github.com/GuilhermeEd/react-native-maps/blob/e759876c696806e7f97022b275cf346acf14ab2b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java), lines 590-605. If you could take a look at my algorithm that would be great. I commented it so you could check it out.

juergengunz:
@GuilhermeEd nice! Check out this comment, he made an example for iOS and this seems to work. Maybe you could just adapt the algorithm for android.
https://github.com/react-community/react-native-maps/issues/1726#issuecomment-347593978

EricPKerr:
@GuilhermeEd what is the `offset` being used for?

Wouldn't it be better to have this alignment stuff be relative to the map zoom (or lat/long delta)?  Ideally this just looks at what the coordinates of the edge of the screen are if you want to align the user's center point to the bottom of the screen.  Better to ship this without the offset logic for now ?

GuilhermeEd:
@EricPKerr The offset should be the distance from the current location to the bottom of the map. For instance, most of the navigation apps out there shows the driver position not at the center of the map, but a little to the bottom, instead.
Your idea is good, tough.
Unfortunely lately i've been overwhelmed with work. Maybe ship it without the offset logic would do the trick for now.

equesteo:
I've been watching this for the past month and a half, super excited for it to land. I think the offset is not super necessary, I'd mostly just love to be able to set the angle and bearing at the same time. Thanks for your work, hope I get to use it soon!

GuilhermeEd:
I have already used this functionality in one of my projects and the offset wasn't much of a must-have in this case because most of the time i needed to display some info at the bottom of the screen (which overlapped the bottom part of the map).
Back to the business, how can i submit it without the offset funcionality for now?
Pull > Resolve Conflicts > Pass Travis Build > Pull Request?

rborn:
@GuilhermeEd yeah 🤗

GuilhermeEd:
There you go :)

lib/android/src/main/java/com/airbnb/android/react/maps/AirMapManager.java
```diff
@@ -332,14 +345,15 @@ public Map getExportedCustomDirectEventTypeConstants() {
   @Nullable
   @Override
   public Map<String, Integer> getCommandsMap() {
-    Map<String, Integer> map = MapBuilder.of(
+    Map<String, Integer> map = this.CreateMap(
```
rborn: Why this? `CreateMap` looks to me now very unreadable and limiting (only 8 key/values pairs allowed and ambiguous naming ->  `k1,v1`)

rborn:
@GuilhermeEd added few comments 🤗

equesteo:
Gonna be so excited when this lands.

GuilhermeEd:
@rborn I've deletes the IDE and .class files as you asked. Do you think there is something else that needs to be done? If you have a good solution for the Map function that we've discussed above that would be great as well.

equesteo:
@rborn Is there anything I can do to get this through?

rborn:
@alvelig need some help here, can you take a look? 🐽

alvelig:
@rborn I am! Thanks for the reminder! LGTM
@equesteo @brunocascio have you tried the branch? Can you confirm that it's working ok for you? We'd be very grateful if you helped us with testing this PR.

equesteo:
@alvelig Tested and works for me.

jonathanmauer:
Can i pass a object with all my pins locations to it? like in fitToCoordinates(objectWithLocations) ?

ayushnawani:
How to focus on the current location while using animateToNavigation? In the example also map position is fixed. Can anyone help?

techieyann:
I've been waiting on this functionality, I even wrote my own version for Android but didn't get around to the iOS code. Thank you so much!

Why is there no zoom parameter though? I know most of the package uses regions which negates the need for zoom, but this function just uses a center. It's just one number passed down to the GoogleMap API. 

@ayushnawani It's up to you to supply the lat/long of where you want the map to be. It sounds like you need to access the GPS of your device and then pass the location results to the function.

PR #2050, opened 2018-03-02
add Android support for onPoiClick

julienR2:
Exposes the onPoiClick event to react-native-maps wrapper.
It send back the location of the poi, the name and the googleId (docs updated)

I haven't dug in IOS implementation yet since I'm more confortable with android. But I'll definitely do it through a next PR ! (or if anyone is keen to, feel free)

Let me know if any other steps need to be done 👍

### DIFF
--- a/docs/mapview.md
+++ b/docs/mapview.md
@@ -50,6 +50,7 @@ To access event data, you will need to use `e.nativeEvent`. For example, `onPres
 | `onRegionChangeComplete` | `Region` | Callback that is called once when the region changes, such as when the user is done moving the map.
 | `onPress` | `{ coordinate: LatLng, position: Point }` | Callback that is called when user taps on the map.
 | `onPanDrag` | `{ coordinate: LatLng, position: Point }` | Callback that is called when user presses and drags the map. **NOTE**: for iOS `scrollEnabled` should be set to false to trigger the event
+| `onPoiClick` | `{ coordinate: LatLng, position: Point, placeId: string, name: string }` | Callback that is called when user click on a POI. **NOTE**: Android only (IOS soon)
 | `onLongPress` | `{ coordinate: LatLng, position: Point }` | Callback that is called when user makes a "long press" somewhere on the map.
 | `onMarkerPress` |  | Callback that is called when a marker on the map is tapped by the user.
 | `onMarkerSelect` |  | Callback that is called when a marker on the map becomes selected. This will be called when the callout for that marker is about to be shown. **Note**: iOS only.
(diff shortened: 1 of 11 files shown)


alvelig:
Great! Can you please also add an example of usage. And we'll have to rebase a little bit later it as there's another PR adding feature to the same files. So if you are going to add iOS implementation, absolutely welcome (will it be only GM specific?)

julienR2:
A pleasure to collaborate !
I'll definitely add an example of usage today 👍
And I'll take advantage of the time before the other PR is added, to dig into the IOS implementation. I quite new in the SDK-to-React Native behavior so I'll see what I can do ! Idealy I would make it non GM specific but I'll check if it's possible depending on the other maps lib !

julienR2:
Here is the example ! Let me know if it's good.

There is on thing I still feel weird about, it to have to get the `nativeEvent` of the `event` to get the object define in Android (https://github.com/nowmad-io/react-native-maps/blob/ff8e45bd9c2645a9baf6611da875ebc33c672543/example/examples/OnPoiClick.js#L37)

I'm open to update it if there is a better way, that I didn't get yet.

rborn:
LGTM @alvelig 🐽

julienR2:
@rborn @alvelig 
I added the support for the onPoiClick event in IOS.
It seems MapKit doesn't expose any event on their POI, so this event is GM specific (I updated the README to reflect the information).

Hope it's all good, let me know otherwise !

rborn:
@julienR2 LGTM, one question only: what's happening if we use the apple map? 
Are there any drawbacks?

julienR2:
@rborn Since the callback doesn't exist on MapKit, I implemented it only on AirGoogleMaps. Therefore if you try using onPoiClick callback on with native IOS map, nothing will happen since no listener nor callback are set.

I'm wondering if it would be worth logging a warning `AirMap doesn't provide onPoiClick event` in case we try to set the callback on an IOS env, but I don't recall seeing this kind of practice in the rest of the lib.

Let me know if it's needed !

rborn:
Thats great, and you specify in the docs it's google only so no need for the log.

docs/mapview.md
```diff
@@ -50,6 +50,7 @@ To access event data, you will need to use `e.nativeEvent`. For example, `onPres
 | `onRegionChangeComplete` | `Region` | Callback that is called once when the region changes, such as when the user is done moving the map.
 | `onPress` | `{ coordinate: LatLng, position: Point }` | Callback that is called when user taps on the map.
 | `onPanDrag` | `{ coordinate: LatLng, position: Point }` | Callback that is called when user presses and drags the map. **NOTE**: for iOS `scrollEnabled` should be set to false to trigger the event
+| `onPoiClick` | `{ coordinate: LatLng, position: Point, placeId: string, name: string }` | Callback that is called when user click on a POI. **NOTE**: Android only (IOS soon)
```
rborn: We have iOS now, right? So the `soon` line can be removed

example/examples/OnPoiClick.js
```diff
@@ -0,0 +1,86 @@
...
+      region: {
+        latitude: LATITUDE,
+        longitude: LONGITUDE,
+        latitudeDelta: LATITUDE_DELTA,
+        longitudeDelta: LONGITUDE_DELTA,
+      },
+      poi: null,
+    };
+
+    this.onPoiClick = this.onPoiClick.bind(this);
```
rborn: If you use an arrow function the bind is not needed, but it's up to you :)

rborn:
@julienR2 small comments, then we merge 🎉

julienR2:
Oups well seen 👍  I forgot to change this after implementing IOS. I'll do it immediately !

julienR2:
Awesome ! Thanks for merging it 👍

PR #2057, opened 2018-03-05
Fixed polygon and polyline not re-render when changing tile URL

LeMinh1995:


LeMinh1995:
Hi. Currently, I'm working with UrlTile component. But I found an issue when changing tile URL dynamically. The polygons and polylines are disappeared.

lib/ios/AirMaps/AIRMapUrlTile.m
```diff
@@ -39,6 +39,13 @@ - (void) update
     if (_map == nil) return;
     [_map removeOverlay:self];
     [_map addOverlay:self level:MKOverlayLevelAboveLabels];
+    for (id<MKOverlay> overlay in _map.overlays) {
```
rborn: Won't this have performance issues. Isn't there any way to decide the "zIndex" of an overlay so we don't have to remove and add the other overlays?

LeMinh1995: I have tried insertOverlay with index equal to 0 but still not working.

LeMinh1995: I have tried to use insertOverlay with atIndex equals 0. But still not working. So this is the solution I came up so far.

rborn: I had a look and it seems that it's done the same way for the polylines and the rest of the views. Thank you.

rborn:
LGTM @alvelig 🐽

rborn:
@alvelig 🐽🐽 - aka double 🐽  😹

alvelig:
@rborn thank you for pushing us forward. LGTM. At least I don't have another alternative to come up with.

PR #2068, opened 2018-03-07
Stop marker event propagation in order to prevent onPress for MapView…

hysan:
Stop marker event propagation in order to prevent onPress for MapViews from being called when a marker is clicked. This makes the behavior of Apple Maps identical to that of the behavior of Google Maps on Android. This fixes react-community/react-native-maps#1132.

### Is there any other PR opened that does the same ?

No, not that I've seen. Nothing is referenced in any of the issues I've come across either.

### What issue is this PR fixing

https://github.com/react-community/react-native-maps/issues/1132


rborn:
LGTM @alvelig 🐽

alvelig:
@hysan you yourself say that it would be a switchable adjustment with a prop... Let's be consistent for this behavior and I suggest keeping propagation by default but make it configurable with a prop `stopPropagation={true}`.

hysan:
@alvelig Ok, so add a `stopPropagation` prop and default it to false + update the `Marker` docs? How much explanation should I put into the docs?

alvelig:
As much as you wish. You may just reference your comment in the issue with short comments. That will do.

hysan:
Done, but I'm not familiar with TypeScript so all I did was copy paste. Also, I could have not defined defaultProps and just relied on undefined being falsy, but I wasn't sure if that was idiomatic or not. Someone should double check. Thanks.

alvelig:
LGTM

PR #2079, opened 2018-03-10
Update installation.md

aperhaite:
### Is there any other PR opened that does the same ?

(please answer here)

### What issue is this PR fixing

(please link the issue here)


rborn:
@aperhaite thanks for the PR, would be good if you could specify that the sections need to be enabled **for the current project** (in my case I have multiple projects in Google Console, and the link opened in one that's not relate to the current one) 🤗

PR #2131, opened 2018-03-27
Fix initialRegion when using RN 0.54

lachenmayer:
See https://github.com/react-community/react-native-maps/issues/2100 for discussion.


rborn:
@lachenmayer how does this influence the lower versions of RN ?

Agontuk:
AFAIK, this'll work fine in previous versions too. I tested it with RN 0.52 and 0.54, both worked fine.

PR #2132, opened 2018-03-27
Add Mock Provider boolean on each location update

deorukhkarchaitanya:
No

This adds a check to whether the returned location is authentic or mocked

### How did you test this PR?
Tested it on my application where I am checking for mock providers

1. Set showsUserLocation={true} in MapView component
2. Set 'google' as your map provider
3. Listen for onUserLocationChange which will return a callback with lat,long,speed,accuracy & boolean for isFromMockProvider

(Works only for Android)
I have tested this on Android (real device)


rborn:
@deorukhkarchaitanya thanks for the PR 🤗
If I understand this right `isFromMockProvider` is available only on android.

So please update the docs as well, specifying also the above (if correct of course)

deorukhkarchaitanya:
@rborn Sorry for that. Let me just run a check and update it.

rborn:
@deorukhkarchaitanya I mean the real docs page, not only the PR text https://github.com/react-community/react-native-maps/blob/master/docs/mapview.md 😄

deorukhkarchaitanya:
Updated the docs. Thanks again for the guidance :)

PR #2136, opened 2018-03-29
Modifications/Enhancements to MapView.UrlTile

akmjenkins:
1. Added propType shouldReplaceMapContent (effectively maps to canReplaceMapContent on MKTileOverlay) so that it's no longer hard coded to YES (is that a good default?) on iOS, MapKit.

2. Let GoogleMaps obey maximumZ (no longer just for iOS, MapKit!). 

3. While we're at it, add a minimumZ for all maps on both platforms.

Not that I'm aware

N/A

### How did you test this PR?

Tested on both simulators on a Mac


rborn:
LGTM @alvelig 🐽

alvelig:
LGTM

Issue #2061, opened 2018-03-05
coordinateForPoint is misnamed in lib/components/MapView.js as coordinateFromPoint so if you try to call it you get undefined


hlz2103:
### Is this a bug report?

yes

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

(Write your answer here.)
yes
### Steps to Reproduce

if you try to call this.map.coordinateForPoint, you get undefined. This is because the function is called coordinateFromPoint
(Write your steps here:)

1. 2. 3.

### Expected Behavior

(Write what you thought would happen.)

### Actual Behavior

(Write what happened. Add screenshots!)

### Reproducible Demo

(Paste the link to an example project and exact instructions to reproduce the issue.)

rborn:
@hlz2103 could you please try this PR? https://github.com/react-community/react-native-maps/pull/2039 and report back if it works? Thank you.

PR #2039, opened 2018-02-22
Fix for pointForCoordinate and coordinateForPoint

wasedaigo:
* Fixed typo "coordinateFromPoint"
* Moved coordinateForPoint/pointForCoordinate from AirMap to AirMapManager
* Fixed how resolve/reject are handled

I have tested on iOS. Yet not confirmed on Android. If anybody can confirm this works on Android, I think this PR is good to go.

PR #2150, opened 2018-04-01
Fixed `pointForCoordinate`/`coordinateForPoint` Promises in iOS (#2039, #2061)

danielgindi:

No

Changes to these functions that did not reflect in both providers. Currently the API is broken - expecting callback in GoogleMaps and returning Promise in Air. While the docs says it returns Promise.

### How did you test this PR?

Written simple code with `await` to print results of both functions, in Air and Google provider.

### DIFF
--- a/lib/ios/AirGoogleMaps/AIRGoogleMapManager.m
+++ b/lib/ios/AirGoogleMaps/AIRGoogleMapManager.m
@@ -298,7 +298,8 @@ RCT_EXPORT_METHOD(takeSnapshot:(nonnull NSNumber *)reactTag
 
 RCT_EXPORT_METHOD(pointForCoordinate:(nonnull NSNumber *)reactTag
                   coordinate:(NSDictionary *)coordinate
-                  withCallback:(RCTResponseSenderBlock)callback)
+                  resolver: (RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject)
 {
   CLLocationCoordinate2D coord =
   CLLocationCoordinate2DMake(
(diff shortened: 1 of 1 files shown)


rborn:
@alvelig 🐽

lachenmayer:
Hey there, I was actually looking at this same issue too before the holidays.
I spent a considerable amount of effort trying to get an integration test running using `RCTTest`.
The code for that is here: https://github.com/react-community/react-native-maps/compare/master...lachenmayer:pointforcoordinate-bug
The main difference is a new `AirMapsExplorerTest` target which uses `RCTTestRunner` to run tests. These can be run in Xcode by pressing ⌘-U in the workspace (while the packager is running).

Currently, the only test is for exactly this issue: https://github.com/lachenmayer/react-native-maps/blob/pointforcoordinate-bug/example/tests/PointForCoordinate.js
I'm going to try apply this patch & see if it passes (it should!).

In general, I was wondering if there is any interest in adding these kinds of tests - I feel like we could avoid a lot of these kinds of issues with a test suite which works very similarly to how React Native itself is tested.
If you agree, I'll make a PR with a cleaned up version of this & add some instructions on how to add new test cases.
Cheers :)

alvelig:
LGTM

@lachenmayer I was thinking about tests too, but did not know the best way. Let's try to add example tests for this, so we can ask contributors to make tests for their PRs.

lachenmayer:
@alvelig please check out https://github.com/react-community/react-native-maps/pull/2161 for a test for this PR :)

davija:
Any updates on this?  It has been almost a month since there was any traffic on this and this is currently a blocking issue for me.

PR #2165, opened 2018-04-04
Typings for v0.21.0

aMarCruz:

I think so, but this PR is updated and more refined.

Mainly, event interface and missing methods and properties.

### How did you test this PR?

Tested with VS Code in real work, but not all the typings (We are not using all).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,25 +1,63 @@
 declare module "react-native-maps" {
+
     import * as React from 'react';
-    import { Animated } from 'react-native';
+    import {
+        Animated,
+        ImageRequireSource,
+        ImageURISource,
+        NativeSyntheticEvent,
+        ViewProperties
+    } from 'react-native';
 
     export interface Region {
-        latitude: number
-        longitude: number
-        latitudeDelta: number
-        longitudeDelta: number
+        latitude: number;
+        longitude: number;
+        latitudeDelta: number;
+        longitudeDelta: number;
     }
 
     export interface LatLng {
-        latitude: number
-        longitude: number
+        latitude: number;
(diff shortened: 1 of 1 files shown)


rborn:
@aMarCruz could you please let me know what other PRs do the same but are outdated so I can close them? 🤗

firefueled:
Hey @aMarCruz I'm wondering why you wrote the type definition for MapEvent the way you did
```typescript    
export interface MapEvent<T = {}> extends NativeSyntheticEvent<T & {
    coordinate: LatLng;
    position: Point;
}> { }
```
, instead of
```typescript
export interface MapEvent<T = {}> extends NativeSyntheticEvent<T> {
    coordinate: LatLng;
    position: Point;
}
```

The way it is now, the `coordinate` and `position` end up hidden under the `nativeEvent` key. Shouldn't they be added to the parent object so that one could access them more directly?
`e => e.coordinate.latitude`

Am I missing something?

aMarCruz:
@firefueled I really don't know. May be a typo. Right now I'm not using maps but will check it next week.

firefueled:
I was thinking about submitting a PR like this:
```typescript
type MapEventType = {
    coordinate: LatLng;
    position: Point;
}
export interface MapEvent<T = {}> extends NativeSyntheticEvent<T & MapEventType>, MapEventType { }
```
but I wasn't sure if the thing you did was on purpose so I decided to ask first.
The redundancy there is to make sure people's builds don't break.
Also, we need to check that those values really are NOT expected to be included into `NativeSyntheticEvent`

aMarCruz:
yeah, it needs testing.

PR #2181, opened 2018-04-11
Add note about NSLocationWhenInUseUsageDescription

sfyfe31:
Apple are now rejecting app submissions without an explanation of why you need to user's location.

No

Explaining a required step for location tracking, which will prevent Apple rejecting an app submission.

### How did you test this PR?

It is a readme note. No code changes.


rborn:
Perfect, thank you 🤗

PR #2196, opened 2018-04-14
Upgrade old dependency on babel-preset-react-native 2.x.x

hedgepigdaniel:

No

Peer dependency warnings when using an up to date version of babel-preset-react-native

### How did you test this PR?

I tried to compile the test app, but I was stopped by this issue: https://github.com/react-community/react-native-maps/issues/2188

However, it would be very surprising if this broke anything, since most projects currently using react-native-maps (and compiling it with their own babel setup) are probably already using babel-preset-react-native 4.x.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,6 @@
   },
   "dependencies": {
     "babel-plugin-module-resolver": "^2.3.0",
-    "babel-preset-react-native": "1.9.0"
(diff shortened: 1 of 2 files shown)


rborn:
Closing, old PR and the package is deprecated https://www.npmjs.com/package/babel-preset-react-native

Thanks for all the effort 😻

Issue #2027, opened 2018-02-14
Big memory and cpu leak on android


v3solutions:
### Is this a bug report?

Memory and cpu leak on android using any navigation (react native navigation, react navigation, even simple if switching pages). Seems like a very serious issue. It increases until it just crashes (in my case ~400mb and cpu 7% on galaxy s5)

### Environment
Android
react-native: 0.53.0
react-native-maps: 0.20.0

### Steps to Reproduce

(Write your steps here:)

1. Any two components, one or both containing MapView. While testing i was using just plain MapView, no markers, just <MapView></Mapview> with basic options like initialRegion and zoom.
2. Switching between them will add to memory and cpu significantly. Using most popular navigation tools or simple ifs statements to switch components.

### Expected Behavior

Memory and cpu should be somewhat in the same range, now it just simply adding up. Possibly something to do with lifecycle methods.

### Actual Behavior

Every single time you switch to MapView it will add cpu and memory until it crashes.

### Reproducible Demo

If there's a need i can make a demo project, but this is just basic stuff, 2 components with one of them having a mapview and simply switching between them.

alvelig:
That's an interesting case. Any ideas how to localize the leak?

ltchurch88t:
A potential avenue to explore would be nesting your MapView inside a screen accessible through the DrawerNavigator for android. This forces you to re-render the MapView when you return to the screen holding that view. I will give it a shot here in a little bit and post some code for you if I come up with a solution.

v3solutions:
I'd be happy to sponsor anyone who can come up with a solution this week. It's urgent :(
Btw. it would be great to know how can i compile my own version of react-native-maps after editing library. I can't find any docs regarding this.

khsily:
having same problem on ios :(

DanGDroid:
@v3solutions 
hi,
to edit a RN lib and use it you have to add it to your project file and use your edited version, but it wont be updated for npm any more.
to add a lib locally:
in your project directory create a `Lib` folder, and cut and paste the module you want from `node-modules` folder.
do your changes - it is now in your local project.
in `package.json` change the import to:
`"react-native-maps": "file:./App/Lib/react-native-maps",`

v3solutions:
@DanGDroid 
Thanks, I'll take a look.

wjielim:
@v3solutions Which navigator did you used for that two components containing the MapView?
[...]

PR #2197, opened 2018-04-14
Fix overlay removal bug

kabbi:

Nope

When `<Overlay/>` component is unmounted before the image had a chance to load (either a large image, or slow internet connection), after image finishes loading overlay is still rendered, and cannot be removed afterwards.

### How did you test this PR?

There was a bug on Android which didn't cancel image rendering after removing from map. My change erases map reference, thus avoiding this bug.

Tested on several Android devices, does not seem to affect iOS.


rborn:
LGTM 
@alvelig 🐽

alvelig:
Could #2027 relate to it?

kabbi:
@alvelig I don't really think so. This has a very specific and narrow use-case, and the issue you've linked does not ever use any `Overlay`s, so it should not be related.

PR #2207, opened 2018-04-18
Fix https://github.com/react-community/react-native-maps/issues/2203

sytolk:

no

https://github.com/react-community/react-native-maps/issues/2203

### How did you test this PR?

every location is have timestamp and it's now received on RN side.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -108,6 +108,7 @@ declare module "react-native-maps" {
                 latitude: number,
                 longitude: number,
                 altitude: number,
+                timestamp: number,
                 accuracy: number,
                 speed: number,
                 isFromMockProvider: boolean,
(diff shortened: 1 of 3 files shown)


rborn:
Please add the documentation too with platform specific  🤗
Can this be extended to iOS too? AppleMap and GoogleMap ?

Thank you

sytolk:
@rborn I think that Documentation for onUserLocationChange exist: 
https://github.com/react-community/react-native-maps/blob/master/docs/mapview.md#events

 *Callback that is called when the underlying map figures our users current location (coordinate also includes isFromMockProvider value for Android API 18 and above). Make sure showsUserLocation is set to true and that the provider is "google"*

this is only minor change that add timestamp parameter to location object received on RN side and I don't think that that there is need documentation changes

rborn:
@sytolk what about iOS ?

sytolk:
sorry but I cannot test onUserLocationChange on iOS. But from documentation if the map provider is google and showsUserLocation is set to true should work.

rborn:
LGTM @alvelig 🐽

PR #2217, opened 2018-04-26
Fix color prop types

pedro:

No

Using react-native's `ColorPropType` to validate all color props more accurately.

More specifically I'm doing this to avoid getting yellow box alerts when using different formats for colors. For instance, using `setNormalizedColorAlpha()` I have colors defined as integers in my app.

### How did you test this PR?

Just tested against my app, both in iOS and Android. Colors still work/no more alerts.

### DIFF
--- a/lib/components/MapCircle.js
+++ b/lib/components/MapCircle.js
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import {
+  ColorPropType,
   ViewPropTypes,
   View,
 } from 'react-native';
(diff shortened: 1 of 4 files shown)


pedro:
Ah, worth noticing that [`ColorPropType` is already used in `MapView.js`](https://github.com/react-community/react-native-maps/blob/71c74cdd8f5f30efd64bd20f2790db65d959b328/lib/components/MapView.js#L179), so I don't think there are any concerns about compatibility with this change.

rborn:
LGTM @alvelig 🐽

Issue #2215, opened 2018-04-24
Compile error in iOS example project: Illegal type 'NSDate *' used in a boxed expression


cmbasnett:
### Is this a bug report?
Yes.

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?
Yes.

### Environment
```
{ 'react-native-maps': '0.21.0',
  npm: '5.7.1',
  ares: '1.10.1-DEV',
  cldr: '31.0.1',
  http_parser: '2.7.0',
  icu: '59.1',
  modules: '57',
  nghttp2: '1.25.0',
  node: '8.9.4',
  openssl: '1.0.2n',
  tz: '2017b',
  unicode: '9.0',
  uv: '1.15.0',
  v8: '6.1.534.50',
  zlib: '1.2.11' }
```
Podfile.lock is [here](https://gist.github.com/cmbasnett/d288cf550c09eb25aeab6bcb327438c9).

### Steps to Reproduce
1. Clone react-native-maps (master@71c74cdd)
2. Run `npm install`
3. Run `rm /example/ios/Podfile.lock` (necessary on my machine in order for `yoga` to be installed properly, will probably make another issue around this later).
4. Run `run build:ios`.
5. Open `/example/ios/AirMapsExplorer.xcworkspace` in Xcode 9.3 (9E145).
6. Select any simulator (eg. iPhone 8) as the target and select **Product > Build**.

### Expected Behavior
Project would build without error.

### Actual Behavior
Project builds with the following error:

```
~/react-native-maps/lib/ios/AirGoogleMaps/AIRGoogleMap.m:452:35: Illegal type 'NSDate *' used in a boxed expression
```

The line in question is here:

```
    id event = @{@"coordinate": @{
                    @"latitude": @(location.coordinate.latitude),
                    @"longitude": @(location.coordinate.longitude),
                    @"altitude": @(location.altitude),
                    @"timestamp": @(location.timestamp), // <------- THIS LINE FAILS
                    @"accuracy": @(location.horizontalAccuracy),
                    @"altitudeAccuracy": @(location.verticalAccuracy),
                    @"speed": @(location.speed),
                    }
                };
```

I am able to "fix" it by changing the line to be `@(location.timestamp.timeIntervalSinceReferenceDate)`, though I'm not sure if this is what is actually desired.
 
### Reproducible Demo
N/A

rborn:
@cmbasnett mind a PR ?
[...]

PR #2232, opened 2018-05-01
Fix for compile error (#2215).

cmbasnett:
No.

https://github.com/react-community/react-native-maps/issues/2215

### How did you test this PR?
Compiled for iOS Simulator in XCode 9.3.


cmbasnett:
Had a quick look at the Android equivalent, and it looks like that this may introduce a disparity between the `timestamp` values (or there may have already been one to begin with). Android uses [Location.getTime()](https://developer.android.com/reference/android/location/Location.html#getTime()) which is in milliseconds, while iOS's [timeIntervalSinceReferenceDate](https://developer.apple.com/documentation/foundation/nsdate/1417376-timeintervalsincereferencedate) is in seconds. I'll leave it up to whoever is reviewing this to make the call.

rborn:
@cmbasnett thanks for this ❤️

Taking in account that the android version was before and might be used in production we should adapt the iOs one to it, so if you could make it in milliseconds would be amazing.
Also if this information is needed in the docs, please add it there too

Thanks again

cmbasnett:
@rborn Should be sorted out now. Have a look and let me know if anything else is needed.

PR #2238, opened 2018-05-05
Support the asset scheme for images

janicduplessis:

No

Expo uses the `asset://` scheme to load images from the app's bundle on Android. This is supported by Fresco (http://frescolib.org/docs/supported-uris.html) so we can simply add the new scheme to the list that we load using Fresco.

### How did you test this PR?

Verified that bundled assets in Expo work with the Marker component by applying this diff locally.

This should also work everywhere else we allow assets with the file:// scheme.


rborn:
LGTM @alvelig

PR #2248, opened 2018-05-11
Make tiles display at the same physical size regardless of pixel dens…

funkyfourier:
…ity on iOS devices

Not that I know of.

https://github.com/react-community/react-native-maps/issues/2242

### How did you test this PR?

This makes UrlTile, when using Google Maps API, behave the same way on iOS as it currently does on Android. That is, displaying tiles at the same physical size regardless of pixel density.

Have tested on simulators iPhone X and iPhone 6, as well as a physical iPhone 5S device.

Here is a snack, which should also work without Expo: https://snack.expo.io/@jmr/urltile-ios-tile-size

Without my modification it will display the tiles very tiny on high dpi devices, and with the modification it will display them at the same size regardless.


rborn:
LGTM @alvelig

alvelig:
LGTM

PR #2253, opened 2018-05-14
TestID's for e2e automation

compojoom:
It seems that the following pull https://github.com/react-community/react-native-maps/pull/1170 request has been abandoned, but testID is still required if one wants to make e2e tests

https://github.com/react-community/react-native-maps/pull/1170 

 - the previous PR is stalled, as it's really old and the author didn't continue working on it
 - there are conflicts with the `master` branch and the author didn't fix them

rendered markers don't have testID prop which is necessary for e2e tests
Closes https://github.com/react-community/react-native-maps/issues/2252
Closes https://github.com/react-community/react-native-maps/issues/1104
### How did you test this PR?
1. Create a marker and add it a testID
2. Launch your app in ios simulator, then in XCode navigate to the debug navigator and press the debug View Hierarchy button. This will create a snapshot of your UI similar to this:
grafik

Select the marker and it should have an identifier set in the Accessibility section on the right.

This was tested on a real device and simulator. The change applies to ios only.


andyesp:
Any chance to review and merge this? Would be great to allow Detox, for example, to find markers.

AngelVenchev:
Hey, just a quick update for anyone wondering about this. 'react-native-maps' currently supports the testID prop for markers for google maps on iOS (v0.23.0). However markers are not interactable, meaning you cannot tap on them.

There are two workarounds I am currently implementing:
1. TapAtPoint, assuming I know where the marker is drawn on the map.
2. Custom marker view and adding the testID on that and trying to tap on that. This one is not tested.

I am using the first approach because in my e2e tests I am able to create all my markers and then I can test with the newly created ones.

danielgindi:
Currently with UITests (XCTest) I can't even test for the marker's existence via its id. I don't think its working.

PR #2314, opened 2018-06-10
Adding onPanDrag compatibility for iOS

jherencia:

No, not any other PR does this.

Currently, onPanDrag does not work in iOS, this PR adds this behavior.

### How did you test this PR?

- Which platform (eg. Android/iOS) & Maps API (eg. Google/Apple) does this change affect, if any?
IOS + Google Map

- Did you test this on a real device, or in a simulator?
Both

- Are there any platforms you were not able to test?
No.

### DIFF
--- a/lib/ios/AirGoogleMaps/AIRGoogleMap.h
+++ b/lib/ios/AirGoogleMaps/AIRGoogleMap.h
@@ -26,6 +26,7 @@
 @property (nonatomic, copy) RCTBubblingEventBlock onKmlReady;
 @property (nonatomic, copy) RCTBubblingEventBlock onPress;
 @property (nonatomic, copy) RCTBubblingEventBlock onLongPress;
+@property (nonatomic, copy) RCTBubblingEventBlock onPanDrag;
 @property (nonatomic, copy) RCTBubblingEventBlock onUserLocationChange;
 @property (nonatomic, copy) RCTBubblingEventBlock onMarkerPress;
 @property (nonatomic, copy) RCTBubblingEventBlock onChange;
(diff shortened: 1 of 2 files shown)


rborn:
@jherencia thanks for the PR., few questions:

- is the event working on ios and apple maps? 
- is the event working on android? 
- do we need to adapt the docs in any way?

rborn:
@jherencia 

> The documentation currently does not mention anything about this issue in iOS, it says that you need to use scrollEnabled, but it is not right.

Could you update the docs? And what's wrong? Maybe its for Apple maps ?

mrtaheri:
Hi,
we need this option for production, so when can you merge it and release a new version, contains this changes??

oliverdolgener:
Too bad it's not in the recent release :(

danrivett:
I too am blocked by this. Was taken aback it didn't work too as from what I read, I though I just needed to disable `scrollEnabled` but that's not working for me, so assume this PR is required.

Would really appreciate a committer review this PR.

danrivett:
As a follow-up, slightly tangentally, where isn't there an `onPanDragEnd` event like for markers? Even with this PR, I'm struggling to understand how to detect the end of a drag.

I'm guessing I'm going to have to schedule timeout and keep track of the timestamp of last `onPanDrag` event and once the elapsed time crosses a threshold, assume end of drag. Seems far from ideal as user could keep the finger pressed but just paused dragging.

Be much better to get notified when the user lifts their finger. 🤔

gbmahili:
Is there any progress on this? Can we have it in the next release?

chakrihacker:
Please merge this

salah-ghanim:
merged into master

Issue #2048, opened 2018-02-28
Custom markers doesn't always update on android


vvusts:
RN - 0.52
RN Maps - 0.20.1
IOS - work perfect
Android - is problem

I render a lot of markers and use clustering (react-native-maps-super-cluster) but this is related to map.
Markers use custom images like: **marker.png** and **selectedMarker.png**
I have list of buttons bellow map. When button is pressed it` .animateToRegion()` where some marker is.
Selecting button update state variable like `selectedMarkerId`.
In function that render marker I check this variable and based on that set different marker image.
Problem is that on android this image replace doesn't work. And then I move map so marker is not visible and back it again and then image is updated.
I tried with setting key to be `Math.random()` but this make marker flicker as I move map and change selections.
Does anybody have problem like this and is there solution?

vuphuctho:
Same issue with you but in my case I don't call `animateToRegion` but update region by changing state. using `Math.random()` for updating key values work for me, no more flickering happens.

vvusts:
Where exactly you set math.random?
On marker key? On marker image?
I set id of location from database for marker key so for me that is not random. I set on custom image but then I get flickering.

bramus:
Could it be that this is a duplicate of https://github.com/react-community/react-native-maps/issues/1288 ?

vvusts:
No, markers are moving fine it’s image for marker that is not update.

For example: you have two markers in a view region and two buttons one should update image one one and second on another. 
Click on button update once. 
But when you click on second button marker image is not changed. 
But if you move map so that marker is not visible and back agin it update marker image. 
And this is only on android.

bramus:
Ah, thanks for the clarification … I get it now :)

vuphuctho:
@vvusts I updated marker key, based on [this explanation.](https://github.com/react-community/react-native-maps/issues/1611#issuecomment-334619684)

Khsed4:
@vvusts I'm stuck the same issue , have u found the solution ? if yes kindly post it here

khuongdv:
@vuphuctho I tried that way but in real devices the issue still occurs @@

abhideepmallick:
Working solution https://github.com/react-native-community/react-native-maps/issues/2734#issuecomment-515428771

Issue #1611, opened 2017-09-01
Custom Markers does not update after props change on Android


levelasquez:
Hi everyone,

I'm having an issue on Android where the custom marker does not update after its props have change.

When i press the marker, its marked like selected and vice versa. Then its backgroundColor changes to identify that it has beed selected or not. This is working fine on iOS - Mapkit but on Android - Google Map does not update the change.

On Android - Google Map. Does not update the change.

![android](https://user-images.githubusercontent.com/5843706/29985669-54ca3b2e-8f36-11e7-876e-776b46b8040f.gif)

On iOS - Mapkit works perfectly. This is the expected behavior.

![ios](https://user-images.githubusercontent.com/5843706/29985731-9920d468-8f36-11e7-947d-5654ad21751f.gif)

Here is my code.

State and method
```javascript
state = {
    selected: [],
  }

toggleSelect(id) {
    this.state.selected.indexOf(id) > -1
      ? this.setState({ selected: this.state.selected.filter(x => x !== id) })
      : this.setState({ selected: this.state.selected.concat(id) })
  }
```

MapView Component
```javascript
<MapView
                style={styles.contentTabView}
                initialRegion={{
                  latitude: -33.447984,
                  longitude: -70.659059,
                  latitudeDelta: 0.0922,
                  longitudeDelta: 0.0421,
                }}
              >
                {data.map(x =>
                  <MapView.Marker
                    key={x._id}
                    coordinate={{
                      longitude: x.location.coordinates[0],
                      latitude: x.location.coordinates[1],
                    }}
                    title={x.name}
                    centerOffset={{ x: 0, y: -20 }}
                    onPress={() => !(x.origin || x.destination) && this.toggleSelect(x._id)}
                  >

                    <Pin
                      pinStyle='pinIcon'
                      sourceIcon={x.clasification.icon}
                      id={x._id}
                      selected={selected}
                      origin={x.origin}
                      destination={x.destination}
                      color={x.timePeriod.color}
                    />
                  </MapView.Marker>
                )}
              </MapView>
```

Custom Marker Component
```javascript
<View style={styles[pinStyle].container}>
        <View style={[
          styles[pinStyle].wrapper,
          wrapperColor,
          isSelected(id, selected) && styles[pinStyle].pointSelected,
          (origin || destination) && styles[pinStyle].pointSelected,
        ]}>
          <View style={styles[pinStyle].inner}>
            {type === 'icon'
              ? <IconAwesome
[...]

PR #2320, opened 2018-06-11
Fix custom marker updates on android #1611 #2048

fqborges:
No

On Android, using Google Maps, markers with custom views/images are not being updated correctly when other properties are changed (mainly coordinates). Sometimes markers are not draw anymore, until the child view/image is changed.

#1611 and #2048 are related.

### How did you test this PR?

1. Create a simple app with a MapView and 3 or more markers that have distinct views for normal/selected states with distinct sizes.
2. After MapView first rendered, dispatch changes to the markers coordinates and selected state (use a timer, button or marker touch event).
3. Expect markers to not update or disappear from map after a couple of changes.
4. Apply this patch and markers update as expected.

- Which platform/api: Android & Google
- Did you test this on a real device, or in a simulator? Simulator and Real device

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
@@ -300,8 +300,10 @@ public class AirMapMarker extends AirMapFeature {
 
   public MarkerOptions getMarkerOptions() {
     if (markerOptions == null) {
-      markerOptions = createMarkerOptions();
+      markerOptions = new MarkerOptions();
     }
+
+    fillMarkerOptions(markerOptions);
     return markerOptions;
   }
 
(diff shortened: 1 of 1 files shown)


rborn:
LGTM @alvelig 🐽

alvelig:
@fqborges Wow! So the problem was just `new MarkerOptions()` that were empty in certain situations?

fqborges:
@alvelig Actually not empty, just outdated. And there was a chance of the outdated referenced bitmaps getting collected and not being valid for drawing anymore.

I found that in certain updates the markers are added then removed from map, instead of just get the properties updated. In this cases, the marker options used are the original created, with the initial values.

alvelig:
Great job. I'll test and maybe close even more issues. Thank you!

PR #2326, opened 2018-06-13
Note that children components can be added

ericjames:
I literally just trial and errorred this after banging my head.

(please answer here)

(please link the issue here)

### How did you test this PR?

(please answer here)

### DIFF
--- a/docs/marker.md
+++ b/docs/marker.md
@@ -62,3 +62,16 @@ type Point {
   y: Number,
 }
 ```
+
+## Children Components
+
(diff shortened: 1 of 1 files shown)


docs/marker.md
```diff
@@ -62,3 +62,16 @@ type Point {
   y: Number,
 }
 ```
+
+## Children Components
+
+Children components can be added within a Marker and rendered content will replace the marker symbol.  This is a way of creating custom markers and allowing use of native SVGs.
```
rborn: Maybe try to be consistent between the explanation and the example? 
Or make the explanation more broad?

Just my 2c

Issue #2264, opened 2018-05-16
how to update react-native-maps to work with create-react-native-app


rss81:
### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

(yes)

### Steps to Reproduce

1. run create-react-native-app

### Expected Behavior

expect to get an clean operating environment
### Actual Behavior

got, react-native-maps requires peer dependency react 0.51,0.52,0.53 or 0.54

When I run create-react-native-app I'm getting a peer dependency for react-native-maps to use react-native@0.54. I need to be able to use react-native@0.55.2 and react@16.3.1

How should I edit  the package.json file for react-native-maps. Notice line 81,82,97 and 98

could someone show how to edit this json file and how to update react-native-maps? When I tried with:

npm --depth 9999 update
I ended up getting tons of errors about duplicate packages.

rss81:
ok, so if we just update the package.json file for react-native-maps in the following way:

    under devDependencies:
        "react": "^16.3.1",
        "react-native": "^0.55.2"

    under peerDependencies:
        "react": "^16.3.1",
        "react-native": "^0.55",

then back in the top project install eslint@4.0.0 and ajv@6.0.0 and add them to the dependencies in the package.json.

could someone make these changes in the official github package for react-native-maps?

rborn:
I’m closing this issue because it has been inactive for a while. This probably means that it is not reproducible or it has been fixed in a newer version. If it’s an enhancement and hasn’t been taken on for so long, then it seems no one has the time to implement this.

Please reopen if you still encounter this issue with [the latest stable version](https://github.com/react-native-community/react-native-maps/releases).

Thank you!

PR #2332, opened 2018-06-15
Add react-native@^0.55 to peerDependencies

mikemorris:

No.

#2264 

### How did you test this PR?

Installing react-native-maps with this patch applied in an app with `react-native@0.55.4` and `react@16.3.1` as dependencies.

With this patch applied, the following install error is no longer present.

```
npm WARN react-native-maps@0.21.0 requires a peer of react-native@^0.51 || ^0.52 || ^0.53 || ^0.54 but none is installed. You must install peer dependencies yourself.

npm ERR! Cannot read property '0' of undefined
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   ],
   "peerDependencies": {
     "react": "^16.0",
-    "react-native": "^0.51 || ^0.52 || ^0.53 || ^0.54",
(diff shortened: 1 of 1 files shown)


rborn:
LGTM @alvelig 🐽

alvelig:
LGMT. I think next update could be more strict with deps, just cut off all below 55.

Issue #2296, opened 2018-06-01
[Android] Could not build lib from source


fqborges:
How to reproduce:

```
git clone https://github.com/react-community/react-native-maps.git
cd react-native-maps
yarn install
yarn build:android
```

result:
```
FAILURE: Build failed with an exception.

* Where:
Build file 'C:\git\react-native-maps\lib\android\build.gradle' line: 46

* What went wrong:
A problem occurred evaluating project ':react-native-maps-lib'.
> Could not find method compileOnly() for arguments [com.facebook.react:react-native:+] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
```

fqborges:
This is related to this pull https://github.com/react-community/react-native-maps/pull/2096 since it updated gradle files that are used when this lib is included on rn apps, but not the gradle files to build the lib itself.

adhavalboy:
got to react-native-maps\lib\android\build.gradle'

and change compileOnly() with provided 
and implementation as compile

PR #2342, opened 2018-06-19
update android build configuration for gradle 3+ in example project

fqborges:
No

#2296 

It was not possible to just clone the repo, build the lib, and run the example project using gradle or android studio.

### How did you test this PR?
git clone https://github.com/fqborges/react-native-maps.git
cd react-native-maps
git checkout fix-gradle3-build
yarn install
yarn build:android
yarn run:android

**Edit:**
As discussed bellow, the latest version of this lib is already using Gradle 3, which can lead to build issues when integrating with other libs ([react-navigation comes to mind](https://github.com/react-navigation/react-navigation/issues/3097)) caused by how [react-native package the resources assets](https://github.com/facebook/react-native/issues/16906).

An alternative solution would be roll back the changes made on #2096

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,10 +2,14 @@
 
 buildscript {
   repositories {
+    // Gradle 4.1 and higher include support for Google's Maven repo using
+    // the google() method. And you need to include this repo to download
+    // Android plugin 3.0.0 or higher.
+    google()
     jcenter()
   }
   dependencies {
-    classpath 'com.android.tools.build:gradle:2.3.2'
+    classpath 'com.android.tools.build:gradle:3.1.2'
   }
 }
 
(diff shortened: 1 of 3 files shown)


Moreno97:
RN doesn't support Gradle3+ versions at the moment, this change will lead to break builds.

fqborges:
Unfortunately, the build is already broken. The example app does not build (see https://github.com/react-community/react-native-maps/issues/2296), and this PR doesn't touch the [lib build.gradle file](https://github.com/react-community/react-native-maps/blob/master/lib/android/build.gradle), that already is using gradle 3.

Moreno97:
Yes, is already broken. They should roll back the changes made with #2096

friederbluemle:
@alevig @rborn - This also has been superseded by #2720, #2727, and #2765 (and others), and can safely be closed. 👍

PR #2390, opened 2018-07-19
Consider pixel density in coordinate<->point conversion (Android)

danielgindi:
Nope

`pointForCoordinate` and `coordinateForPoint` use raw pixel values, which is inconsistent with other functions and with iOS implementation.

### How did you test this PR?

This affects Android code only, and was tested locally in a sample app.


lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
```diff
@@ -142,12 +142,14 @@ public void onSnapshotReady(@Nullable Bitmap snapshot) {
 
   @ReactMethod
   public void pointForCoordinate(final int tag, ReadableMap coordinate, final Promise promise) {
+    final ReactApplicationContext context = getReactApplicationContext();
+    double density = (double) context.getResources().getDisplayMetrics().density;
```
jamiesonbates: I applied these changes to my local node module and was unable to build for android. 

I believe this line needs read `final double density = (double) [context.getResources().getDisplayMetrics().density;`

Once I made this change to the local file, I was able to build. I can also confirm that the PR's code does fix the issue described for android.

danielgindi:
Thanks, fixed!

danielgindi:
There are not releases since April, seems like the guys are busy :-)

rborn:
LGTM @alvelig 🐽

PR #2392, opened 2018-07-24
add support for loading base64 encoded image

muhamad-rizki:

(please answer here)
No PR do same thing is stalled

(please link the issue here)
Cannot load base64 encoded image to Marker component on android
### How did you test this PR?

(please answer here)
- tested on real device with android version 7.0
- ios not tested


zhaomingli007:
@christopherdro  would you consider to fix this as well?
https://github.com/muhamad-rizki/react-native-maps/blob/4dbdc1933aae99c2c601b8e59ab61e3d1f438909/lib/android/src/main/java/com/airbnb/android/react/maps/ImageReader.java#L92

PR #2310, opened 2018-06-08
Remove AirGoogleMaps from AirMaps.xcodeproj

mikemorris:
### Does any other open PR do the same thing?

No

### What issue(s) is this PR fixing?

#2159, #2270 

The non-Cocoapods manual install instructions appear to be gone, but IIRC it was previously required to manually link the Google Maps libraries in your Xcode project if you wished to use Google Maps on iOS - this patch restores that behaviour.

### How did you test this PR?

`npm link` with the patch applied on `master` into a project where I was reproducing #2270, [after using `git bisect` to identify the likely first bad commit](https://github.com/react-community/react-native-maps/issues/2270#issuecomment-395903178).

PR #2396, opened 2018-07-25
Enable custom configuration for iOS maps

wbyoung:
This allows setting up the library to work with Google Maps on iOS via `react-native link` instead of only via CocoaPods.

Because Google-Maps-iOS-Utils is not distributed as a binary (at this time), it disables functionality that depends on this library and raises an exception that points to the reason why there's a problem.

Additionally:

- A missing file has been added back into the AirMaps.xcodeproj project.
- The use of `Google-Maps-iOS-Utils` is now based on conditional configuration as well.

Yes, in a sense. https://github.com/react-community/react-native-maps/pull/2310 is related, and I believe it should be closed in favor of this request.

https://github.com/react-community/react-native-maps/issues/2159, https://github.com/react-community/react-native-maps/issues/2270, https://github.com/react-community/react-native-maps/issues/2289, https://github.com/react-community/react-native-maps/issues/2066 (for users who are upgrading from an older version & perhaps others).

### How did you test this PR?

I tried using it with and without the included script being run to ensure that it compiles and links properly. I did not, however, test it with CocoaPods and will request that someone who has that configured ensure that everything still works as expected with Google Maps enabled (and without) when using CocoaPods.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -80,20 +80,24 @@ post_install do |installer|
 end
 ~~~
 
+## iOS - ReactNative Link
 
-## IMPORTANT!!
+Run `react-native link react-native-maps`. Note that by default this will use
+Apple Maps and that the configuration of Google Maps will be more difficult.
 
-**!!  DO NOT USE  !!** `react-native link`
+Functionality that depends on `Google-Maps-iOS-Utils` has been disabled for this
+configuration via runtime errors due to the fact that this framework is not
+available for download as a pre-compiled binary. An exception will be raised if
+you try to use the following features:
 
-Have ran it already? Read [this](#on-ios).
+- Making markers via KML files
 
 ## If you want to use Google maps
-
 
 Add to `ios/_YOUR_PROJECT_NAME_/AppDelegate.m:
 
 ```objc
-+ @import GoogleMaps; //add this line if you want to use Google Maps
++ #import <GoogleMaps/GoogleMaps.h>
 
 @implementation AppDelegate
 ...
@@ -105,6 +109,25 @@ Add to `ios/_YOUR_PROJECT_NAME_/AppDelegate.m:
 ```
 
 This should be the **first line** of the method.
+
+#### If you are not using CocoaPods
+
+If you installed via `react-native-link`, add the following to your
+`package.json` and replace the `REPLACE_ME_RELATIVE_PATH_TO_GOOGLE_MAPS_INSTALL`
+with the relative path from your project root to the directory in which you
+installed the Google Maps frameworks:
+
+```json
+{
+  "name": "your-app",
+  "scripts": {
+    "postinstall": "./node_modules/react-native-maps/enable-google-maps REPLACE_ME_RELATIVE_PATH_TO_GOOGLE_MAPS_INSTALL"
+  }
+}
+```
+
+Re-run `npm install` or `yarn` to ensure the `postinstall` script is run.
+
 
 ## Notes on running on a real ios device
 
@@ -207,16 +230,6 @@ If you have a blank map issue, ([#118](https://github.com/airbnb/react-native-ma
 ### On iOS:
 
 If google logo/markers/polylines etc are displayed, this is likely an API key issue. Verify your API keys and their restrictions. Ensure the native `provideAPIKey` call is the first line of `didFinishLaunchingWithOptions`.
-
-If you have ran 'react-native link` by mistake:
-
-1. delete node_modules
-2. delete ios/Pods
-3. delete ios/Podfile.lock
-4. open Xcode and delete `AIRMaps.xcodeproj` from Libraries if it exists
-5. in Build Phases -> Link Binary With Libraries delete `libAIRMaps.a` if it exists
-6. delete ios/build folder
-7. start again with the installation steps
 
 If you use Xcode with version less than 9 you may get `use of undeclared identifier 'MKMapTypeMutedStandard'` or `Entry, ":CFBundleIdentifier", Does Not Exist` errors. In this case you have to update your Xcode.
 
--- a/enable-google-maps
+++ b/enable-google-maps
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+GOOGLE_MAPS_INSTALL_LOCATION=$1
+
+if [[ -z "$GOOGLE_MAPS_INSTALL_LOCATION" ]]; then
+  echo "usage: enable-google-maps <google-frameworks-relative-install-dir>"
+  exit 1
+fi
+
+cat > lib/ios/User.xcconfig <<EOF
+FRAMEWORK_SEARCH_PATHS = \$(inherited) \$(SRCROOT)/../../../../$GOOGLE_MAPS_INSTALL_LOCATION
+GCC_PREPROCESSOR_DEFINITIONS = \$(inherited) HAVE_GOOGLE_MAPS=1
+EOF
--- a/lib/ios/AirGoogleMaps/AIRDummyView.h
+++ b/lib/ios/AirGoogleMaps/AIRDummyView.h
@@ -5,6 +5,8 @@
 //  Created by Gil Birman on 10/4/16.
 //
 
+#ifdef HAVE_GOOGLE_MAPS
+
 #import <UIKit/UIKit.h>
 
 
@@ -12,3 +14,5 @@
 @property (nonatomic, weak) UIView *view;
 - (instancetype)initWithView:(UIView*)view;
 @end
+
+#endif
--- a/lib/ios/AirGoogleMaps/AIRDummyView.m
+++ b/lib/ios/AirGoogleMaps/AIRDummyView.m
@@ -5,6 +5,8 @@
 //  Created by Gil Birman on 10/4/16.
 //
 
+#ifdef HAVE_GOOGLE_MAPS
+
 #import <Foundation/Foundation.h>
 #import "AIRDummyView.h"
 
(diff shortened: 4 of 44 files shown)


thanhdevapp:
Can you give me like step by step using gmap on ios

rborn:
@wbyoung what about an update form an older version of the module? Did you test this?

@alvelig I like this PR and I'll try to play with it but I need another pair of eyes 🐽

rborn:
@wbyoung also what if we don't want google maps, does this PR handles the `react-native link` correctly? and how?
🤗

wbyoung:
@rborn I haven't used an older version of this module, but based on the comments in #2310, I would think that this a \<insert word for whatever the opposite of breaking change is here\> since it actually intends to improve the upgrade experience for prior users (in the same spirit as #2310). From what I can tell (and I haven't used the project except for the current version), the following is true:

- Google maps used to be an optional dependency of this project requiring the addition of a specific set of source files or a specific `.xcodeproj` (and perhaps linker flags and/or build steps to go with it).
- At some point, the library moved to Cocoapods specific instructions for setup and more or less removed both the ability to `react-native link` as well as (perhaps unintentionally) the need for or use of the Xcode projects that would be used by `react-native link` since all of the Cocoapods config files simply point to raw sources.
- Made the Google Maps optional setup work via Cocoapods configs that point to which source files to compile (ignore Google Maps specific files or not) and not the `.xcodeproj`.
- Broke the upgrade path for existing `react-native link` users who did not use Google Maps since some files and/or code was added to the now presumably unused (but actually included for prior users) `.xcodeproj` file.

So the long answer, is that I considered this when making the changes, but the short answer is: no, I did not _test_ this (mainly because I didn't use this library before now).

The gist of what's happening here is that we're setting up for conditional compilation of all of the Google Maps files where the static library will still be compiled, but it won't actually contain anything pertaining to Google. This works for everything except the code that's for `Google-Maps-iOS-Utils` since there's no pre-built binary for that. Of course one could pre-compile it or clone the source themselves and then overwrite or append to the `User.xcconfig` inside of `node_modules` to have `HAVE_GOOGLE_MAPS_UTILS=1` as well as `HAVE_GOOGLE_MAPS=1`.

The conditional `#ifdef` compilation strategy (if new to you) is pretty standard in `./configure` based open source packages that I've worked with before, and bringing it to Xcode projects is a little annoying since they have weird special setup with their `.xcconfig` files. It is what it is, I guess.

Let me know if I can help explain any better.

rborn:
@christopherdro @alvelig I could use some help here, maybe we can push this forward as it seems a great improvement 🐽

alvelig:
So we can have `react-native link` as a standard method with no hustle?

docs/installation.md
```diff
@@ -106,6 +110,25 @@ Add to `ios/_YOUR_PROJECT_NAME_/AppDelegate.m:
...
+`package.json` and replace the `REPLACE_ME_RELATIVE_PATH_TO_GOOGLE_MAPS_INSTALL`
+with the relative path from your project root to the directory in which you
+installed the Google Maps frameworks:
+
+```json
+{
+  "name": "your-app",
+  "scripts": {
+    "postinstall": "./node_modules/react-native-maps/enable-google-maps REPLACE_ME_RELATIVE_PATH_TO_GOOGLE_MAPS_INSTALL"
+  }
```
alvelig: we may add this to lib if we define a standard method of installation @rborn

wbyoung:
@alvelig

> So we can have react-native link as a standard method with no hustle?

Yup!

> we may add this to lib if we define a standard method of installation

Not sure what you mean here, but if you mean adding it to the install scripts for the library, that would kind of defeat the purpose as the goal is to make Google Maps usage opt in (and for non Cocoapods users, to make it clear that there’s _extra work_ that must be done _following an exact procedure_ to get things working for Google Maps) while keeping it as pain free as possible and allowing a friction free upgrade path.

docs/installation.md
rborn: @alvelig @wbyoung I think we need to educate the user how to install the google maps sdk (I have no idea for example right now at 1:45 AM and some 🍻)

rborn:
@wbyoung I think I need some more instructions from you on how to actually install  it :(
I tried to add your PR with yarn, used the pod files we already have in instruction, run pod install, run your post script manually with some path I downloaded the sdk manually https://developers.google.com/maps/documentation/ios-sdk/start but for some reason the app still shows the apple map. 🤗

Brand new project.

wbyoung:
@rbron if not installed correctly, you should get compiler errors as the script will enable code that uses those headers and requires the symbols from those frameworks in order to link the binary. So you likely got it right.

Installing the framework should literally be downloading it and placing it in a folder, but giving instructions may be difficult since Goolgle’s instructions are more comprehensive, so linking out means also saying “ignore all the stuff here”. So maybe the best course of action is to explain that you just need the files, `X`, `Y`, and `Z` from Google which you can download “here” while ignoring all other install instructions.

That being said, I had the same issue at some point because I didn’t quite understand that enabling google maps didn’t change the default map and that you needed to use a specific provider. Perhaps making this type of change should be done as a separate PR as it’s a bit outside the scope? This may or may not be your issue, but it certainly slowed me down getting things set up.

alvelig:
@wbyoung I agree. This should be kind of Facebook SDK installation. Like place GoogleSDK folder in ~/Documents and add header search path...

rborn:
I set the `provider="google"` and got a black screen.

wbyoung:
@rborn can you push what you currently have to GitHub and post a link here? I can take a look and see what differs from what I have working. It could also double as an example of how to get up and running this way & we can include a link in the docs (if you guys want to continue to maintain that example).

rborn:
@wbyoung I'll try on a clean project, maybe I messed up something.

rborn:
@wbyoung I was wrong, all works fine, it was me 😆 

@alvelig @christopherdro I think we should merge this one, create a new release with **possible breaking** warning for people updating from an older version who added libs manually.
What do you say?

alvelig:
@rborn I concur.

chrise86:
@wbyoung I tried this and it all compiles and runs fine, but when trying to use a marker it produces the following error:

```
Invariant Violation: Invariant Violation: requireNativeComponent:"AIRMapMarker" was not found in the UIManager.
```

rborn:
@wbyoung I didn't use any marker yet, I'll let you know soon. 
@chrise86 - no, we don't need googlemaps for that

chrise86:
@wbyoung I just created a pull request against your branch to fix the issue above. There were `#ifdef HAVE_GOOGLE_MAPS` checks missing in `AIRGoogleMapOverlay.h` and `AIRGoogleMapOverlay.m`.

wbyoung:
@chrise86 I added your changes to this PR just now.

rborn:
@wbyoung can you have a look at https://github.com/react-community/react-native-maps/pull/2455 and let us know what you think? maybe we can make a mix of both?

cc @alvelig

wbyoung:
@rborn [commented over there](https://github.com/react-community/react-native-maps/pull/2455#issuecomment-415876830).

/cc @alvelig

cayasso:
Hi guys! any update on merging this wonderful PR ?

mikemorris:
This appears to be working for me, thanks for your working on this @wbyoung! Closing #2310 in favor this instead.

jamesone:
Any idea when this PR will be merged?

alvelig:
LGTM. Actually a killer feature we'd need. So let's merge.

abarisic86:
It would be helpful to add two things for Google Maps on iOS option:
- explanation what it means to install Google map framework or actually [how to install google maps sdk](https://developers.google.com/maps/documentation/ios-sdk/start)
- instruction how to add Google Maps key as an environment variable

rborn:
@abarisic86 fancy a PR? 😊

abarisic86:
@rborn yes, I would like to contribute but I wasn't able to get it working (react native link with google maps). As if some part of the installation process is skipped in the documentation.

docs/installation.md
abarisic86: @wbyoung can you help a bit with @rborn suggestion, I wasn't able to do it myself in my own project (only inside a clean repo) so I'm not sure am I doing it right

annelorraineuy: I don't understand how to get `REPLACE_ME_RELATIVE_PATH_TO_GOOGLE_MAPS_INSTALL`. Does someone have more info about this? Is it to our `ios` folder?

rborn: @annelorraineuy I have the google maps sdk downloaded to `/Work/sdks/GoogleMaps-2.7.0/` so you need to put that path instead of the `REPLACE_ME_RELATIVE_PATH_TO_GOOGLE_MAPS_INSTALL`

wbyoung:
@abarisic86 I tried to make the docs clear, but there is certainly some language that can be confusing (especially for those who haven't developed for iOS). I went back and reviewed the docs, but I can't really find anywhere to make things more clear. An example project could make it more clear, but I don't have the time to create or maintain one that keeps up to date with this project.

Specifically, the issues you had:

- The `@import GoogleMaps` should not make any difference. It should work as it's documented.
- The docs state that you should use a _relative path_ to the installed location, so the use of `ios/` is exactly as the docs are suggesting. Is there a way that you'd word that to make it more clear?

docs/installation.md
wbyoung: @annelorraineuy if you got it working & have documentation improvements, please send a PR. I followed the style of various documentation sources that I've used over time to try to make it clear and succinct, but it's always good to try to improve for everyone!

termux-io:
`496426c9d47640b9f76bfc2075c7e49975dada15`

PR #2398, opened 2018-07-26
Android cleanup: little improved way of getting SDK values

yeomann:
just a little improved version rather then many explicit def for each value. hence, less cluttering. also made the react-native package to get the dynamic versions as well

Android Target API Level 26 will be required in August 2018.
https://android-developers.googleblog.com/2017/12/improving-app-security-and-performance.html

I was going to write 26 but Since we are already using latest Gradle, Therefore I wrote target values of 27 instead of 25.

And the React Native team is already working on this:
facebook/react-native#18095
facebook/react-native#17741

Not that I could find.

its a android cleanup for less `def` variables mainly but also addressing SDK changes
Android Target API Level 26 will be required in August 2018.
https://android-developers.googleblog.com/2017/12/improving-app-security-and-performance.html

### How did you test this PR?

Real device and simulator both Android and IOS, with (SDK 23 real device), 25 and 26 on Simulators

### DIFF
--- a/lib/android/build.gradle
+++ b/lib/android/build.gradle
@@ -1,19 +1,17 @@
 apply plugin: 'com.android.library'
 apply from: 'gradle-maven-push.gradle'
 
-def DEFAULT_COMPILE_SDK_VERSION             = 25
-def DEFAULT_BUILD_TOOLS_VERSION             = "25.0.3"
-def DEFAULT_TARGET_SDK_VERSION              = 25
-def DEFAULT_GOOGLE_PLAY_SERVICES_VERSION    = "10.2.4"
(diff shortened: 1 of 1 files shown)


salah-ghanim:
how come there is no activity here, this is blocking our Android SDK update

alvelig:
I'm not sure we need this change. For now versions are in a separate place altogether. @rborn what do you think?

rborn:
@alvelig not sure tbh, my android-fu is very rusty :(

friederbluemle:
@alevig @rborn - This can safely be closed, as this (and more) was added in #2720, #2727, and #2765

PR #2433, opened 2018-08-16
[iOS] Fix coordinateForPoint returns nothing (Google Maps)

StasDoskalenko:
https://github.com/react-community/react-native-maps/pull/2039
this PR is not suitable for iOS Google Maps

1) The absence of returned Promise
2) `AIRGoogleMapManager.coordinateForPoint was called with 2 arguments but expects 3 arguments. If you haven't changed this method yourself, this usually means that your versions of the native code and JavaScript code are out of sync. Updating both should make this error go away.`

### How did you test this PR?

Affects the only iOS, works fine both for Google Maps and Apple Maps

### DIFF
--- a/lib/components/MapView.js
+++ b/lib/components/MapView.js
@@ -696,7 +696,15 @@ class MapView extends React.Component {
     if (Platform.OS === 'android') {
       return NativeModules.AirMapModule.coordinateForPoint(this._getHandle(), point);
     } else if (Platform.OS === 'ios') {
-      return this._runCommand('coordinateForPoint', [point]);
+      return new Promise((resolve, reject) => {
+        this._runCommand('coordinateForPoint', [point, (err, res) => {
+          if (err) {
+            reject(err)
+          } else {
+            resolve(res)
+          }
+        }]);
+      })
     }
     return Promise.reject('coordinateForPoint not supported on this platform');
   }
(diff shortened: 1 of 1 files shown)


danielgindi:
I think you are not working with a latest version (or maybe some changes did not get a release version yet). 
As there WAS a problem with one version taking a callback and the other taking a callback, it does not exist anymore.
If you'll look at the native code you'll see that these always return a Promise.

StasDoskalenko:
@danielgindi okay, thanks! I'll take a look again 👍

mbrehme:
I am not sure about the last commit but the first one really fixes the issue with `coordinateForPoint` on ios
see workarround described here: https://stackoverflow.com/questions/50510078/react-native-maps-airgooglemapmanager-pointforcoordinate-was-called-with-2-arg#

danielgindi:
Did you try to just work with `master` instead of latest release?

StasDoskalenko:
@danielgindi you should try to update your pods.

danielgindi:
@StasDoskalenko what do you mean? Do you have a problem only after issueing a pod update?

rborn:
This is very old and it seems to be fixed in a later version. I'll close it now but please feel free to re-open in case you still have the issue.

PR #2407, opened 2018-07-31
Use region on navigation method

dorshay6:
### What issue is this PR fixing?

https://github.com/react-community/react-native-maps/issues/2406
@techieyann comment in this PR: https://github.com/react-community/react-native-maps/pull/2049

### How did you test this PR?
WIP: We tested this changes in our app (Android only) Full QA will be done within @autofleet team soon.

Based on: http://luan-itworld.blogspot.com/2013/07/calculate-google-map-zoom-level-on.html

PR #2436, opened 2018-08-17
Implement zoom on animateToNavigation

techieyann:

PR #2435 -- I had no idea the git branch name was used as the pull request title, and it was not informative, and now I've realized I could have just edited the title.. Oh well, I still closed the other one.
PR #2407 -- Use region on navigation method
That PR implements a region which is then converted to Lat/Lng and zoom; and only on Android

My PR implements a zoom argument for both iOS/Android:
 

- regions don't make sense when dealing with rotated map positions, as regions are by definition with bearing 0
- much more direct solution to the lack of zoom option on animateToNavigation()

PR #2049 specifically [my comment](https://github.com/react-community/react-native-maps/pull/2049#issuecomment-403230410)

The animateToNavigation function provides a (much needed) smooth transition to a new location and bearing at the same time, but does not allow for modifying the zoom. If we need to change the zoom/location/bearing at the same time, we're back to the same step by step animations.

This PR just adds a zoom argument to animateToNavigation()

### How did you test this PR?

I tested this on an Android simulator, but don't have the ability to test on iOS presently.
This PR does support both platforms and maps as it was relatively minimal effort.

### DIFF
--- a/lib/ios/AirMaps/AIRMapManager.m
+++ b/lib/ios/AirMaps/AIRMapManager.m
@@ -38,7 +38,7 @@ static NSString *const RCTMapViewKey = @"MapView";
 @end
 
 @implementation AIRMapManager
-  
+
 RCT_EXPORT_MODULE()
 
 - (UIView *)view
@@ -133,6 +133,7 @@ RCT_EXPORT_METHOD(animateToNavigation:(nonnull NSNumber *)reactTag
         withRegion:(MKCoordinateRegion)region
         withBearing:(CGFloat)bearing
         withAngle:(double)angle
+        withZoom:(CGFloat)zoom
         withDuration:(CGFloat)duration)
 {
     [self.bridge.uiManager addUIBlock:^(__unused RCTUIManager *uiManager, NSDictionary<NSNumber *, UIView *> *viewRegistry) {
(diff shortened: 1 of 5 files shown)


lib/ios/AirMaps/AIRMapManager.m
```diff
@@ -144,7 +145,7 @@ - (UIView *)view
             MKMapCamera *mapCamera = [[mapView camera] copy];
              [mapCamera setPitch:angle];
              [mapCamera setHeading:bearing];
-
+             [mapCamera setZoom:zoom];
```
miracle2k: I don't think this will work - MKMapCamera does not have a zoom property - it is called "altitude": https://developer.apple.com/documentation/mapkit/mkmapcamera and it is in meters.

The problem here then is that the zoom level in Google Maps is a completely different measurement than the altitude on iOS. My suggestion would be to initially implement both properties separately: A user wanting to set the height of the camera would need to provide a property "zoom" on Google Maps and an property "altitude" on iOS.

Later, we night want to add the necessary calculations ourselves to convert between the two.

rikur: Those calculations would be very handy to have.

Issue #356, opened 2016-07-04
Get current bounding box (north-east/south-west coordinates)


aforty:
I can keep track of the maps position with `onRegionChange` but that only gives me the coordinates and delta. Is there anyway to retrieve (or calculate?) the north-east/south-west position displayed within the map?

aforty:
I found a solution. For anyone coming across this later...

```
<MapView onRegionChangeComplete={(center) => {
  let northeast = {
      latitude: center.latitude + center.latitudeDelta / 2,
      longitude: center.longitude + center.longitudeDelta / 2,
    }
    , southwest = {
        lat: center.latitude - center.latitudeDelta / 2,
        lng: center.longitude - center.longitudeDelta / 2,
    };

    console.log(center, northeast, southwest);
}} />
```

rgabrielsson:
This doesn't work, first it doesn't comply to the outline of the map coords system. More importantly , however, latitudeDelta does not represent the current view span.

oliverbenns:
This only works if user isn't rotating map.

aforty:
Agh. Well it has been good enough for me, however I would love some sort of official, documented, supported way of getting these values. Seems fairly important to a lot of different use cases.

I'm going to reopen this. Would love one of the developers to comment on this.

jessym:
+1, need this as well.

The code sample above will be a decent approximation near the equator, but it will fail at the southern parts of the southern hemisphere and the northern parts of the northern hemisphere, because the (default) Mercator projection stretches out certain parts of the globe.

For example, you could have Antarctica in the south of your viewport (`latitude = -80`) and Japan in the north of your viewport (`latitude = 36`). The center of the screen (i.e. the projection of Earth we're looking at) would be somewhere south of Australia at `latitude = -49`.

The current maps API only gives us the values `region.latitude = -49` and `region.latitudeDelta = 116`, where `116` is calculated as the difference in latitude between Antarctica and Japan.

However, from these two numbers (`-49` and `116`), it's not possible to get the absolute latitudes of Antarctica and Japan without some clever mathematics by inverting the map projection. Also note that the code sample above fails for these numbers: `-49 - (116 / 2) = -107` (should be `-80`; a number lower than `-90` is not even allowed in fact) and `-49 + (116 / 2) = 9` (should be `36`).

definitelycarter:
👍 - I would also like to know how to do this

1mjwebb:
would exporting these do it?

[...]

PR #2440, opened 2018-08-20
Adding onIdleCamera event for Google Maps

ShivamKJJW:
Calculation to get the bounds of the map northEast and southWest corners are not exact with the help of region data provided with deltas. Hence a new event is being added when the camera is idle at a position. It adds the enhancement claimed under issue #356 

Usage - Wants to show the favourite places inside a particular region when user moves camera.

No.

### How did you test this PR?

Tested via Xcode and Android Studio in both iOS & Android platform for compatibility.


docs/mapview.md
```diff
@@ -50,6 +50,7 @@ To access event data, you will need to use `e.nativeEvent`. For example, `onPres
 | `onKmlReady` | `KmlContainer` | Callback that is called once the kml is fully loaded.
 | `onRegionChange` | `Region` | Callback that is called continuously when the region changes, such as when a user is dragging the map.
 | `onRegionChangeComplete` | `Region` | Callback that is called once when the region changes, such as when the user is done moving the map.
+| `onIdleCamera` | `{ bounding: { lat_max: Number, lon_max: Number, lat_min: Number, lon_min: Number,}, Region }` | Callback that is called once when the camera is idle at its position, such as when the user is done moving the map.(Only for GoogleMaps)
```
rborn: Why this notation and not the same with `getMapBoundaries`?

rborn:
@ShivamKJJW besides my comment LGTM, also please fix the conflict ❤️

@alvelig @christopherdro 🐽

docs/mapview.md
ShivamKJJW: I feel working with max and min is much easier and robust than using northEast and southWest. With the Map rotation there will be no confusion for direction and it can easily be figured out with max/min

rborn: @alvelig @christopherdro any thoughts on this? I tend to go to consistent APIs but maybe you have another opinions?

alvelig: Yes, I'm also pro Region object format rather than introducing not consistent options. Just because of the custom.
Region is easier to treat when zoom (you change 2 parameters instead of 4 and you don't have to calculate center position in that case)

ShivamKJJW: Region parameter is also being passed along with the bounding areas. Bounding area has been added to get the exact lat and lon min/max values which helps in many api requests. There are many PR for such values so I though of  keeping it with same name.

rborn: @ShivamKJJW I agree with you that sometimes it would be easier to use this notation (I also had to deal with it a few times).
However we already have some inconsistencies between the APIs so I wouldn't introduce another one. Somebody who's using region probably already has the adapter in place if they need to convert to X notation. 
Adding a new one would add an extra layer of complexity for everybody - old users would need to re-adapt, new users will be confused by different notations.

As an aside, there is  no good or bad notation, is just a matter of conventions 😀

So it would be amazing if you'd want to move to region as it is now ❤️

ShivamKJJW: @rborn Right yeah! I will change the things and push it when I get time. 😸

christopherdro:
Will this API work now that we're using Google Maps 3.0.3?

rborn:

PR #2445, opened 2018-08-21
coordinateForPoint/pointForCoordinate worked with coordinate shorthands on AirMaps

danielgindi:

No

coordinateForPoint/pointForCoordinate in AirMaps are working differently than Google Maps. The coordinates are lat/lng instead of latitude/longitude, which is a mistake and is incompatible with this library.

### How did you test this PR?

Tested on an iOS 11 simulator, my own project, while trying AirMaps.


rborn:
Thanks for the PR 🤗

I feel this is a breaking change, an update of the lib will break any code that uses `lat/lng` we need to find a way to accept maybe both ways or some other solution

@alvelig  any ideas?

danielgindi:
It only works on one platform with lat/lng, which IS the breaking change... I’m suprised nobody noticed this before :)

The original PR was sent by me and I tend to use the shorthand, but anyone who passes in coordinates from that’s not from google sources might get an error there.

PR #2451, opened 2018-08-22
Remove the podspec dependency

SudoPlz:
No need for that dependency there.
People already use it in their projects if they need it, and it's only messing up our pod installation if we're not using cocoapods for react.


oblador:
It's needed if you import React Native with CocoaPods.

SudoPlz:
Hmm I assumed if you have it installed anyway it won't matter but I didn't test. Closing this.

PR #2455, opened 2018-08-23
Smart React util imports

SudoPlz:
This PR changes the way React components/utils/objects are imported so that imports work both with `react-native link` AND `cocoapods` React embedded projects.

### How did you test this PR?

Tested it on my react-native link project and it works.

### DIFF
--- a/lib/ios/AirGoogleMaps/AIRGMSMarker.h
+++ b/lib/ios/AirGoogleMaps/AIRGMSMarker.h
@@ -6,7 +6,11 @@
 //
 
 #import <GoogleMaps/GoogleMaps.h>
-#import <React/UIView+React.h>
+#if __has_include(<React/UIView+React.h>)
+    #import <React/UIView+React.h>
+#else
+    #import "UIView+React.h"
+#endif
 
 @class AIRGoogleMapMarker;
 
--- a/lib/ios/AirGoogleMaps/AIRGMSPolygon.h
+++ b/lib/ios/AirGoogleMaps/AIRGMSPolygon.h
@@ -6,7 +6,11 @@
 //
 
 #import <GoogleMaps/GoogleMaps.h>
-#import <React/UIView+React.h>
+#if __has_include(<React/UIView+React.h>)
+    #import <React/UIView+React.h>
+#else
+    #import "UIView+React.h"
+#endif
 
 @class AIRGoogleMapPolygon;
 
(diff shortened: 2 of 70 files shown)


rborn:
Have you seen this? https://github.com/react-community/react-native-maps/pull/2396

@wbyoung could you guys work together on that?

SudoPlz:
My PR will help #2396 too, it's not an either or situation.

wbyoung:
@SudoPlz I understand that you're trying to resolve compiler issues w/ this PR, but I'm not sure I really understand how it resolves the issue that you're facing that #2396 would not.

I think it'd be preferable to stick with the `React/*.h` imports. That's working just fine for me and others in #2396 and I believe is how pretty much every project will import things (i.e. `react-native-camera` imports w/ the [same style](https://github.com/react-native-community/react-native-camera/blob/1c9e83dc943bb9a6dde5b90ba73b643b98061c13/ios/RN/RNCamera.h#L2)).

When used via `react-native link`, the project will be included as a sub-project of a main React Native app. It's been a while since I've worked in depth with Xcode, but as I understand it, this also allows Xcode to determine how to resolve `React/*.h`.

Can you try pulling and using #2396 in your project as it should resolve the issue that you're trying to address in this PR as well.

If it doesn't, can you post an example GitHub project that includes the proper `package.json` that fails to compile properly with `npm run ios`?

SudoPlz:
@wbyoung you're right, I did pull #2396 here https://github.com/SudoPlz/react-native-maps/tree/bin and everything works fine now (I had to add a few missing `#ifdef HAVE_GOOGLE_MAPS` clauses there as well).

Keep in mind in this PR I also included my code too.
It never harms to have an extra if clause just in case something went wrong and the class is not there, but I haven't tested enough to see if it wouldn't work if those if/else statements were not there.

I'll close that PR since I don't have the time to investigate further atm, but hopefully that issue will help someone in the future.

wbyoung:
@SudoPlz yeah, it depends on what branch of the code base you're working off of. @chrise86 did the same thing off of a `stable` tag that I have [here](https://github.com/wbyoung/react-native-maps/pull/2/commits/058a942f5ec10a1873f704df1ca80ec53455cd9e) and shared it on https://github.com/wbyoung/react-native-maps/pull/2.

It's really just that all the Google Maps files need it (which I'm sure you already figured out, but the comment may be helpful for others who come here), but it depends on which branch you're including the changes. The open PR is based off of `master` which doesn't have that file.

Issue #2105, opened 2018-03-17
Custom map styles, custom markers and marker pin color not working on Android


prathammehta:
### Is this a bug report?

Yes

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

Yes

### Environment

react-native: 0.51.0
react: 16.0.0

### Steps to Reproduce

1. Create a react native app using react-native init.
2. Follow installation instructions to the dot.
3. Use Google custom style generator to generate a custom style.
4. Add custom style prop to Map component

### Expected Behavior

1. Custom map style would show up on Android and iOS

### Actual Behavior

- Custom style works on iOS
- Custom style does not work on Android.

![img_4365](https://user-images.githubusercontent.com/3429759/37558777-5f6106fe-2a3f-11e8-8219-51d00e62deab.PNG)
![screenshot_1521311365](https://user-images.githubusercontent.com/3429759/37558779-62e96f3c-2a3f-11e8-90a5-e52b11838f7e.png)

wbattel4607:
Can you please share what `react-native-maps` version and `play-services-maps` version you're using? The latest versions are `0.20.1` and `11.8.0` respectively.

prathammehta:
Hey, 
Sorry for the delayed response.

`react-native-maps` version is 0.20.1 
`Google play services` are set to 11.8.0

Still doesn't work.

wbattel4607:
Okay. Can you share the code for you MapView component?

Also, a new version of react-native-maps was just recently published. Try updating to v0.21.0. (Note that this will require Gradle version 3.0.0 or greater. See [here](https://github.com/react-community/react-native-maps/pull/2096#issuecomment-376860933).)

prathammehta:
This is the code for the `MapView` component:

`
       <MapView 
          style={{position:'absolute', height:'100%', width:'100%'}}
          initialRegion={{
            latitude: currentLocation.coords.latitude,
            longitude: currentLocation.coords.longitude,
            latitudeDelta: intitialMapDelta,
            longitudeDelta: intitialMapDelta,
          }}
          showsUserLocation={true}
          mapPadding={mapPadding}
          ref={c => this.mapView = c}
          customMapStyle={Static.MapStyle}
          provider={Platform.OS == 'ios'?PROVIDER_GOOGLE:null}
        >
`

Also just realised the following aren't working on Android either:

- Custom marker images
- Marker pin color

[...]

Issue #2082, opened 2018-03-11
Custom marker image very slow performance


minotaurrr:
### Is this a bug report?

Yes

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

Yes

### Environment

```
  "dependencies": {
    "expo": "^25.0.0",
    "react": "16.2.0",
    "react-native": "https://github.com/expo/react-native/archive/sdk-25.0.0.tar.gz"
  }
```
### Steps to Reproduce

Markers with default Google Maps marker image show up very quickly, no issues at all.

But when I use a custom image (size is 1kb) it struggles to render.

### Expected Behavior

For markers to show up quickly.

### Actual Behavior

Markers take too long to show up. For about 200+ markers, it takes around 10-20 seconds.

### Reproducible Demo

App.js
```
import React from 'react';
import { Dimensions } from 'react-native';
import { MapView } from 'expo';
import Marker from './Marker';
import haversine from 'haversine';

const { width, height } = Dimensions.get('window');
const ASPECT_RATIO = width / height;
const LATITUDE = -37.812365;
const LONGITUDE = 144.962338;
const LATITUDE_DELTA = 0.0222;
const LONGITUDE_DELTA = LATITUDE_DELTA * ASPECT_RATIO;

export default class App extends React.Component {
  constructor() {
    super();
    this.state = {
      region: {
        latitude: LATITUDE,
        longitude: LONGITUDE,
        latitudeDelta: LATITUDE_DELTA,
        longitudeDelta: LONGITUDE_DELTA
      },
      data: [],
      isLoaded: false,
      boundingBox: {
        westLng: 0,
        southLat: 0,
        eastLng: 0,
        northLat: 0
      }
    }
  }

  componentDidMount() {
    this._mounted = true;
  }

  componentWillUnmount() {
    this._mounted = false;
  }

  getData = async (region) => {
    const centre = {
      latitude: region.latitude,
      longitude: region.longitude
    }

    const screenVerticalBoundary = {
      latitude: this.state.boundingBox.northLat,
      longitude: region.longitude
    }

    const distance = haversine(centre, screenVerticalBoundary, {unit: 'meter'});
    const url = `http://test.com`;
    let data = await(await fetch(url)).json();
    this.setState({data: data, isLoaded: true});
  }

  onRegionChangeComplete = (region) => {
    let boundingBox = this.getBoundingBox(region);
    this.setState({boundingBox});
[...]

PR #2477, opened 2018-09-08
Implemented `tracksViewChanges` for Android

danielgindi:

No

This is related to the following issues:
#1870 #2105 #2048 #2082

### How did you test this PR?

In my own project, with a progressing pie animation inside the marker.
I've also made sure that tracking stops when the marker is removed from the map etc.

### Some background and explanation

iOS has the `tracksViewChanges` property which causes re-renders of the markers all the time, and slows down performance of the map greatly. 
Still, it is very handy especially given the async nature of React.
As many times the first images cached is of a blank component, it is very useful to use `tracksViewChanges=true` and then set it to `false` when `componentDidUpdate` is hit.

Another issue is loading images. Assuming the image loading issue will be fixed one day - we will still want `tracksViewChanges` and disabled that when `onLoadEnd` hits.
We could even start with it as a false value, and then turn it on and off when the loading is done.

Now some people use several hacks for the image issue on Android, like actual changes to Fesco code and usage of SVG images inside the Marker. But in those cases it is still slightly unstable and sometimes blank - but just because of the asyncronisity of react. This will also solve that.

I've set a default fps of 25, which seemed reasonable to me. But based on user feedback this could be easily adjusted.

### DIFF
--- a/docs/marker.md
+++ b/docs/marker.md
@@ -17,7 +17,7 @@
 | `identifier` | `String` |  | An identifier used to reference this marker at a later date.
 | `rotation` | `Float` | 0 | A float number indicating marker's rotation angle, in degrees.
 | `draggable` | `<null>` |  | This is a non-value based prop. Adding this allows the marker to be draggable (re-positioned).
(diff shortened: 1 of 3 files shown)


rborn:
@alvelig LGTM but this is android , could you 🐽

alvelig:
LGTM

PR #2479, opened 2018-09-09
add edgePadding to fitToSuppliedMarkers function

marcelkalveram:

No.

Adds the `edgePadding` option to the `fitToSuppliedMarkers` function which is supported by Google Maps on both, iOS and Android. (no support for Apple Maps as the [native function](https://developer.apple.com/documentation/mapkit/mkmapview/1452309-showannotations) doesn't support this option).

The feature has been requested in those two issue and I had another use case on a personal project where I had to use a workaround using `fitToCoordinates`:
https://github.com/react-community/react-native-maps/issues/780
https://github.com/react-community/react-native-maps/issues/876

Another issue is requesting it for the function `fitToElements ` which should be an easy addition if this PR gets accepted:
https://github.com/react-community/react-native-maps/issues/922

### How did you test this PR?

Tested on real devices: Android Pixel and iPhone 7.
Tested on iOS simulator: iPhone 6


rborn:
@marcelkalveram thank you ❤️

LGTM @alvelig 🐽

PR #2487, opened 2018-09-12
Performance improvements for tracksViewChanges

danielgindi:

No

Performance improvements for #2477 #1870 

### How did you test this PR?

Full project

### DIFF
--- a/docs/marker.md
+++ b/docs/marker.md
@@ -44,6 +44,7 @@ To access event data, you will need to use `e.nativeEvent`. For example, `onPres
 | `showCallout` |  | Shows the callout for this marker
 | `hideCallout` |  | Hides the callout for this marker
 | `animateMarkerToCoordinate` | `coordinate: LatLng, duration: number` | Animates marker movement. **Note**: Android only
+| `redraw` |  | Causes a redraw of the marker. Useful when there are updates to the marker and `tracksViewChanges` comes with a cost that is too high.
 
 
 
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Canvas;
+import android.graphics.Color;
 import android.graphics.drawable.Animatable;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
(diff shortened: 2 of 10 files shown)


danielgindi:
@rborn some improvements related to the latest issue with tracksViewChanges and custom Image tags in markers. Should be merged in before a release to avoid a too heavy toll on users with `tracksViewChanges` as default.

rborn:
@alvelig could you have a look 🐽 I'm heading to ReactAlicante so I won't be able to do anything until Monday

alvelig:
@danielgindi Do I get it right, that it comes as a replacement for `tracksViewChanges`? I think that we should enable this behavior by default.

tmclemons:
This redraw update, will this also help with the memory leaks issues related to maps rerendering?

alvelig:
Excuse me, but why do we need `tracksViewChanges`? I missed that one

tmclemons:
@danielgindi awesome i cant wait till this update is pushed to the repo

danielgindi:
@alvelig `tracksViewChanges` as the name implies - tracks the view's changes and updates the markers accordingly. This basically means that you could use animations in the markers.

alvelig:
So by default `tracksViewChanges` is not required. That's why you introduced `redraw`, if it is required to redraw a marker once. Now I see. By default this shouldn't be required, right?

danielgindi:
Well it depends :)
I have some custom markers that have animations (kind of progress bars) whete `tracksViewChanges` is required, and some where a simple `redraw` call would suffice.

The thing is that even with iOS until now the way to ensure that a custom marker will be drawn is to keep the default `tracksViewChanges: true`, and set it ti false on `componentDidUpdate` for performance. This is because RN rendering is async and you first take a frame of an empty view.

So by default it is the correct mechanism, but then the user should disable it on the correct event like componentDidUpdate or Image’s onLoad or other custom events. 

It’s even more complex than that, as you might have a custom marker that has an image AND other views.
So first frame takes an empty image. After first render happened, the `trackViewChanges` will take an image of the actual view, and after loading the image it will update to the marker with the image loaded- only then you want to disable tracking.


So my main recommendation is adding docs and more docs :)

danielgindi:
We could also wrap the marker component with another, that will have built in logic for `componentDidUpdate` for the simple custom markers - but that would be a breaking change.

alvelig:
@danielgindi why would it be a breaking change?

danielgindi:
As people expect it now to be `true` by default (as it has always been in iOS), and be disable when they set it to disabled.

alvelig:
Great, thanks for the explanations. I agree!

slorber:
hey @danielgindi wondering if you have any idea how tracksViewChange could play well with native driver animations?

See https://github.com/react-native-community/react-native-maps/pull/1705#issuecomment-446604282

PR #2501, opened 2018-09-21
[iOS] Fixed issue that app will crash after trigger  Marker onDragEnd

guoliang1206:
No

What issue is this PR fixing?
When drag the Marker on map, sometimes app will crash when onDragEnd.

How did you test this PR?
Tested on iOS simulator with iOS11.4, iOS 12.0
iPhone6 12.0


rborn:
@guoliang1206 thank you ❤️

LGTM @alvelig 🐽

PR #2524, opened 2018-10-01
fix fitToSuppliedMarkers function

marcelkalveram:
Fixes https://github.com/react-community/react-native-maps/issues/2513

https://github.com/react-community/react-native-maps/pull/2520

seems to fix the same issue

https://github.com/react-community/react-native-maps/issues/2513

### How did you test this PR?

iOS simulator. Builds were failing without this fix.


marcelkalveram:
btw: the bug got originally introduced here: https://github.com/react-community/react-native-maps/pull/2479

rborn:
@marcelkalveram do we need to adjust docs too?

marcelkalveram:
@rborn this doesn't work with Apple Maps (only Google, both Android and iOS), as mentioned in my previous PR. Not sure if this is something that should be mentioned in the docs? At least, after this PR, it shouldn't fail anymore with Apple Maps.

rborn:
We should mention that it only works with google maps or we'll get a lot of issues reported (it will be, trust me). 
So if the docs need to be amended please do so and I'll merge asap.
Thank you ❤️

rborn:
@marcelkalveram I'll merge this

Issue #2447, opened 2018-08-22
There was an error React Native.(React-Native-Maps)


Yokoi-K:
### I would like to draw a map using react-native-maps with react native. I succeeded at ios. However, Android gives an error.

The error contents are 「**The specified child already has a parent.You must call removeView() on the parent first.**」

I tried it a lot, but it will not work.

Does anyone fall into the same situation as I am?

Please let me know the solution.

ChristopheHochP:
Hi @Yokoi-K, I am running into the same issue as you described. Did you found a solution to your problem?

alvelig:
@ShrutiGarg019 remove all your in-marker components and try. Then go adding one by one and find which element. Let me know when found.

PR #2545, opened 2018-10-11
Fix for “The specified child already has a parent”

shanereid:

No.

I haven't created an issue for it, but seems to be the same issue found here:
https://github.com/react-community/react-native-maps/issues/2447

Error it fixes is the following, which for me personally occurred when re-rendering with the map with a partially different set of Markers.
For performance reasons in my project on render only markers that are within the specified region are returned with render, with the map re-rendering markers using `onRegionChangeComplete`. With small changes it is generally fine however I found when animating a region change where a lot of markers change this crash would occur.
This simple change in the java code fixed the issue for me.
screenshot_20181011-105356_dusk

(please link the issue here)

### How did you test this PR?

I only have this issue on Android, and I could consistently replicate the crash on a suite of android test devices. After the fix it was easy to confirm that it fixed the issue for me. Also the code snippet is pretty harmless and contained, checks for null and should not introduce any crashes.

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
@@ -542,6 +542,13 @@ public class AirMapView extends MapView implements GoogleMap.InfoWindowAdapter,
       int visibility = annotation.getVisibility();
       annotation.setVisibility(INVISIBLE);
 
+      // Remove from a view group if already present, prevent "specified child
+      // already had a parent" error.
+      ViewGroup annotationParent = (ViewGroup)annotation.getParent();
+      if (annotationParent != null) {
+        annotationParent.removeView(annotation);
+      }
+
       // Add to the parent group
       attacherGroup.addView(annotation);
 


rborn:
LGTM @alvelig 🐽

craigcoles:
Massive props @shanereid - and thanks for the speedy merge 🎉

ShrutiGarg019:
> 
> ### Does any other open PR do the same thing?
> 
> No.
> ### What issue is this PR fixing?
> 
> I haven't created an issue for it, but seems to be the same issue found here:
> #2447
> 
> Error it fixes is the following, which for me personally occurred when re-rendering with the map with a partially different set of Markers.
> For performance reasons in my project on render only markers that are within the specified region are returned with render, with the map re-rendering markers using `onRegionChangeComplete`. With small changes it is generally fine however I found when animating a region change where a lot of markers change this crash would occur.
> This simple change in the java code fixed the issue for me.
> screenshot_20181011-105356_dusk
> 
> (please link the issue here)
> ### How did you test this PR?
> 
> I only have this issue on Android, and I could consistently replicate the crash on a suite of android test devices. After the fix it was easy to confirm that it fixed the issue for me. Also the code snippet is pretty harmless and contained, checks for null and should not introduce any crashes.

I am still facing same issue after this fix too.Any other solution for this??

Issue #733, opened 2016-10-26
image marker size 


yellowbang:
I am using a customize image as a marker. However my image size is 400 x 400. How to make it smaller so that it fits the map?

Arturszott:
I would try the following:

```
<Image
    source={require('./marker.png')}
    style={{ width: 40, height: 40 }}
 />
```

OtacilioN:
I am trying to use: 
<MapView.Marker>
            <Image source={require('src/assets/marker.png')} style={{ width: 40, height: 40 }} />
            coordinate={{
              latitude: this.state.latitude,
              longitude: this.state.longitude
            }}
          </MapView.Marker>
And got this error:

<img width="375" alt="screen shot 2018-01-15 at 10 57 34" src="https://user-images.githubusercontent.com/10578275/34945706-f110fd44-f9e2-11e7-8a3c-838bd9cae667.png">

rborn:
@OtacilioN you are uisng `coordinate` as child of the marker instead of a prop.

OtacilioN:
Thanks @rborn solved now, follow the example to make it clear:

<MapView.Marker
            coordinate={{
              latitude: this.state.latitude,
              longitude: this.state.longitude
            }}>
            <Image source={require('src/assets/marker.png')} style={{ width: 40, height: 40 }} />
          </MapView.Marker>

OtacilioN:
Now is working in the simulator but is showing in the build version =/ 

1.

rborn:
What's wrong? without a description of what's going on we can't help :|

OtacilioN:
Okay was build problems, working perfect now, thank you so much for your patience @rborn

rborn:
@OtacilioN 🤗

douglasjunior:
Dont work for Android?

douglasjunior:
For Android only works with `Marker.Animated`:

```xml
      <Marker.Animated
                identifier={`VagaMarker:${vaga.id}`}
                key={vaga.id}
                coordinate={new AnimatedRegion({
                    latitude: coord[0],
                    longitude: coord[1],
                })}
            >
                <Image
                    source={imageSource}
                    style={{ height: 50, width: 50 }}
                    resizeMode="contain"
                />
            </Marker.Animated>
```

But now work with normal `Marker`:

```xml
      <Marker
                identifier={`VagaMarker:${vaga.id}`}
                key={vaga.id}
                coordinate={{
                    latitude: coord[0],
[...]

PR #2547, opened 2018-10-12
marker icon is scalable now,needs width and height in marker props

shashi080195:
no

#733 

### How did you test this PR?
i have tested it in my application on mobiles of different api versions
- Which platform (eg. Android/iOS) & Maps API (eg. Google/Apple) does this change affect, if any?Andorid
- Did you test this on a real device, or in a simulator?
I have tested it a real device
- Are there any platforms you were not able to test?
no

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
@@ -416,12 +416,17 @@ public class AirMapMarker extends AirMapFeature {
       } else {
         return BitmapDescriptorFactory.fromBitmap(createDrawable());
       }
-    } else if (iconBitmapDescriptor != null) {
-      // use local image as a marker
-      return iconBitmapDescriptor;
-    } else {
-      // render the default marker pin
-      return BitmapDescriptorFactory.defaultMarker(this.markerHue);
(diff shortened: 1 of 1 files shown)


chakrihacker:
Any update on this??

oliverdolgener:
Can someome please review this. It would be really helpful to have reliable custom icon sizes for markers on Android and iOS.

dakairus:
any update on this, it would solve a issue I currently got on small screen devices

stockrel:
+1

alvelig:
LGTM. Though not sure about the performance.
And of course missing iOS version.
@rborn 🐽

rborn:
@alvelig I'm worried about performance too.
On the other side we should not merge non cross-platform features unless they are specific to one platform only.

shashi080195:
will update with the ios version also

chakrihacker:
We are using with 10 - 20 markers, I haven't seen any performance regression.

andreencar:
@shashi080195 any news about the ios fix?

jramalho:
having some issues with Marker icon on iPhone, to have the same size on android you need to / by the DPI  ratio on the original image.

rborn:
@chakrihacker 20 markers are not much, what about 200 or more?

rborn:

PR #2548, opened 2018-10-12
Move babel-plugin-module-resolver and babel-preset-react-native to devDependencies

jmheik:
No.

Babel-plugin-module-resolver and babel-preset-react-native are included as runtime dependencies. As far as I can see they are only referred from .babelrc which is also in .npmignore, so they are not used by end users who install react-native-maps as npm module. However, these deps have quite a lot of transitive dependencies. This PR moves them to devDependencies to spare end users from downloading them and bloating the node_modules.

### How did you test this PR?

I have searched the code base, and not found any other use for these two deps than .babelrc, which is excluded from npm published version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,9 @@
   },
   "devDependencies": {
     "babel-eslint": "^6.1.2",
+    "babel-plugin-module-resolver": "^2.3.0",
     "babel-preset-airbnb": "^1.1.1",
+    "babel-preset-react-native": "1.9.0",
     "eslint": "^3.3.1",
     "eslint-config-airbnb": "^10.0.1",
     "eslint-plugin-import": "^1.14.0",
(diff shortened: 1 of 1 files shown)


jmheik:
What do you think @rborn @alvelig ? This would help cleaning up quite some transitive deps from people's projects as can be seen here https://npm.anvaka.com/#/view/2d/react-native-maps

alvelig:
I like your idea, @jmheik. @rborn, what do you think?

todorone:
These type of dependencies are for sure not suitable for `dependencies` section.
@rborn Please look if this PR can be merged.

rborn:
@todorone done, can you add it to the 0.22.1 changelog? 🤗

rborn:
@jmheik ❤️

Issue #2362, opened 2018-07-04
can't show google maps on iOS


shoki2shoki:
I couldn't display google maps on iOS, only can show apple map even though I add some codes to ios/YOUR_PROJECT_NAME/AppDelegate.m according to installation guide below.
```
+ @import GoogleMaps; //add this line if you want to use Google Maps

@implementation AppDelegate
...

- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
+  [GMSServices provideAPIKey:@"_YOUR_API_KEY_"]; // add this line using the api key obtained from Google Console
...
```
↑I put my google api key instead of "_YOUR_API_KEY_"

https://github.com/react-community/react-native-maps/blob/master/docs/installation.md

Did someone have same problem?

declandowling:
You probably haven't specified Google maps as the provider. 

`import MapView, { PROVIDER_GOOGLE } from 'react-native-maps'`
`<MapView provider={PROVIDER_GOOGLE}>`

The docs aren't very clear on this, I had trouble with it myself until I read down to **[Customizing the map style](https://github.com/react-community/react-native-maps#customizing-the-map-style)**

shoki2shoki:
Thank you for your comment. I resolved it.

shoki2shoki:
I'm closing this issue because I found out the way to display google maps.

SiSa68:
Please add @declandowling 's comment to [Enabling Google Maps](https://github.com/react-native-maps/react-native-maps/blob/master/docs/installation.md#enabling-google-maps) section of the documentation!
It took me a few hours to find this!!

PR #2550, opened 2018-10-15
Specify how to use Google Maps

platonish:
No.

(please answer here)

Issue #2362 

Installation docs do not mention that `{PROVIDER_GOOGLE}` needs to be added in order to make Google Maps work on iOS. This PR adds the required lines to make Google Maps work on iOS.

### How did you test this PR?

This is an update to the documentation only.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -283,7 +283,7 @@ example is below:
 
 
 ```jsx
-import MapView from 'react-native-maps';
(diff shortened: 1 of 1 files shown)


rborn:
@platonish looks good, thanks a lot, could you please add some lines about this in https://github.com/react-community/react-native-maps/blob/master/docs/installation.md#enabling-google-maps-on-ios section too? 
Thank you!

PR #2555, opened 2018-10-18
r2507: remove marker: Attempt to invoke virtual method 'void com.goog…

salah-ghanim:
…le.android.gms.maps.model.setIcon(com.google.android.gms.maps.model.BitmapDescription)' on a null object reference #: remove marker: Attempt to invoke virtual method 'void com.google.android.gms.maps.model.setIcon(com.google.android.gms.maps.model.BitmapDescription)' on a null object reference

  * null check is enough

No

https://github.com/react-community/react-native-maps/issues/2507

### How did you test this PR?

Android
and tested on real device

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapMarker.java
@@ -285,8 +285,9 @@ public class AirMapMarker extends AirMapFeature {
       // No more updates for this, as it's a simple icon
       hasViewChanges = false;
     }
-
-    marker.setIcon(getIcon());
+    if (marker != null) {
+      marker.setIcon(getIcon());
+    }
   }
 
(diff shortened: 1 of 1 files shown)


hooba91:
Instead of the Eroor Attempt to invoke virtual method 'void com.google.android.gms.maps.model.setIcon(com.google.android.gms.maps.model.BitmapDescription)' on a null object reference

I get now the following Error:
 
The specified child already has a parent. you must call removeView() on the child's parent first.

screenshot 2018-10-19 at 15 34 13

jerryjj:
This fixed the problem for me at least. Is there a timeline when this could get merged and published?
Thanks!

todorone:
@salah-ghanim Can You please look please at report of problem that appeared as a result of Your commit. Do You have idea if it could be fixed?
https://github.com/react-community/react-native-maps/issues/2507#issuecomment-433039328

Issue #2454, opened 2018-08-23
Module not found 


markib:
### Is this a bug report?

(write your answer here)

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

yes

### Environment
{
    "private": true,
    "scripts": {
        "dev": "npm run development",
        "development": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
        "watch": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --mode development --watch --progress --hide-modules --config=./webpack.config.js",
        "watch-poll": "npm run watch -- --watch-poll",
        "web": "webpack --mode development  --config ./webpack.config.js  --inline --hot --colors --content-base public/index.php --history-api-fallback",
        "hot": "cross-env NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
        "prod": "npm run production",
        "production": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --no-progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js"
    },
    "devDependencies": {
        "@babel/plugin-proposal-object-rest-spread": "^7.0.0-rc.1",
        "axios": "^0.18.0",
        "babel-loader": "^7.1.5",
        "babel-plugin-react-native-web": "^0.8.9",
        "babel-plugin-transform-decorators-legacy": "^1.3.5",
        "babel-plugin-transform-imports": "^1.5.0",
        "babel-plugin-transform-react-jsx-source": "^6.22.0",
        "babel-plugin-transform-runtime": "^6.23.0",
        "babel-preset-env": "^1.7.0",
        "babel-preset-react": "^6.24.1",
        "babel-preset-react-native": "^4.0.0",
        "bootstrap-sass": "^3.3.7",
        "cross-env": "^5.1",
        "css-loader": "^1.0.0",
        "file-loader": "^1.1.11",
        "jquery": "^3.2",
        "laravel-mix": "^2.1.11",
        "lodash": "^4.17.4",
        "react": "^16.4.1",
        "react-dom": "^16.4.1",
        "redux-logger": "^3.0.6",
        "style-loader": "^0.22.1",
        "webpack": "^4.16.5",
[...]

Issue #2453, opened 2018-08-23
Module not found


markib:
ERROR in ./node_modules/react-native-maps/lib/components/MapOverlay.js
Module not found: Error: Can't resolve 'react-native/Libraries/Image/resolveAssetSource'

PR #2576, opened 2018-11-08
fix import error on compilation

digicontributer:

Not that I know of.

#2454
#2453

PR fixes compatibility issue with 'react-native-web' and compiling.
### How did you test this PR?

Compiled successfully on Windows & OSX for Android & IOS.


rborn:
@alvelig 🐽

PR #2589, opened 2018-11-17
Tweaked AirMapUrlTile to allow for 512 tile size

albullington:

No

https://github.com/react-community/react-native-maps/issues/2567

### How did you test this PR?

1. Took snapshots of what the map looked like before this fix, outlined in the issue above.
2. Tested that new map displays tiles correctly on an iPhone SE device running iOS 12. 

This change affects iOS/Apple Maps only.

### DIFF
--- a/lib/ios/AirMaps/AIRMapUrlTile.m
+++ b/lib/ios/AirMaps/AIRMapUrlTile.m
@@ -60,6 +60,7 @@
     if (self.maximumZ) {
         self.tileOverlay.maximumZ = self.maximumZ;
     }
+    self.tileOverlay.tileSize = CGSizeMake(512, 512);
     self.renderer = [[MKTileOverlayRenderer alloc] initWithTileOverlay:self.tileOverlay];
 }
 


lib/ios/AirMaps/AIRMapUrlTile.m
```diff
@@ -60,6 +60,7 @@ - (void) createTileOverlayAndRendererIfPossible
     if (self.maximumZ) {
         self.tileOverlay.maximumZ = self.maximumZ;
     }
+    self.tileOverlay.tileSize = CGSizeMake(512, 512);
```
rborn: Can't we provide the tileSize in the props somehow so we don't need to hardcoded to 512?
And we can default to 512 in case is not defined?

albullington: Definitely open to suggestions on better ways to do this... I'm not familiar with Swift, so this was the best workaround I came up with.

rborn: maybe "steal" something from here? https://github.com/react-community/react-native-maps/blob/72b0f77e31d18959e91acb036d5f91ab46e11ef2/lib/ios/AirMaps/AIRMapLocalTile.m#L26 (I'm not familiar with swift either 😄  )

fozhao: It will be perfect if it's not hardcoded. I am dealing with some tiles with size of (128, 128). So, this patch won't be good to me unless I fork and hardcode it with the expected size. Will a tileSize props be welcomed? I can try to make a PR of it though I am not an expert of swift either.

fozhao: Made a PR of tile size for iOS. Please take a look. thanks. 
https://github.com/react-native-community/react-native-maps/pull/2627

christopherdro:
Closing for https://github.com/react-native-community/react-native-maps/pull/2627

PR #2598, opened 2018-11-22
[Android] Fix app crash in Android if building found but cannot getActiveLevelIndex

luciferes:
No, it doesn't.

Fix Android app crash because building.getActiveLevelIndex() return index that out of bounds of 
 the building when map found a building but showsIndoors is false. Exception error message as below

`Fatal Exception: java.lang.ArrayIndexOutOfBoundsException
length=8; index=-1
java.util.ArrayList.get (ArrayList.java:439)
com.airbnb.android.react.maps.AirMapView.onIndoorLevelActivated (AirMapView.java:1112)`

### How did you test this PR?

Tested on Android emulator (Nexus 5), real Android device (HTC U11),  iOS Simulator (iPhone X), real iOS device (iPhone 6S) by running example.


miracle2k:
👍

Issue #2616, opened 2018-12-07
Google Maps on iOS is very slow


nicomontanari98:
### Is this a bug report?

Yes

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

Yes

### Environment

 react: 16.6.3 
 react-native: 0.57.7 
 react-native-maps: 0.21.0
 Target Platform: eg: iOS - 10 Simulator/iPhone XS/iPhone 6s
 Map: GoogleMaps

### Expected Behavior

Fluid map end 60 fp/s, like Android

### Actual Behavior

Very slow map and 9 or 10 fp/s, also if i remove all markers

<img width="235" alt="schermata 2018-12-07 alle 11 58 09" src="https://user-images.githubusercontent.com/15747776/49643975-6aee6400-fa17-11e8-99c2-d575980fbbe1.png">

### Reproducible Demo

Use google maps on iOS

SvetlozarKalchev:
It's unusable on iOS for me too. I am using **react-native 0.55.2**. I've tried using **react-native-maps 0.21.0** and **0.22.1**, but performance is atrocious with both versions. There's extremely high CPU usage when rendering even few markers (10-20). Even if I don't render any markers, the CPU usage goes to 100% when dragging the map and the frame rate drops extremely. 

This package is currently unusable if you want to show a Google map.

iOS 12, iPhone 7+ and various simulators.

luispeerez:
I had an issue yesterday with a customer's iphone 7 plus, I have tested on iphone xr, 8, and none of those had problems with the maps, I am going to test if tracksViewChanges prop on Marker component fix the issue

bruteforcecat:
interesting. the performance in iphone 5S, SE, 6 are faster. UI fps stay close to 60 when dragging or animating

alexandrius:
This was best what I could come up with.

```
shouldComponentUpdate(nextProps) {
    if (this.props.activities !== nextProps.activities) {
        this.setState({ tracksViewChanges: true }, () => {
            setTimeout(() => {
                this.setState({ tracksViewChanges: false });
            }, 2000)
        })
    }
    return true;
}
```

setting tracksViewChanges: false in componentDidUpdate didn't work for me

victormartingarcia:
We are experiencing this performance issues with Google Maps on iOS as well.

@alexandrius does this solution have any side effect? Are all markers drawn properly?

alexandrius:
[...]

Issue #2575, opened 2018-11-08
Custom markers slow performance with reproduceble demo


compojoom:
I was experiencing high CPU usage and lags on the map with what I consider a lot of markers - 400. To my surprise however the map was fast with 300 markers. So, it wasn't really making sense. 
After debugging this for several hours I managed to come with a reproducible demo of the issue.

https://github.com/compojoom/react-native-maps/tree/high_cpu

launch the examples and switch to google maps as provider and open the "custom markers" example. 

in xCode one should see
<img width="271" alt="image" src="https://user-images.githubusercontent.com/693770/48213168-7baaac00-e37d-11e8-9199-d12b7c880ace.png">

If you go into
example/examples/CustomMarkers.js
and remove the `dsafasdf` on line 81 and refresh the example you will see that the CPU usage goes to 1-2%
<img width="269" alt="image" src="https://user-images.githubusercontent.com/693770/48213327-d2b08100-e37d-11e8-8c65-735b98720285.png">

if one changes the width of the custom marker to 100 px for example, then we get again CPU usage of 100%

I've tested this currently only on iOS and I'm not sure if the issue apply to Android. 

### Environment
 "react": "16.5.0",
"react-mixin": "^2.0.2",
"react-native": "0.57.1",
react-native-maps: > 0.21.0
Target Platform: eg: iOS - 12.1 Simulator/iPhone 6

### Steps to Reproduce
Use the reproducible demo. The issue is hard to reproduce as one needs a lot of markers and the "correct" styling...

### Expected Behavior
I would expect CPU to go down once the markers are rendered

### Actual Behavior
CPU stays at 100%

### Reproducible Demo
https://github.com/compojoom/react-native-maps/tree/high_cpu

rborn:
@compojoom nice research, could you please update the issue to follow the template? 
Thanks

compojoom:
updated

cklinx:
Try to set `tracksViewChanges={false}` in your markers

compojoom:
@cklinx - this bug report has nothing to do with tracksViewChanges. If you look at the example code you'll see that tracksViewChanges is already set to false: https://github.com/compojoom/react-native-maps/blob/d60402e2c62f26f3be91023381bdb0cafc9c8318/example/examples/CustomMarkers.js#L73

idumachika:
i have issue styling my map i want to implement border raduis on my map ,if
i add border radius  it does work

tomtom94:
[...]

PR #2650, opened 2019-01-07
Expose GoogleMaps iOS SDK icon property to improve Marker rendering performance

victormartingarcia:

This PR is based on https://github.com/react-native-community/react-native-maps/pull/1491 changeset

It hasn't been merged (presumably unsolved conflicts), but I see Marker **tracksViewChange** property is already exposed on master branch, so it makes sense to expose Marker **icon** property as well

Rendering Markers on iOS using Google Maps have some performance issues (see #2616, #1031, #2575 ..)
Some people have solved that with some extra logic tweaking **tracksViewChange** property (see https://github.com/react-native-community/react-native-maps/issues/1031#issuecomment-378881118 ), but [as Google explains in the SDK docs](https://developers.google.com/maps/documentation/ios-sdk/marker#use_the_markers_icon_property), for static Markers is much better to use icon property.

Using this icon property solves all iOS performance issues, making the map much smoother :)

### How did you test this PR?

This fix has been tested on iPhone simulator and real devices (iPhone 6 and iPhone X) with extremely good results. 

On Android, `icon` property works like `image` property, as Android SDK does not differentiate between static image Markers and View-based ones

### Example
```
<Marker
    coordinate={{
        latitude: this.state.myLatitude,
        longitude: this.state.myLongitude
    }}
    icon={require('../Images/myIcon.png')}
/>
```


rborn:
Thanks for the PR ❤️

I wonder of we could unify the `icon/image` props between the platforms? so we use one name only? Or on iOS we can use both `image` and `icon` ?

@alvelig thoughts ? and 🐽
LGTM

victormartingarcia:
@rborn on iOS you can use both `image` and `icon` for different purposes (icon only works for static images)

Let me check if I can make it compatible with android, for not having to differentiate it on Marker props

victormartingarcia:
I managed to support `icon` property on android as well, so no need to differentiate platform on Marker definition ;)

Using `icon` on android is exactly the same as using `image`, as android Google Maps SDK does not differentiate between static image Markers and View-based ones

rborn:
@victormartingarcia looks good, thanks.
One question though: is there anything we need to do to handle `@2x/@3x` images? How do the marker decide the size of the icon? (maybe we should add some docs saying what's the ideal marker icon size if google specifies something about it?)

victormartingarcia:
> One question though: is there anything we need to do to handle `@2x/@3x` images? 

I've tested and `@2x/@3x` images are handled automatically 👍 

> How do the marker decide the size of the icon? 

Just like with `image` property, the Marker takes picture size by default. 

> (maybe we should add some docs saying what's the ideal marker icon size if google specifies something about it?)

Haven't found anything about Marker size in Google Maps SDKs docs...

rborn:
@victormartingarcia let's wait a little for @alvelig to reply (if) and then we'll merge 🎉

christopherdro:
Related to https://github.com/react-native-community/react-native-maps/pull/1491

christopherdro:
I'm not sure why this wasn't caught in the tests that ran on this PR but I am following up with the following https://github.com/react-native-community/react-native-maps/pull/2651

PR #2653, opened 2019-01-10
Defaults to the map services version instead of play services

vincentvella:
As of now the googlePlayServicesMapsVersion variable defaults to "DEFAULT_GOOGLE_PLAY_SERVICES_VERSION" which is an unpublished version of maps - 16. 0.1, it needs to default to "DEFAULT_GOOGLE_PLAY_SERVICES_MAPS_VERSION" instead which is 16.0.0

No, no other PR's have been submitted at the time of writing.

https://github.com/react-native-community/react-native-maps/issues/2652

### How did you test this PR?

Ran the sync before and after fixing with it symlinked, my PR fixes the issue with downloading the correct play services version.


rborn:
@vincentvella thnx for the PR, is this working with the current git version (that's not yet published on npm)

vincentvella:
Yes, confirmed this change worked from the current master with my added change. Fixed our CI process.

PR #2692, opened 2019-02-11
Extend onPanDrag()

jochem725:
https://github.com/react-native-community/react-native-maps/pull/2314 also solves an issue with pan dragging.

In our app we wanted to extend the behavior of onPandrag to include zoom, rotate and pinch and fix an issue with the gesture recognizer on iOS.

For now I've included them in the onPandrag event, ideally this should become an onZoom and onPinch. For both Android and iOS I wonder if these gesture recognizers are the correct approach, if so I will implement onZoom and onPinch.

### How did you test this PR?

- [ ] Create a map component with onPandrag
- [ ] Pinch, Zoom and Drag, onPandrag should fire
- [ ] onPanDrag should now also fire when scrollEnabled=true

#### **Which platform (eg. Android/iOS) & Maps API (eg. Google/Apple) does this change affect, if any?**
iOS (Apple Maps) + Android
#### **Did you test this on a real device, or in a simulator?**
Both

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
@@ -167,6 +167,14 @@ public class AirMapView extends MapView implements GoogleMap.InfoWindowAdapter,
             }
             return false;
           }
+
+          @Override
+          public boolean onDoubleTap(MotionEvent e) {
+            if (handlePanDrag) {
+              onPanDrag(e);
+            }
+            return false;
+          }
         });
 
     this.addOnLayoutChangeListener(new OnLayoutChangeListener() {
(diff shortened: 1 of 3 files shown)


chakrihacker:
Please merge this

salah-ghanim:
@jochem725 onPanDrag is documented as follows
> Callback that is called when user makes a "drag" somewhere on the map

While I like the idea of having more events exposed through the react api I don't see that we should expose them via this event callback.

And this API is sending events too frequently on Android while on iOS it's sent once per gesture (which makes sense in what you're trying to achieve: onZoom, onPinch ...)

Please refactor into new APIs if you want this to be merged.

michalchudziak:
@jochem725 Any ETA on this? I could jump in and assist you with migration to new events if you don't have time.

jochem725:
@michalchudziak On a short notice I don’t have time to dive into this unfortunately. If you could jump in that would be great :)

I can aid in reviewing or if you there are some clear tasks I can work on I’m willing to help.

nabati:
Does on double tap trigger the events for this on iOS? I'm guessing it wont.

rborn:

PR #2694, opened 2019-02-13
Ability to overzoom on IOS + Fix for blurry tiles in Android

pohnean:
No

Issue 1:
If you set the maximumZ to be greater than the available zoom level (if you use custom raster tiles), you will see a different behavior on android compared to IOS. 

On android, the map renderer will 'overzoom', scaling up the tiles of the maximum zoom level. But on IOS, the tiles simply disappear.

Issue 2:
Custom raster tiles appear to be blurrier compared to iOS devices on retina screens (DPI = 2). This patch also fixes this by dynamically creating higher resolution tiles using the tiles from a higher zoom level. Original solution is found here: https://stackoverflow.com/questions/23806348/blurred-custom-tiles-on-android-maps-v2/36696291

### How did you test this PR?
Tested on both iphone and android device.


sinodko:
whats up with this PR? :) it looks like great feature

pedrolopes10:
@pohnean thanks for this PR that saved me with the blurry tiles in Android!

pohnean:
Glad it helped you :)

cosmarc:
@pohnean Can you please fix the trailing spaces where travis complains? It looks like a nice feature and it would be sad to remain here.

pohnean:
@cosmarc travis issues fixed.

todorone:
@rborn @christopherdro Could you please look at it?

ignitebhdeveloper:
having the same issue, android blurry tiles. will this be merged anytime soon?

hussainahmad:
when this PR will be merged ?

todorone:
@christopherdro @rborn

hussainahmad:
@pohnean Thanks mate , You make my day.  hope it will merge soon

rborn:
LGTM (however my native fu is not that good) @christopherdro 🐽

rajeshkp:
Appreciate if this PR can be merged

T-Spoon:
Having a quick play with the Android code - there are quite a few issues (although I haven't thought of any other workaround to this unfortunate ["feature"](https://issuetracker.google.com/issues/35822292) of the Google Maps SDK).

1. This consumes a lot more more memory, as it downloads 4 tiles and stitches them together in-memory (even without React Native on top, this probably isn't workable)
2. It's slow.  It synchronously downloads 4 tiles one after the other in order to display a single tile. 

It would be possible to setup an executer to download these in parallel and solve issue 2. But I still don't think this is an acceptable general purpose solution for all apps.

Whether or not you want to use this solution should probably be made on a per-app basis - rather than enforcing it as the default (especially if the app is supposed to run on older devices)

wgt322:
I'm new to github and dont even know if this is place for this but how can i use your "version" on my app ? Im stuck with blurry tiles on android and been search for 2 days for a solution. Pretty much sure you have it.

SaeedZhiany:
a

PR #2707, opened 2019-02-24
Add clearer notes to setMapBoundaries method

chris-poe:
When using the setMapBoundaries method, it is not immediately clear why the user is able to scroll past the specified boundaries. Noting that the method uses the maps center coordinates would help clarify.

No, there are no other pull requests open on this issue.

Clarifies the use of the setMapBoundaries method.

### How did you test this PR?

No testing needed for markdown edits.

### DIFF
--- a/docs/mapview.md
+++ b/docs/mapview.md
@@ -82,7 +82,7 @@ To access event data, you will need to use `e.nativeEvent`. For example, `onPres
 | `animateToBearing` | `bearing: Number`, `duration: Number` | Deprecated. Use `animateCamera` instead.
 | `animateToViewingAngle` | `angle: Number`, `duration: Number` | Deprecated. Use `animateCamera` instead.
 | `getMapBoundaries` | | `Promise<{northEast: LatLng, southWest: LatLng}>`
-| `setMapBoundaries` | `northEast: LatLng`, `southWest: LatLng` | `GoogleMaps only`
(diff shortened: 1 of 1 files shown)


EskelCz:
@chris-poe I'm trying to use the function and I must concur, it is not clear. Unfortunately not even with your improvement. Are the boundary coordinates somehow relative to something? Doesn't make any sense to me.
I've put in absolute coordinates of what I want to be available to the user and it seems like it simple doesn't work.

chris-poe:
@EskelCz what is meant by that is **the center coordinates of the phone's viewport will not move past the boundaries you set**.

Maybe a visual would help. As you can see below, we have a bounding box around the United States (red line) which represents the boundaries specified by the `setMapBoundaries` method. Also we have a visual representation of the phones viewport (purple line). The purple dot in the center of the viewport represents the viewport's center coordinates. Those center coordinates won't move past the boundaries you set by the `setMapBoundaries` method.

Selection_003

I think the confusion we were all having is that we thought the viewport itself (purple line) would stay within the bounds of the red box, but that is not the case.

I think I will be editing my change because after coming back I agree that I still didn't quite clarify as well as I could have.

Hope this helps.

EskelCz:
Ah I see, that's where an image is worth a thousand words :) Thanks!
I got it to work, it was a silly mistake in my implementation.

PR #2716, opened 2019-02-27
Fix Memory Leaks on iOS

kuglemic:
This PR fixes a memory leak on iOS using Apple Maps where AIRMap didn't get deallocated if the component is unmounted during an regionChange animation.

No

Didn't file an issue for it and haven't found one.

### How did you test this PR?

I did test it on the iPhone 8 Simulator and on a real iPhone 7 device.

Steps to reproduce: 

1. Create a simple Test App where you can mount and unmount the Map Component.
2. Mount the Map component
3. Change the region of the map animated, e.g. by calling: `animateToRegion`
4. Make sure that `- (void)mapView:(AIRMap *)mapView regionWillChangeAnimated:(__unused BOOL)animated` gets called. This method did cause the leak.
5. Unmount the Map component during the animation, this means before `- (void)mapView:(AIRMap *)mapView regionDidChangeAnimated:(__unused BOOL)animated` did get called.
6. Notice that `dealloc` didn't get called in AIRMap.m
7. If you Mount the Map component again, you will get a new instance of AIRMap, which causes a memory increase of ~30MB in my simulator


alvelig:
These are the PR we are missing. Thank you. LGTM

PR #2741, opened 2019-03-14
Marker share icon image

zupaox:

Looks No.

Android app's memory usage climbs high when there are thousands of markers on the map, even though most of the markers uses same image and the images are only around 1KB.

This update is to use a shared image icon for markers when they share the same image URI, instead of loading and creating bitmap within each marker, which uses more memory in case we have a lot of them.

### How did you test this PR?

Ideally you should test it with an app that contains a lot of custom markers that shared a small image pool.

To make it simple, we have also created a new page in example app, named "MassiveCustomMarker", which will place 100 custom markers on tap on the map. You can check it out and run the example app. Open android studio profile and watch the memory usage before and after the update while placing more markers on the map.

As a reference, following is the memory profile before the update, tested on Samsung J2 (Samsung SM-G532G). The app crashed on about 500 markers:
image

Following is the memory profile after the update, Java memory consumption alomost remains constant regardless of the number of markers added. App does not crash even after 1000 markers.

image

### DIFF
--- a/example/App.js
+++ b/example/App.js
@@ -45,6 +45,7 @@ import AnimatedNavigation from './examples/AnimatedNavigation';
 import OnPoiClick from './examples/OnPoiClick';
 import IndoorMap from './examples/IndoorMap';
 import CameraControl from './examples/CameraControl';
+import MassiveCustomMarkers from './examples/MassiveCustomMarkers';
 
 const IOS = Platform.OS === 'ios';
 const ANDROID = Platform.OS === 'android';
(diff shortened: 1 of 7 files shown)


rborn:
@zupaox that's awesome ❤️
Do we have the same issue on ios?

LGTM @alveig @christopherdro 🐽

zupaox:
> @zupaox that's awesome ❤️
> Do we have the same issue on ios?
> 
> LGTM @alveig @christopherdro 🐽

hey rborn,

I think iOS will use cached UIImage (loadImageWithURLRequest, RCImageCache), so it will not have the issue. We don't have iOS app crash reports because of memory issue as well.
I just did a test on simulator, it uses 700MB memory for 10K markers and is quite stable, so should not be an issue.

rborn:
@alvelig @christopherdro 🐽

PR #2720, opened 2019-03-03
Fix Android dependencies and build errors

friederbluemle:
### Does any other open PR do the same thing?

No

### What issue is this PR fixing?

Currently, the yarn tasks `start:android` and `run:android` to build and run the Android version are failing on the `master` branch with the following error:

```
* What went wrong:
A problem occurred evaluating project ':react-native-maps-lib'.
> Could not find method compileOnly() for arguments [com.facebook.react:react-native:+] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

This is caused by partially and incorrectly updated dependencies of the Android build system, which was addressed by updating the dependencies as described below. Many obsolete and unused leftover files of old Gradle wrapper installations and a couple of files from the Eclipse days were removed as well.

The three main things this PR addresses is:

- Update Gradle wrapper to 5.2.1 (latest stable version)
- Update Android Gradle plugin to 3.2.1 (version that ships with latest version of React Native)
- Remove unused `appcompat-v7` dependency from library project

See individual commits for more details.

### How did you test this PR?

    react-native start
    yarn run-android
    yarn run-ios

Correct behavior (build and run) was verified using an Android device, an Android emulator, and an iOS simulator.

Includes and closes #2712
Also fixes #2663

<sub><sup>Once approved, please use a *normal* GitHub merge (i.e. **NO rebase/squash** merge) to integrate the commit(s) from the branch of this pull request. The changes are broken up into meaningful, [atomic commits](https://www.freshconsulting.com/atomic-commits/), and my branch should already be up-to-date with the latest target branch. If it isn't, or if you'd like me to combine something, please let me know, and I will make the necessary updates as soon as possible. Thank you!</sup></sub>

Issue #2695, opened 2019-02-13
Unable to resolve Gradle dependencies


michaelsinatra:
### Is this a bug report?

Yes

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

Yes

### Environment

react-native: 0.57.4
react: 16.6.0-alpha.8af6728
react-native-maps: 0.23.0

Target Platform: Android 9 Emulator, Nexus 5 X

### Steps to Reproduce

1. Follow the installation guide to configure the global **settings.gradle** and app's **build.gradle**
2. Hit the Android Studio Gradle's **Sync now** button

### Expected Behavior

I expect the `react-native-maps` project to be included in the main project
and all its dependencies to get resolved by Gradle itself.

### Actual Behavior

```
Unable to resolve dependency for ':react-native-maps@debug/compileClasspath': Could not resolve com.android.support:support-compat:25.2.0.
Open File
Show Details


Unable to resolve dependency for ':react-native-maps@debug/compileClasspath': Could not resolve com.android.support:support-annotations:25.2.0.
Open File
Show Details


Unable to resolve dependency for ':react-native-maps@debug/compileClasspath': Could not resolve com.android.support:support-core-utils:25.2.0.
Open File
Show Details


Unable to resolve dependency for ':react-native-maps@debug/compileClasspath': Could not resolve com.android.support:support-core-ui:25.2.0.
Open File
Show Details


Unable to resolve dependency for ':react-native-maps@debug/compileClasspath': Could not resolve com.android.support:support-fragment:25.2.0.
Open File
Show Details


Unable to resolve dependency for ':react-native-maps@debug/compileClasspath': Could not resolve com.android.support:support-annotations:27.1.1.
Open File
Show Details


Unable to resolve dependency for ':react-native-maps@debug/compileClasspath': Could not resolve com.android.support:support-core-utils:27.1.1.
Open File
Show Details


Unable to resolve dependency for ':react-native-maps@debug/compileClasspath': Could not resolve com.android.support:support-fragment:27.1.1.
Open File
Show Details


Unable to resolve dependency for ':react-native-maps@debug/compileClasspath': Could not resolve com.android.support:support-compat:27.1.1.
Open File
Show Details


Unable to resolve dependency for ':react-native-maps@debug/compileClasspath': Could not resolve com.android.support:support-core-ui:27.1.1.
Open File
Show Details


[...]

PR #2743, opened 2019-03-15
Add appcompat-v7 dependency declaration to lib

friederbluemle:

No

Some people have reported dependency issues with the Android support libraries (see #2695, #2720). The reason is a version conflict that Gradle alone can't resolve. Without explicit declaration, Gradle will use the transitive dependency (`26.0.1` from `play-services-base:16.1.0`). By explicit declaration it is possible to use other versions too.

This PR adds back the `appcompat-v7` dependency (even though the library has no dependencies on it) which was removed in #2720

Also (re-)added the conventional `safeExtGet` function seen in many other RN library projects.

### How did you test this PR?

Tested using the example project.

Step 1: Remove `appcompat-v7` dependency from example (The project has no dependency on `appcompat-v7`!)
Step 2: See project build fail. The reason is RN `0.54.2` which has a compile dependency on `appcompat-v7:23.0.1`.
Step 3: Workaround fix for example project:
```
    implementation('com.facebook.react:react-native:+') {
        exclude group: 'com.android.support', module: 'support-v4'
    }
```
This will prevent Gradle from picking the wrong `23.0.1` version of `support-v4`. The latest version of RN will set it to `28.0.0`.
Step 4: Example project builds!

### DIFF
--- a/lib/android/build.gradle
+++ b/lib/android/build.gradle
@@ -3,12 +3,16 @@ if (project.hasProperty('POM_ARTIFACT_ID')) {
   apply from: 'gradle-maven-push.gradle'
 }
 
+def safeExtGet(prop, fallback) {
+  rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
 android {
-  compileSdkVersion rootProject.ext.compileSdkVersion
+  compileSdkVersion safeExtGet('compileSdkVersion', 28)
 
   defaultConfig {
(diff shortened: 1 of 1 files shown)


rborn:
@friederbluemle stupid question: is step 3 needed or this PR solves it? 😅

friederbluemle:
@rborn - Haha, no this was just the test setup. It's only required if you _remove_ the explicit `appcompat-v7` declaration from app project (which projects by default have).

PR #2790, opened 2019-04-04
fix func 'randomColor' in DefaultMarkers.js

DillionApple:

No.

No related issue.
The previous randomColor function may return invalid color string, like "#12345", which length is not valid. This may generate warning messages in the example page.
The new one fills the string with zeros, like "#012345" in the previous example.

### How did you test this PR?

I use this function to generate many colors randomly, and check if every color string is valid.
After the change, when I click many times to add markers to the map, the warning no longer appears.

### DIFF
--- a/example/examples/DefaultMarkers.js
+++ b/example/examples/DefaultMarkers.js
@@ -19,7 +19,12 @@ const LONGITUDE_DELTA = LATITUDE_DELTA * ASPECT_RATIO;
 let id = 0;
 
 function randomColor() {
-  return `#${Math.floor(Math.random() * 16777215).toString(16)}`;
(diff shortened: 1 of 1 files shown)


rborn:
@DillionApple thanks a lot for the PR. 

In some cases the number is even less than 5 digits so maybe try to use `'1234'.padStart(6,0)`
It should work on both platforms but please run a test. 🤗

rborn:
@DillionApple why did you close the PR? not happy with it ?

DillionApple:
@rborn The pull request didn't not pass the eslint check. I will try the method and make another pull request.

rborn:
@DillionApple if you just overwrite this one  should be  enough

PR #2794, opened 2019-04-05
[iOS] Fix CalloutSubview on Apple maps

nabehide:

No

When using Apple maps and CalloutSubview, the following error happens.

Simulator Screen Shot - iPhone X - 2019-04-06 at 02 08 50

It seemed that CalloutSubview was added but the probject file wasn't updated.

### How did you test this PR?

I tested this PR with Apple maps on iOS in a simulator.


rborn:
LGTM @alvelig 🐽

KiziKr:
I have exactly the same error but on android

PR #2799, opened 2019-04-08
fix func 'randomColor' in DefaultMarkers.js

DillionApple:

No.

No related issue.
The previous randomColor function may return invalid color string, like "#12345", which length is not valid. This may generate warning messages in the example page.
The new one fills the string with zeros, like "#012345" in the previous example.

### How did you test this PR?

I use this function to generate many colors randomly, and check if every color string is valid.
After the change, when I click many times to add markers to the map, the warning no longer appears.


rborn:
@DillionApple thank you ❤️

Issue #2252, opened 2018-05-14
Google Maps e2e tests


compojoom:
### Is this a bug report?
Not really. I'm investigating what needs to change in rn-maps in order to be able to do e2e tests.

### Environment
    react-native: 0.55.0
    react: 16.3.0
    react-native-maps: 0.21.0

  Target Platform: eg: iOS

GoogleMaps

I'm trying to do some e2e detox tests with our map, but that is proving to be rather hard. The current version of the library doesn't add a testID on the markers, so I applied this patch:
https://github.com/react-community/react-native-maps/pull/1170
and I'm trying to get it to work.
The patch is working as expected and the rendered markers have the testID identifier. 

The problem is that I can't match anything that is supposed to be inside of the marker. Our markers are custom react components. I made an UI snapshot of the app and it appears that `AIRGoogleMapMarker` is just an Image rendered on top of the map? Is this the way it is supposed to be? Is there a way to associate the marker with the view it's supposed to render?

christopherdro:
I’m closing this issue because it has been inactive for a while. This probably means that it is not reproducible or it has been fixed in a newer version. If it’s an enhancement and hasn’t been taken on for so long, then it seems no one has the time to implement this.

Please reopen if you still encounter this issue with the [latest stable version](https://github.com/react-native-community/react-native-maps/releases).

Thank you!

PR #2826, opened 2019-04-23
Bugfix/accessibility

danielgindi:

Nope. But a recent one did this partially: https://github.com/react-native-community/react-native-maps/pull/2253
It missed the MapViewManager for both Air and Google, and the MarkerManager for Air.
Also the boolean bugs which prevent the markers from being accessible in Xcode.

#2252, #1104

### How did you test this PR?

By creating automation. Markers are still a pain point in XCUI, especially tapping them, so it is not solved completely. But at least they have their ids.


rborn:
❤️

Issue #2843, opened 2019-05-03
componentWillUpdate is deprecated warning in MapView.js


thekrol01:
### Is this a bug report?

No

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

yes

### Environment

 AppleMaps

### Steps to Reproduce

1. Installed the latest MapView release according to the installation instruction.
2. Build with Xcode 10.2 on physical device iPhone 5.
3. When I want to view the map the following deprecated warning appears:
componentWillUpdate is deprecated etc. with a link to :
https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html
4. MapView works as expected. 

1. 2. 3.

As expected. 

(Write what you thought would happen.)

### Actual Behavior

(Write what happened. Add screenshots!)

### Reproducible Demo

(Paste the link to an example project and exact instructions to reproduce the issue.)

jramalho:
ComponentWillUpdate is a legacy method and will be excluded from React in React 17 (by the docs https://reactjs.org/docs/react-component.html?utm_source=caibaojian.com#unsafe_componentwillupdate), just need to change the logic there to getSnapshotBeforeUpdate and the new region on componentDidUpdate i think.

rborn:
@jramalho fancy a PR ? 🤗

Taym95:
I would like to give this a try!

PR #2849, opened 2019-05-08
Remove componentWillUpdate

Taym95:

No similar PR.

#2843

### DIFF
--- a/lib/components/MapView.js
+++ b/lib/components/MapView.js
@@ -532,13 +532,16 @@ class MapView extends React.Component {
     return { provider: this.props.provider };
   }
 
-  componentWillUpdate(nextProps) {
-    if (nextProps.customMapStyle !== this.props.customMapStyle) {
-      this._updateStyle(nextProps);
+  getSnapshotBeforeUpdate(prevProps, prevState) {
+    if (this.props.customMapStyle !== prevProps.customMapStyle) {
+      this._updateStyle(this.props);
     }
+    return this.props.region;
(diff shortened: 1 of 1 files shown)


rborn:
@Taym95 travis isn't happy :)

Did you test? works good? 🤗

jramalho:
you can use null instead of prevState just to please the Travis linter.

jramalho:
omg this linter is so boring (just kidding, we need it ahah)
you can call the getSnapshotBeforeUpdate like this
``` javascript
getSnapshotBeforeUpdate(prevProps) {
```

and componentDidUpdate like this
``` javascript
componentDidUpdate(prevProps, null, snapshot) 
```
and you can fill the const b with snapshot.region (but you have to test it)
``` javascript
const b = snapshot.region
```

Taym95:
Thanks, I will make the changes and test them.

rborn:
@Taym95 all good? let me know and we merge 🤗

jramalho:
Hello. Tested it here. Works like a charm. Nice work @Taym95 .

brownieboy:
Be warned, if you're not passing in **region** as a prop to your MapView, then you'll get a yellow-screen warning that getSnapshotBeforeUpdate must return a value or null but it actually returned undefined.

FYI, **region** is not actually a mandatory prop.  You could provide **initialRegion** instead, which is what I was doing.  When I switched the prop to **region** the warning went away.

Issue #2847, opened 2019-05-07
Proposal - Remove index.d.ts and consolidate @types/react-native-maps


methodbox:
### Is this a bug report?
No. This is a proposal.

`@types/react-native` contains multiple types that conflict with this module.

There are types in your own index.d.ts that don't match their own definitions in the `react-native` definition; the property `initialRegion`, for example, doesn't exist in the `@types/react-native`, but a definition for `MapView` does, yet doesn't match your own definitions. 

As @types/react-native is practically required to use TS properly with RN, this results in your index.d.ts being overridden and causes a conflict. Rather than update both @types/react-native with definitions that no longer apply to the core RN module *and* match them to your types, consolidation seems appropriate.

My proposal is to remove the `index.d.ts` from this module and consolidate it's definitions into an @types module instead.

The PR @types is to remove `MapView` content entirely as it's not part of the core `react-native` module any longer and conflicts with this (your) module types definitions.

The default recommendation from the `react-native` team is to use @types for TS definitions. Having an `index.d.ts` in your module has the potential to cross name spaces (such as your generic use of the word "Camera") in other type definitions. When moving into the @types module, I would rename this to something like `MapCamera`.

Consolidating this @types would prevent this from happening as, since they are developed under the same repo, it is easy to identify namespace conflicts. Consolidation also has the benefit of maintaining congruence with `react-native` types where your individually maintained type definitions do not.

You can compare the conflicting types here:

https://github.com/methodbox/DefinitelyTyped/blob/master/types/react-native/index.d.ts

The PR I'm working on is here:

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35239

rborn:
@alvelig @christopherdro any thoughts on this?

christopherdro:
Sounds like a great idea.

methodbox:
I have reviewed the types on this companion module for `react-native-maps` module `...directions`: https://github.com/bramus/react-native-maps-directions/blob/master/index.d.ts and they've merged my fix  there; this in part how I discovered the conflict.

If you guys are good with this, I can submit a PR, but I would hope it wasn't merged until I'm able to complete the PR to create the new `@types/react-native-maps-directions` module.

[...]

PR #2877, opened 2019-05-22
Removed index.d.ts, add @types/react-native-maps per #2847

methodbox:

No.

This is regarding a proposal for better management of type definitions - see #2847 .

### How did you test this PR?
This is effectively migrating the types into a separate node module, `@types/react-native-maps`.

This was tested by removing the `index.d.ts` from an existing installation of the current master branch of `react-native-maps` and including `@types/react-native-maps @ 0.24.0` in the devDeps.

Types are recognized exactly as they would be with the original `index.d.ts`. Content of types in `@types/react-native-maps` is identical with the exception of this:

```ts
//  Current RNM index.d.s:

export interface Point {  //  This is here for reference - it wasn't changed.
  x: number;
  y: number;
}
...

compassOffset?: { x: number; y: number };
```

`compassOffset` was updated to use the already-defined `interface Point`, which is 100% analogous to it's current definition:

```ts
//  @types/react-native-maps

compassOffset?: Point;
```

The core purpose of this PR is to remove the `index.d.ts` to prevent a conflict when using `@types/react-native-maps` instead. It has been agreed between collaborators already this is a better method of managing types going forward. This gives a centralized reference point to those who might use the package along with other @types packages, as this is quickly becoming the defacto standard for types reference, particularly with `react-native` as it is the recommended method of importing type definitions.


mcmar:
Will this require every consumer of rn-maps to also `yarn add --dev @types/react-native-maps` manually?

methodbox:
@mcmar

That is how it works, but if you read the linked issue, I’m also exploring adding it as a `peerDependency` which would give them a little warning that they need to do that.

As it stands, maintaining the types here has more of a potential for conflict (again see #2847) than managing @types.

DefinitelyTyped runs Travis against RN types from other packages and the issue I fixed (removing deprecated types from `@types/react-native`) would have been found long ago had they been maintained there.

mcmar:
I think we shouldn't merge this PR. This will clearly result in a worse experience for library consumers. It might make it slightly easier for maintainers, but worse for consumers and I don't think that tradeoff is worth it. After all, this library exists to help developers in the first place, so it doesn't really make sense to make their lives harder just because we don't want to maintain our own types.

mcmar:
I know about DT and use it all the time for other libraries that are missing typedefs in the main repo, but only in that case. I always prefer when typedefs are pre-packaged in the main repo rather than requiring me to import a second repo just for the typedefs.

methodbox:
Did you bother to read the original issue?

The lack of maintaining types that didn’t conflict with those found in `react-native` is what started this.

Your description suggests you don’t understand that.

This doesn’t mean you don’t maintain your types. It means you maintain them in the @types repo.

methodbox:
How exactly does better maintenance with more tests for conflicts make for a worse experience for the user?

mcmar:
It's funny that you ask if I read the original issue. I did and it turns out your issue is actually unrelated to this repo.
It looks like you tried this:
`import { MapView } from 'expo';`
and you got the `MapView` typedef from `@types/react-native` instead of the typedef from `react-native-maps`.
That's a problem with expo, not with this library or with react-native. You then got a PR merged to remove the `MapView` types from `@types/react-native`, but that's not really the right call. The `MapView` type previously existed in react-native and was deprecated, so the typedef still belongs there.

For people who are doing `import { MapView } from 'react-native-maps';`, this is a non-issue because the typedefs won't get mixed up. The reason you had an issue was because expo was re-exporting conflicting typedefs. That's a problem with the expo library and the solution is just to invert the order of the exports.

Your quest to remove the MapView typedefs from react-native and migrate the typedefs for react-native-maps is extremely misguided. What you want to do is just fix the export order in expo. Your PR belongs there.

methodbox:
It’s not a problem with Expo.

React Native tells you to use their types from @types/react-native.

The conflicting types were i their deprecated types and styled-components.

Those have been fixed now (by me) but would have been caught much earlier if the types from this repo were @types and being checked against.

Maybe look at the original issue and discuss with others who own this instead of me.

methodbox:
>For people who are doing import { MapView } from 'react-native-maps';, this is a non-issue because the typedefs won't get mixed up. 

This is incorrect. The types conflict with @types/react-native (before they were removed).

mcmar:
You're confused on several points.

> React Native tells you to use their types from @types/react-native.

React Native recommends using `@types/react-native` because they don't maintain their own types. It's not considered universal best practice to use `@types`.

> The conflicting types were i their deprecated types and styled-components.

The "conflicting types" are not actually conflicting because they're being exported by separate modules, namely `react-native` vs `react-native-maps`. There's only a conflict with expo because they re-export both.

> Those have been fixed now (by me) but would have been caught much earlier if the types from this repo were @types and being checked against.

Your PR on the DT repo just removes types for anyone who was still doing `import { MapView } from 'react-native';` (deprecated). Again, the tests wouldn't have caught the "conflict" because there was no conflict.

> Maybe look at the original issue and discuss with others who own this instead of me.

There's no need to get aggressive. I looked at your original issue. I'm responding here because I don't think there's any good reason to remove the types from this library

> This is incorrect. The types conflict with @types/react-native (before they were removed).

I think I've explained sufficiently why there is no conflict. Your issue is only with the expo library and the solution is very simple. Just invert the order of the exports

methodbox:
> You're confused on several points.
> 
> > React Native tells you to use their types from @types/react-native.
> 
> React Native recommends using `@types/react-native` because they don't maintain their own types. It's not considered universal best practice to use `@types`.
> 
> > The conflicting types were i their deprecated types and styled-components.
> 
> The "conflicting types" are not actually conflicting because they're being exported by separate modules, namely `react-native` vs `react-native-maps`. There's only a conflict with expo because they re-export both.
> 
> > Those have been fixed now (by me) but would have been caught much earlier if the types from this repo were @types and being checked against.
> 
> Your PR on the DT repo just removes types for anyone who was still doing `import { MapView } from 'react-native';` (deprecated). Again, the tests wouldn't have caught the "conflict" because there was no conflict.
> 
> > Maybe look at the original issue and discuss with others who own this instead of me.
> 
> There's no need to get aggressive. I looked at your original issue. I'm responding here because I don't think there's any good reason to remove the types from this library
> 
> > This is incorrect. The types conflict with @types/react-native (before they were removed).
> 
> I think I've explained sufficiently why there is no conflict. Your issue is only with the expo library and the solution is very simple. Just invert the order of the exports

Actually, I think you've taken a very combative attitude about an issue that has been discussed by people who own the repo and your position doesn't consider previous testing that has shown this is an issue.

I would encourage you to install a version of `@types/react-native` that still contains the old definitions, then try to import `MapView` from `react-native-maps`. You seem to be making the assumption that I only tested this with Expo (I didn't), and that I don't realize that Expo has it's own import of `react-native-maps` (I do - and it's pulled right from this repo).

What you haven't explained in any way is why having the types in this repo (which would still be maintained by the owners and even myself) in @types, allowing for prevention of conflicts and crossing namespaces (through already-in-place extensive Travis testing - which doesn't happen here) isn't a better idea than maintaining a separate version which could potentially have conflicts (where the other option has less potential for this) or how that coincides with decisions of people who **own** this repo (again, take it up with them, not me).

I'm not sure what you mean by "invert the order of the exports" but there is nothing that comes to mind by what you _might mean_ that makes sense; there's no controlling what order someone might use packages in.

What we identified was an issue, while intermittent, where the TS compiler would find the `react-native` types for `MapView` before finding those in `react-native-maps`. This isn't an issue if the repo doesn't have a types file.

Rather than pointing to what is clearly your **person opinion** of how this might be handled, maybe you have a better suggestion for preventing the potential for conflict? Or you'd like to try to contribute in some way?

I find it interesting you aren't making this comment on the proposal, but instead on the PR - don't you want people who own the repo to see this?

mcmar:
> Actually, I think you've taken a very combative attitude about an issue that has been discussed by people who own the repo and your position doesn't consider previous testing that has shown this is an issue.

Let's not make bad blood here

> I would encourage you to install a version of @types/react-native that still contains the old definitions, then try to import MapView from react-native-maps. You seem to be making the assumption that I only tested this with Expo (I didn't), and that I don't realize that Expo has it's own import of react-native-maps (I do - and it's pulled right from this repo).

I did exactly that and I did not get the type for `MapView` from `@types/react-native` because it's exported under the module `"react-native"` which does not get searched when doing `import { MapView } from 'react-native-maps';`. In that case, `MapView` doesn't get found, but that's because it's not exported using a named export in `index.d.ts`. The solution there is to fix `index.d.ts`. It has nothing to do with `@types/react-native`.

> What you haven't explained in any way is why having the types in this repo (which would still be maintained by the owners and even myself) in @types, allowing for prevention of conflicts and crossing namespaces (through already-in-place extensive Travis testing - which doesn't happen here) isn't a better idea than maintaining a separate version which could potentially have conflicts (where the other option has less potential for this) or how that coincides with decisions of people who own this repo (again, take it up with them, not me).

The reason we shouldn't move the typedefs out of this repo is because
A) It makes the developer experience worse
B) It doesn't fix anything. I've explained what your conflict was caused by.
C) DefinitelyTyped recommends against it. They actually even go so far as to say that when a library exports its own types, it should be removed from DefinitelyTyped. See here: https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package

As far as "take it up with them, not me", I'm commenting publicly on a github issue for all to see, maintainers and contributors. You're a contributor and you're free to respond.

> I'm not sure what you mean by "invert the order of the exports" but there is nothing that comes to mind by what you might mean that makes sense; there's no controlling what order someone might use packages in.

Expo re-exports 2 typedefs from 2 namespaces, `react-native` and `react-native-maps`. So hypothetically, that might look something like this:
```
export * from 'react-native';
// probably separate file
import * as MapView from 'react-native-maps';
export { MapView };
```
So the `MapView` typedef gets moved from `react-native` module to `expo` module and also, the "conflicting" `MapView` typedef gets moved from 'react-native-maps` module to `expo` module. Again, this is a problem with expo and I'm sure they'd be willing to fix it if you point that out to them.

> What we identified was an issue, while intermittent, where the TS compiler would find the react-native types for MapView before finding those in react-native-maps. This isn't an issue if the repo doesn't have a types file.

That's just not how the TS compiler works. It *does not* pull typedefs from the `react-native` module when you do `import { MapView } from 'react-native-maps';`. However, it might do that when you do `import { MapView } from 'expo'`; because expo has re-exported it from `react-native`.

> Rather than pointing to what is clearly your person opinion of how this might be handled, maybe you have a better suggestion for preventing the potential for conflict? Or you'd like to try to contribute in some way?

There isn't any potential for conflict. Your issue is with a completely different repo. Moving the types from here to DT won't even fix your issue because they'll still be getting re-exported from expo in the same order.

> I find it interesting you aren't making this comment on the proposal, but instead on the PR - don't you want people who own the repo to see this?

No need for conspiracy theories. "The people who own the repo" as you call them can see the conversation on the pull request and will hopefully know that this should not be merged. The reason I'm commenting here is because I don't think this should be merged. I'd be perfectly happy to link to our discussion on the issue so that everyone who's following that discussion can be made aware of it

rborn:
@methodbox @mcmar this is such a sad thing happening here, would be really nice if you two make peace 😔

Everybody wants to help here so even if we have strong opinions is important to be expressed in such a way we don't hurt nobody 

🤗

methodbox:
@rborn

Agreed.

My only argument at this point is this:

> Last, regardless of your view of my "misguided quest" - doesn't it make more sense to maintain types in a single location which also has the added benefit of being more thoroughly tested, avoiding the potential for conflicts (regardless if "my" conflicts are correct or not)?

If that isn’t important to the team working on this, then leave it alone and we will close this PR.

I’ve linked to the Travis test as an example of this. If that level of testing against other types packages is valuable to you I stand behind the proposal.

rborn:
@methodbox @mcmar  Is not for me to take a decision. 

However we are in the process to unify somehow all the packages that are under the `react-native-community` organization.
We started with the eslint rules and we are working toward trying to have the same structure for as many common things for all the modules.

I believe `types` can be one of them so I'll go and cry for help and see what's the general opinion.

Sounds good?

rborn:
@methodbox @mcmar just an update - there is a discussion opened now on how to standardise type definitions. But this means it will take a while till a decision will be made so I'll mark the PR as `waiting for reply` and close #2847 to limit the noise.

@methodbox I'll add a reference of this PR in the discussion so we won't forget 🤗
I'm sorry I can't have a more definitive and faster answer.

methodbox:
@rborn No problem. Thanks for the update!

PR #2881, opened 2019-05-23
Update google maps pods to 3.1.0

TVke:

No other PR does the same thing

It fixes a error that occurs in react-native-google-maps 3.0.3 and not in 3.1.0

### How did you test this PR?

I tested it by changing the node modules in my project and the bug was fixed
iOS (android the bug did not occur)
on simulator

### DIFF
--- a/react-native-google-maps.podspec
+++ b/react-native-google-maps.podspec
@@ -17,6 +17,6 @@ Pod::Spec.new do |s|
   s.compiler_flags = '-DHAVE_GOOGLE_MAPS=1', '-DHAVE_GOOGLE_MAPS_UTILS=1', '-fno-modules'
 
   s.dependency 'React'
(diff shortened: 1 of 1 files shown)


rborn:
@TVke 🤗 what was the error fixed by this upgrade? Just curious

TVke:
@rborn This was the issue with the problem and solution:
https://github.com/tolu360/react-native-google-places/issues/222

cormander:
Yes! I'm on an older version of maps and can't update due to the google maps pods version -- just tested this version and it's working for me now. Can't wait for the next release.

thecachedbyte:
When this will be released? and what will be the corresponding compatible react-native version?

jmikhail:
Waiting impatiently for this to be released. Any updates??

PR #2905, opened 2019-06-11
Update polyline.md

bill-pairaktaridis:
Strokecolor also supports rgba

No

None. This is just an update to the documentation.

### How did you test this PR?

Used `rgba(79, 160, 255, 0.15)` for the strokeColor prop and worked as expected


rborn:
❤️

Issue #2756, opened 2019-03-21
API for determining if user manually moved the map


jerryryle:
### Is this a bug report?

No

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

Yes

### Environment

react-native: 0.59.1
react: 16.8.3
react-navigation: 3.3.2
react-native-maps: 0.23.0
iOS+Android

### Feature Request Description

A common use case for maps seems to be to track the user's location ("follow me") until the user manually moves the map--at which point, the app might display a crosshair or similar icon to allow the user to re-enable "follow me" mode. To implement this, we need to know whether a map's region update was the result of a programmatic update or due to user input. Currently, there doesn't seem to be an authoritative means for determining this without impacting the UX. Under iOS, using `onPanDrag` only works if `scrollEnabled` is `false`, so this event isn't useful for this scenario (and I'm not sure if it fires if a user manages to zoom without panning). Using the `onMoveShouldSetResponder` has been proposed as a solution, but it seems to have issues interacting with [react-native-gesture-handler](https://github.com/kmagiera/react-native-gesture-handler), which popular libraries such as [react-navigation](https://github.com/react-navigation/react-navigation) use. Regardless, using the likes of `onMoveShouldSetResponder` seems burdensome and potentially conflicting with the map gesture handlers, so it would be nice if `MapView` exposed an official API for determining whether the user has panned or zoomed the map.

Would it be possible to add something like a `userInputChanged` boolean to the `onRegionChange` and `onRegionChangeComplete` events?

oliverdolgener:
I'd like to do the same thing. Currently I'm using `onPanDrag()` on Android which works exactly the way I want. On iOS though I use `onRegionChangeComplete()` and check if the user's current position is inside the viewport region. If not I disable `follow me`. But a consistent solution for both platform would be awesome!

failbottt:
@oliverdolgener 

> `On iOS though I use onRegionChangeComplete() and check if the user's current position is inside the viewport region. `

Mind sharing how you check if the user's current position is inside of the viewport?

oliverdolgener:
```
isCoordinateInRegion(coordinate, region) {
  const top = region.latitude + region.latitudeDelta / 2;
  const bot = region.latitude - region.latitudeDelta / 2;
  const right = region.longitude + region.logitudeDelta / 2;
  const left = region.longitude - region.logitudeDelta / 2;

[...]

Issue #1620, opened 2017-09-05
how detect onRegionChangeComplete is user action or animation?


kdong007:
After I call `fitToCoordinates` on map, `onRegionChangeComplete` is also triggered. Then how does `onRegionChangeComplete` know if the change event is from user action or auto animation?

heysailor:
It'd be nice, sure. As a workaround, I set an `animation` variable `true` before firing a method like `fitToCoordinates`, and set it back to `false` after the duration of the animation using `setTimeout`.

Then in `onRegionChangeComplete` I can check if the change was due to an animation or from user input.

MoOx:
I have to admit it can be pretty annoying... I would expect that onRegionChangeComplete is only called on user interaction.
I think now that we have a flag with onMapReady, onRegionChange(Complete) should not be triggered until the map is ready.

IjzerenHein:
On Android, this is possible with the new Google Maps `onCameraMoveStartedListeners` API. Some work has been done in PR #1710 to support this, but it still needs to be added the `onRegionChange` API and propagated to JS. And obviously, iOS support needs to be added for this as well.

Jun711:
I think when 'adding new markers to the map' action is done, it fires onRegionChangeComplete too. Is that right?

outaTiME:
So? any update? 
im dealing with this issue on iOS the onRegionChangeComplete method fires intermittently but on Android fires according each region update.

Any way to achieve that? thanks !!!

z4qx:
Same problem for me. Please can you provide a real solution like another callback ?

harrisrobin:
Mapbox provides a [bool](https://github.com/react-native-mapbox-gl/maps/blob/c675b0566480656b84880be5418639bcb382e463/example/src/examples/ShowRegionDidChange.js#L122) in the event object, but would be nice to have something similar here!

yairopro:
Any news about this issue ?
Any workaround to distinct user action from animation ?

yairopro:
🐸

yairopro:
🐼

yairopro:
🐸

aliwaqar981:
Anyone found solution for this?

yairopro:
@aliwaqar981 

```jacascript
function onRegionChangeComplete(region, {isGesture}){
  if (isGesture){
    // ...
  }
}
```

Issue #1038, opened 2017-02-12
"Follow Me Button": Map stutters slowly when using onPanDrag + scrollEnabled={false} on iOS


yaronlevi:
I am trying to implement a simple "follow me" switch (like Google Maps).
When the switch is ON, the map region is automatically updated every couple of seconds and follows the user.
But when the user **manually dose any drag gesture** on the map, the switch will go to OFF mode and the map will not follow the user. 

I used **onPanDrag** to know when the user manually moves the map. On **android** it's working fine:

```
<MapView 
            region={this.state.region}
            onPanDrag={(e) => { this.setState({isFollowingUser:false}) }}
            onRegionChange={(region) => { this.setState({region}) }}>
```

The code above works great: When the user drags the map, I set isFollowingUser to FALSE. Later in the geoLocationWatchPosition callback I see that isFollowingUser equals FALSE, and do not update the region to the new location.

But on **iOS** , I must set scrollEnabled:false in order make onPanDrag work. But now, because scrollEnabled is FALSE, the map won't move when the user interacts with it. So I tried to do this inside the  onPanDrag:

```
onPanDrag(e) {
    const coord = e.nativeEvent.coordinate;
    const newRegion = {
      latitude: coord.latitude,
      longitude: coord.longitude,
      latitudeDelta: LATITUDE_DELTA,
      longitudeDelta: LONGITUDE_DELTA
    };
      this.setState({
        region: newRegion
      });
  }

```
The map moves, but it is barley usable. It's very slow and stutters. I think it is related to the fact that setState is called many times a second, but how can I do it otherwise?

tianhaoz95:
I ran into the same problem. I guess the problem is that the coordinates are not that accurate and re-render too frequently will make it unusable. So I tried something similar to a PID controller. When I get a new coordinate, I compare with the last one and only set a small change. It worked better than before, but still kind of shitty.


`var new_latitude = this.state.show_latitude + (e.nativeEvent.coordinate.latitude - this.state.show_latitude) * 0.2;`
    
`var new_longitude = this.state.show_longitude + (e.nativeEvent.coordinate.longitude - this.state.show_longitude) * 0.2;`
    
`this.setState({
      show_latitude: new_latitude,
      show_longitude: new_longitude,
    });`

svenadlung:
Having the same problem. Any solutions on this?

Rafkraft:
+1

rexlow:
+1

[...]

PR #2935, opened 2019-07-01
Allow using onPanDrag while scrollEnabled=true

nabati:

Previous work has been done in #2692 - where this was extracted from - and #2314. 

The previous PR, while good, has been open since February and has seen no activity lately. Also, there are some concerns regarding the API it defines, where `onZoom`, `onPinch` etc could be a better alternative. 

This PR is a much narrower scope than the previous one, and should be simpler to get merged. It also presents no conflict if the former one would get merged.

Allows `onPanDrag` to be used while scrollEnabled is true, which is currently not supported. This seems like the best entry point for determining when the user has interacted with the Map. There are multiple requests relating to this requirement; #2756 #1038 #1620 (last not explicitly the same - but seems like the same base issue).

Also, removes the `[drag setMaximumNumberOfTouches:1];` call so that if one pans with more than one finger, the handler still fires. This is consistent with how it works on Android.

### How did you test this PR?

1. Create a Map component with an `onPanDrag` handler
2. Pan the map
3. onPanDrag should fire

# Verification

This is only enabled on iOS with Apple Maps. It's been tested both on the simulator and on device.

### DIFF
--- a/lib/ios/AirMaps/AIRMapManager.m
+++ b/lib/ios/AirMaps/AIRMapManager.m
@@ -60,7 +60,6 @@ RCT_EXPORT_MODULE()
     UILongPressGestureRecognizer *longPress = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(handleMapLongPress:)];
     UIPanGestureRecognizer *drag = [[UIPanGestureRecognizer alloc] initWithTarget:self action:@selector(handleMapDrag:)];
     [drag setMinimumNumberOfTouches:1];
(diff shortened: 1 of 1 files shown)


nabati:
Any chance to get this reviewed, @christopherdro ?

ffpetrovic:
Is the note on https://github.com/react-native-community/react-native-maps/blob/master/docs/mapview.md#events
> Callback that is called when user presses and drags the map. NOTE: for iOS scrollEnabled should be set to false to trigger the event

then still valid?

PR #2937, opened 2019-07-03
Add onDoublePress callback

nabati:

Adds `onDoublePress` callback to Google Maps on Android and Apple Maps on iOS. Previous related work in #2692 and #2935. Created this PR as an alternative to #2692 with the reasoning that 

- the API is more explicit; adds a new `onDoublePress` callback instead of mixing multiple gestures to be handled by a single callback.
- the PR only does a single thing rather than multiple changes at once
- the previous PR has stalled

Adds `onDoublePress` callback to Google Maps on Android and Apple Maps on iOS.

This can be used for multiple things, but primarily, to know when the user has interacted with the map by double tapping the screen.

### How did you test this PR?

Tested with emulator on Android and simulator on iOS.

1. Add onPress and onDoublePress callback handlers to the map
2. Observe that tapping once fires only onPress once, and tapping twice fires only onDoublePress once. 

I did not include changes for Google Maps on iOS since this does not match our use case. Let me know if it's a requirement to get it merged.


lib/ios/AirMaps/AIRMapManager.m
```diff
@@ -55,18 +57,26 @@ - (UIView *)view
...
-
+    
+    doubleTap.delegate = self;
+    
     // disable drag by default
     drag.enabled = NO;
-
+    
     [map addGestureRecognizer:tap];
+    [map addGestureRecognizer:doubleTap];
```
nabati: I wanted to shuffle around the above code - I think it would make more sense to initialise and configure them type-wise rather than operation-wise, but I didn't want to make the review harder than it has to be.

nabati:
Any way to get this reviewed @rborn ?

nabati:
Since you're on a roll @christopherdro, could you please have a look at this as well?

rborn:
@nabati could you please fix the conflicts? Thnx

nabati:
Fixed 👌 @rborn

nabati:
Thanks @christopherdro !

PR #2959, opened 2019-07-11
Added GeoJSON support by default

Peretz30:
No

None

### How did you test this PR?

Tested on Android Real Device

I think that the less dependencies the better and GeoJSON support should be by default in this package.

```
import React from 'react';
import MapView, {Geojson} from 'react-native-maps';

const alcatraz = {
  type: 'FeatureCollection',
  features: [
    {
      type: 'Feature',
      properties: {},
      geometry: {
        type: 'Point',
        coordinates: [-122.42305755615234, 37.82687023785448],
      }
    }
  ]
};

const Map = props => (
  <MapView>
    <Geojson geojson={alcatraz} />
  </MapView>
);
```

### DIFF
--- a/lib/components/Geojson.js
+++ b/lib/components/Geojson.js
@@ -1,0 +1,117 @@
+import React from 'react';
+import MapView from './MapView';
+import uuid from 'uuid';
+
+export const makeOverlays = features => {
+    const points = features
+        .filter(f => f.geometry && (f.geometry.type === 'Point' || f.geometry.type === 'MultiPoint'))
+        .map(feature => makeCoordinates(feature).map(coordinates => makeOverlay(coordinates, feature)))
+        .reduce(flatten, [])
+        .map(overlay => ({ ...overlay, type: 'point' }));
+
+    const lines = features
+        .filter(
+            f => f.geometry && (f.geometry.type === 'LineString' || f.geometry.type === 'MultiLineString')
+        )
+        .map(feature => makeCoordinates(feature).map(coordinates => makeOverlay(coordinates, feature)))
+        .reduce(flatten, [])
+        .map(overlay => ({ ...overlay, type: 'polyline' }));
+
+    const multipolygons = features
+        .filter(f => f.geometry && f.geometry.type === 'MultiPolygon')
+        .map(feature => makeCoordinates(feature).map(coordinates => makeOverlay(coordinates, feature)))
+        .reduce(flatten, []);
+
+    const polygons = features
+        .filter(f => f.geometry && f.geometry.type === 'Polygon')
+        .map(feature => makeOverlay(makeCoordinates(feature), feature))
+        .reduce(flatten, [])
+        .concat(multipolygons)
+        .map(overlay => ({ ...overlay, type: 'polygon' }));
+
+    return points.concat(lines).concat(polygons);
+};
+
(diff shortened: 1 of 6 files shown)


rborn:
@christopherdro @alvelig any thoughts on this?

lib/components/Geojson.js
```diff
@@ -0,0 +1,117 @@
...
+
+    return points.concat(lines).concat(polygons);
+};
+
+const flatten = (prev, curr) => prev.concat(curr);
+
+const makeOverlay = (coordinates, feature) => {
+    let overlay = {
+        feature,
+        id: feature.id ? feature.id : uuid(),
```
nabati: Does this use of uuid really make sense? It will be a new uuid on every pass. Same further down.

Peretz30:
@nabati are we really need this `GeometryCollection` type? There are no issues about that in https://github.com/frankrowe/react-native-geojson/issues for 2 years. So, I think wasting time to implement this feature is not worth it.

lib/components/Geojson.js
Peretz30: Id is used for key prop of array of React components.

alvelig: Right, `uuid` doesn't make any sense here. You could use `index` in
```
overlays.map((overlay, index) => {
  key={overlay.id || index}
}
```
Also it isn't necessary in
```
<React.Fragment key={props.geojson.id || uuid()}>
```

alvelig: uuid is a quite costly operation, it shouldn't be used everywhere

alvelig:
Considering my notes on the `uuid` use case, LGTM

Peretz30:
Deleted uuid generation, used index instead.

Peretz30:
So, that about to merge?

alvelig:
LGTM
@rborn @christopherdro 🐽

christopherdro:
LGTM. Could we update the docs and add a basic version in the example app?

Peretz30:
@christopherdro Added geojson description to docs and examples

Peretz30:
@christopherdro @rborn @alvelig Guys, is there anything else I should do?

christopherdro:
@peretz30 Could you address the lint errors show in travis?

Peretz30:
@christopherdro done

GuillotJessica:
@Peretz30 will you add bbox feature as it is a common thing on geojson too ? but i don't think your library handles it yet

Peretz30:
@GuillotJessica did you mean `getMapBoundaries` ? It's method of `MapView`. `Geojson` is a component that you should to put inside `MapView` as shown in example. You can use `getMapBoundaries` as before.

GuillotJessica:
@Peretz30  i meant to display a map regarding the bbox received in a geojson not to get the bbox displayed

Peretz30:
@GuillotJessica I am sorry, but I don't understand what you want to do. Can you, provide more details about it?

GuillotJessica:
i mean if i provide a geojson with boundingBox, i would like the library mapBoundaries to be set accordingly

Peretz30:
@GuillotJessica How do you want to provide geojson with boundingBox? Can you, please, post your question with some code on [StackOverflow](https://stackoverflow.com/)

Issue #2923, opened 2019-06-24
Imprecise user location on Android


funkyfourier:
# Bug

When the app is run for the first time on Android, or rather when the location permission dialog pops up and is granted by the user, the position of the user location is very imprecise. In those cases, the location icon in the status bar is also not shown.

User location is displayed using the `showsUserLocation` property.

If the app is killed then opened again, positioning is fine and the location icon is shown in the status bar.

## Environment info

React native info output:

```bash
  React Native Environment Info:
    System:
      OS: macOS 10.14
      CPU: (8) x64 Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
      Memory: 118.55 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 11.11.0 - /usr/local/bin/node
      Yarn: 1.3.2 - /usr/local/bin/yarn
      npm: 6.7.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 16, 23, 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 25.0.0, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 27.0.3, 28.0.1, 28.0.2, 28.0.3
        System Images: android-19 | Google APIs Intel x86 Atom, android-25 | Google APIs Intel x86 Atom, android-25 | Google APIs Intel x86 Atom_64, android-25 | Google Play Intel x86 Atom, android-26 | Google APIs Intel x86 Atom_64
    IDEs:
      Android Studio: 3.3 AI-182.5107.16.33.5199772
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: 0.59.3 => 0.59.3 
    npmGlobalPackages:
      create-react-native-app: 1.0.0
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7
```

Library version: 0.24.2

## Steps To Reproduce

1. Open Android version of the app.
2. Grant permission in dialog.
3. showsUserPosition is now very imprecise, and location icon in status bar is not shown.
4. Kill/restart app. User positioning is now working fine, and location icon is shown in status bar.
[...]

PR #2993, opened 2019-07-30
Improved user location on Android

funkyfourier:

Not that I can see.

https://github.com/react-native-community/react-native-maps/issues/2923

Also, it brings some new possibilities for controlling power consumption and accuracy of user location in maps, specifically `userLocationPriority`, `userLocationUpdateInterval` and `userLocationFastestInterval`. This is achieved by setting a custom LocationSource on the GoogleMap object, which exposes these properties and does not cause issue #2923. 

### How did you test this PR?

This PR applies only to Android. I have described in the issue mention above how to recreate the issue, and this PR solves it. It has been tested on a OnePlus 6 and a Sony XZ1.


lib/android/build.gradle
```diff
@@ -21,6 +21,7 @@ dependencies {
     exclude group: 'com.android.support'
   }
   implementation "com.android.support:appcompat-v7:${safeExtGet('supportLibVersion', '28.0.0')}"
+  implementation "com.google.android.gms:play-services-location:16.0.0"
```
felipecsl: I guess this should be using the same `${safeExtGet('playServicesVersion', '16.1.0')` mechanism?

lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
```diff
@@ -107,6 +107,7 @@
   private boolean destroyed = false;
   private final ThemedReactContext context;
   private final EventDispatcher eventDispatcher;
+  private FusedLocationSource fusedLocationSource;
```
felipecsl: make this `final` too

lib/android/src/main/java/com/airbnb/android/react/maps/FusedLocationSource.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.android.gms.location.LocationResult;
+import com.google.android.gms.location.LocationServices;
+import com.google.android.gms.maps.LocationSource;
+import com.google.android.gms.tasks.OnSuccessListener;
+
+public class FusedLocationSource implements LocationSource {
+
+    private FusedLocationProviderClient fusedLocationClientProviderClient;
+    private LocationRequest locationRequest;
+    private LocationCallback locationCallback;
```
felipecsl: final `fusedLocationProviderClient` and `locationRequest `

lib/android/build.gradle
funkyfourier: Actually I discovered that play-services-location does not exactly follow the same versions as the rest of Play Services.

For example, play-services-location does not have a version 16.1.0:
https://mvnrepository.com/artifact/com.google.android.gms/play-services-maps

But for example play-services-maps does:
https://mvnrepository.com/artifact/com.google.android.gms/play-services-maps

Or am I missing something? If not, there are two ways to do it:

1) Use a constant hard coded version like I did.
2) Expose another Gradle variable like "locationServicesVersion" or something like that.

felipecsl: You're right. I looked directly in the [Google maven repo](https://dl.google.com/dl/android/maven2/index.html) too and it doesnt seem to have version 16.1.0 for `play-services-location`. That's confusing!
Thanks for explaining, this seems reasonable then!

lib/android/src/main/java/com/airbnb/android/react/maps/FusedLocationSource.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.google.android.gms.location.FusedLocationProviderClient;
+import com.google.android.gms.location.LocationCallback;
+import com.google.android.gms.location.LocationRequest;
+import com.google.android.gms.location.LocationResult;
+import com.google.android.gms.location.LocationServices;
+import com.google.android.gms.maps.LocationSource;
+import com.google.android.gms.tasks.OnSuccessListener;
+
+public class FusedLocationSource implements LocationSource {
```
felipecsl: Just for the record:
This file is indented with 4 spaces while the rest of the project is indented with 2 spaces.
It's non blocking, we should create a separate pull request for automatic style checking/formatting.

christopherdro:
@funkyfourier Could you resolve the conflicts when you get a chance?  Here are the changes that cause the conflict. https://github.com/react-native-community/react-native-maps/commit/801614efcdca00fa2dbad89fca6af1607b4184e1#diff-ac45350dc94234f6d3c8b856aacbc395

funkyfourier:
Was the fixes ok? When will this be merged?

Thank you.

christopherdro:
@funkyfourier Any thoughts on the issue linked to this PR?

funkyfourier:
I'll have a look tomorrow.

PR #3007, opened 2019-08-07
Add Overlay onPress event for Android

formatlos:

No

Adds `onPress` callback and `tappable` prop to `Overlay` on Android.
Fixes the issue that the MapView onPress events are getting swallowed by Overlays.

### How did you test this PR?

Tested with emulator and physical device on Android.

Add onPress callback handlers to the overlay and the MapView. 
With `tappable` set `true` the callback of the overlay fires as expected. With `tappable` set `false` the callback of the MapView fires.


rborn:
LGTM ❤️

@christopherdro  @alvelig  🐽

Question: do we have this for iOS already? i not can be implemented for platform parity?

formatlos:
@rborn as far as I saw it's not implemented for iOS. I gave it a quick try, but I'm so not into iOS stuff

adnkh:
Really thanks for this ❤️

Exited to see it in the next release

Issue #2318, opened 2018-06-11
MapView's onPress-event is not fired when clicking on an Overlay-image (Android only)


Timmitry:
### Is this a bug report?

yes

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

yes

### Environment

"react": "16.3.1"
"react-native": "0.55.3",
"react-native-maps": "0.21.0",

### Steps to Reproduce

Render the `MyMapView` component posted below.
This should show a 300x300px `MapView` with ocean in the background, an `Overlay` showing Octocat, and a `Marker` in the middle:

<img width="190" alt="image" src="https://user-images.githubusercontent.com/20927653/41222975-c9dc93c4-6d68-11e8-8e77-504b0fdeb82f.png">

Click/press on the `MapView`, either on the Octocat-image, or on the ocean in the background.

### Expected Behavior

The `onPress` event of the `MapView` should be fired, and a marker be added at the position that was clicked.

### Actual Behavior

On iOS, the marker is always added as expected.

On Android, the marker is added as expected when clicking on the blue background. However, it is not added when clicking on the `Overlay`-image. I verified in the debugger that the event is not being fired at all when clicking on octocat.

### Reproducible Demo

```js
import * as React from 'react';
import MapView from 'react-native-maps';

export default class MyMapView extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      markers: [
        { latitude: 0.001, longitude: 0.0 },
      ],
    };
  }

  addMarker(coordinates) {
    this.setState({
      markers: [...this.state.markers, { latitude: coordinates.latitude, longitude: coordinates.longitude }]
    });
  }

  render() {
    return (
      <MapView
        initialRegion={{
          latitude: 0,
          longitude: 0,
          latitudeDelta: 0.03,
          longitudeDelta: 0.03,
        }}
        width={300}
        height={300}
        onPress={event => this.addMarker(event.nativeEvent.coordinate)}
      >
        <MapView.Overlay
          image={'https://assets-cdn.github.com/images/modules/logos_page/GitHub-Mark.png'}
          bounds={[[0.01, -0.01], [-0.01, 0.01]]}
        />
        {this.state.markers.map(marker =>
          (<MapView.Marker
            key={marker.index}
            coordinate={{ latitude: marker.latitude, longitude: marker.longitude }}
          />)
        )}
      </MapView>
    );
  }
}
```

[...]

PR #3008, opened 2019-08-07
Fix MapView OnPress when pressing on Overlay

raphaelv1:

Might conflict with #3007, but doesn't do the same thing.

#2318

### How did you test this PR?

Changed it in my local node_modules and the problem was fixed. Didn't seem to cause any other problems. It's definitely not elegant, but it should work as a temporary fix.


formatlos:
this problem will also be fixed with https://github.com/react-native-community/react-native-maps/pull/3007 as it makes the value accessible through the `tappable` prop

PR #3018, opened 2019-08-15
add support for passing headers to MapOverlay image

formatlos:

No

Adds support for passing `headers` object to `Overlay` `image` to be able to send along HTTP headers with requests for a remote image.

### How did you test this PR?

Tested with emulator and physical device on Android and simulator on iOS. 

Things I've tested:
- passing `image` as `string` instead of `ImageURISource`
- passing local (`require`) images
- passing `image` as `ImageURISource` without headers
- passing `image` as `ImageURISource` with headers (loading an image from an API which needs an api key)

### DIFF
--- a/docs/overlay.md
+++ b/docs/overlay.md
@@ -22,3 +22,10 @@ type LatLng = [
   longitude: Number,
 ]
 ```
+
+```
+type ImageSource = {
+  uri?: string;
+  headers?: { [key: string]: string };
+}
+```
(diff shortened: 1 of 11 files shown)


SaeedZhiany:
a

sfadiga:
Hi,
Sorry if I'm posting this in the wrong place. But I'm using Overlay with image headers to request an image with an Authentication token. And it seems to me that this is not working. Is this branch aims to solve this? Or was this feature was already merged? Can someone please point me in the right direction? Regards.

PR #3020, opened 2019-08-15
rn 0.60.5 support

jgtoriginal:
`react-native 0.60.5 - package android.support.v4.content does not exist`

### DIFF
--- a/lib/android/build.gradle
+++ b/lib/android/build.gradle
@@ -20,7 +20,7 @@ dependencies {
   implementation('com.facebook.react:react-native:+') {
     exclude group: 'com.android.support'
   }
-  implementation "com.android.support:appcompat-v7:${safeExtGet('supportLibVersion', '28.0.0')}"
+  implementation "${safeExtGet('coreLibName', 'com.android.support:appcompat-v7')}:${safeExtGet('supportLibVersion', '28.0.0')}"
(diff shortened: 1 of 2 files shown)


DavidPineda:
Hi @jgtoriginal, I have that same change and it generates this error.
 
https://github.com/react-native-community/react-native-maps/issues/2495

Do you have the library with autolink?

cipriancaba:
Have you tried using `jetify` in a `postinstall` in `package.json`? That should fix your issue and it's how older libraries are being transitioned to AndroidX

matinzd:
Any release on these pull requests?!??!? Last release is for month ago !

jgtoriginal:
I can confirm that following @cipriancaba suggestion on the main project (the one that requires this module) sorts things out. Which puts us in a chicken and an egg situation.
Because if we think that this library should support RN 0.60 out of the box without needing further configurations on the host project in order to work, then this PR should be merged to master, no doubt.
But if we think that there are many users still on RN =< 0.59, merging this would brake their projects and force them to upgrade to RN 0.60 and go for `jetify` for other modules. Which will eventually happen, maybe not just yet.

salah-ghanim:
@jgtoriginal this library should support the latest RN version. users with lower RN versions don't have to update.

paolorovella:
@aurangzaib @jgtoriginal any news on this? `react-native` was updated to `v0.61.1` so i think it's time to merge this PR 🙂

Issue #3100, opened 2019-10-01
[iOS][cacheEnabled = true] Apple map goes black when switch to another app or goes to background


r0b0t3d:
# Bug

## Environment info

React native info output:

```bash
System:
    OS: macOS 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
    Memory: 18.67 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.15.3 - /usr/local/opt/node@10/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.4.1 - /usr/local/opt/node@10/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.0, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
    Android SDK:
      API Levels: 25, 27, 28, 29
      Build Tools: 23.0.1, 23.0.2, 23.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 27.0.0, 27.0.1, 27.0.2, 27.0.3, 28.0.0, 28.0.0, 28.0.0, 28.0.1, 28.0.2, 28.0.3, 29.0.2
      System Images: android-27 | Google APIs Intel x86 Atom, android-27 | Google Play Intel x86 Atom, android-28 | Google Play Intel x86 Atom, android-29 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 11.0/11A420a - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0
    react-native: 0.61.1 => 0.61.1
  npmGlobalPackages:
    react-native-cli: 2.0.1
    react-native-create-library: 3.1.2
    react-native-git-upgrade: 0.2.7
```

Library version: 0.25.0

## Steps To Reproduce

1. Enable `cacheEnabled`
2. Open screen with apple mapview then hide app to background
3. Open app again -> map goes black
![before](https://user-images.githubusercontent.com/8109285/65954102-88d9f900-e46f-11e9-8a1e-3417cec5145f.png)
[...]

PR #3104, opened 2019-10-02
only handle cacheEnable when app is active

r0b0t3d:

No

#3100 
MapView went black in iOS (apple map) with `cacheEnable` when switching to other apps or goes to background

### How did you test this PR?

Local test on my phone.
Step 1. Open mapview with cacheEnable
Step 2. Switch to other apps then switch back -> mapview still visible
Step 3. Hide app to background then open again -> mapview still visible


rborn:
LGTM 

@christopherdro 🐽

markrickert:
We identified the bug and while researching a fix, i tried this and it worked. 👍 :shipit:

PR #3106, opened 2019-10-03
Add supportLibVersion safeguard in gradle build

cristeahub:

Not that I can see

In 801614efcdca00fa2dbad89fca6af1607b4184e1 support for react native 0.60.5 was added, however it didn't safely check if `supportLibVersion` was set, which broke my build. A simple change like this fixed this issue.

Commit message:
```
This change makes sure that we safely check to see if supportLibVersion
is set before we try to use this.
```

### How did you test this PR?

Before this change nothing would build, but with this change it builds just fine.

### DIFF
--- a/lib/android/build.gradle
+++ b/lib/android/build.gradle
@@ -17,6 +17,7 @@ android {
 }
 
 dependencies {
+  def supportLibVersion = safeExtGet('supportLibVersion', '28.0.0')
   def supportLibMajorVersion = supportLibVersion.split('\\.')[0] as int
   def appCompatLibName =  (supportLibMajorVersion < 20) ? "androidx.appcompat:appcompat" : "com.android.support:appcompat-v7"
   implementation "$appCompatLibName:$supportLibVersion"


lib/android/build.gradle
```diff
@@ -17,6 +17,7 @@ android {
 }
 
 dependencies {
+  def supportLibVersion = safeExtGet('supportLibVersion', '28.0.0')
```
NoumanSakhawat: Work perfectly!

huzaifa3115: You guys should properly on npm as this line is missing when run npm install react-native-maps

Shahrukh0396: working perfectly!

naderghanbari: This still happens for me on `"react-native": "0.61.5"` and `"react-native-maps": "0.26.1"`:

```
* Where:
Build file '.../node_modules/react-native-maps/lib/android/build.gradle' line: 20

* What went wrong:
A problem occurred evaluating project ':react-native-maps'.
> Could not get unknown property 'supportLibVersion' for object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

naderghanbari: Works fine! Any plan to release a patch version with this fix?

PR #3115, opened 2019-10-07
[iOS] Expose isAccessibilityElement as a prop

computerjazz:

No

https://github.com/react-native-community/react-native-maps/issues/3114

When the MapView has `isAccessibilityElement: YES`, the map captures any accessibility taps and does not forward them on to the markers. Passing an `isAccessibilityElement` prop allows the developer to expose map children (ie markers) to VoiceOver and/or other accessibility tools.

### How did you test this PR?

- Created a map with custom markers.
- Set `isAccessibilityElement: false` on map.
- Turned VoiceOver on
- Tapped on markers
- Verified that VoiceOver recognized markers, allowed me to tap into them

### DIFF
--- a/lib/ios/AirGoogleMaps/AIRGoogleMapManager.m
+++ b/lib/ios/AirGoogleMaps/AIRGoogleMapManager.m
@@ -65,6 +65,7 @@ RCT_EXPORT_MODULE()
   return map;
 }
 
+RCT_EXPORT_VIEW_PROPERTY(isAccessibilityElement, BOOL)
 RCT_REMAP_VIEW_PROPERTY(testID, accessibilityIdentifier, NSString)
 RCT_EXPORT_VIEW_PROPERTY(initialCamera, GMSCameraPosition)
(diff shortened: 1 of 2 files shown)


rborn:
This looks great, any chance to do it for android too (if it's not there already) ? 😻

computerjazz:
Makes sense, I can do that.

The reason I didn't do that here is that Android does not exhibit the same issues as iOS -- when TalkBack is on, markers are still tappable. But I agree exposing the prop cross-platform is the right move.

computerjazz:
although hm -- `isAccessibilityElement` is an iOS-only concept.

rborn:
Ok, in this case can we document the new property  and mention is iOs only?
Also, if this works in Android out of the box, maybe we should set it default to `true` so the users don't really have to do anything to get this, only to be able to disable it if they need it for some reason ?

rborn:
@computerjazz ^^

rborn:
@computerjazz ❤️

PR #3120, opened 2019-10-11
keep current mapType when taking snapshot

r0b0t3d:

No

When taking snapshot, it used default mapType. This PR will keep the current mapType then take snapshot

### How did you test this PR?

Test in iphone simulator


rborn:
Hi, thanks for the PR. Is this an issue on Android too? If yes can you fix it there too? 🤗

r0b0t3d:
@rborn Android is working well, just happens on iOS

PR #3124, opened 2019-10-11
Implement anchor on iOS maps

ssztemberg:

No

There have been many complaints that iOS maps don't have Google Maps `anchor` capabilities. Current approach - using `centerOffset` prop isn't sufficient, because it doesn't calculate actual marker size, which could be dynamic (for example marker including text with different length).

I don't think it's RTM, because it *may* break some current implementations, as some developers may use `anchor` for Google Maps and at the same time, `centerOffset` for iOS maps, which can have different values.

Maybe we should add a flag, for example `useAnchor`. Any thoughts?

### How did you test this PR?

- Create MapView
- Add a few custom markers with anchor
- Test both Google Maps and iOS Maps - on both providers markers placement is the same


rborn:
LGTM 

@christopherdro can you have a look ?🐽  

We could also add a BREAKING warning in the release changelog. As this is making life easier to developers and doesn't imply too much work (just delete some code) I'd push push it tbh. What do you think?

SaeedZhiany:
a

christopherdro:
@RobJacobson  could you resolve the small merge conflict and test these changes ?

Issue #3094, opened 2019-09-28
Calling bridge.imageLoader is deprecated


conor909:
# Bug

After upgrading to RN 61.1 the map displays markers and google logo in bottom corner, but the map tiles / images are not showing.

## Environment info

`"react-native": "0.61.1"`
`"react-native-maps": "0.24.2"`

## Output 

Calling bridge.imageLoader is deprecated and will not work in newer versions of RN. Please update to themoduleForClass API or turboModuleLookupDelegate API.

<img width="375" alt="Screenshot 2019-09-28 at 22 07 42" src="https://user-images.githubusercontent.com/2417359/65821361-f520e680-e23c-11e9-8c6d-28154a13ccc7.png">

PabloGiraudCarrier:
same issue "react-native-maps": "0.25.0"

conor909:
Does anyone know the status of this?  I'v tried installing with the latest commit but I just get a blank map.

<img width="412" alt="Screenshot 2019-10-25 at 10 25 51" src="https://user-images.githubusercontent.com/2417359/67551544-dfa2bd80-f711-11e9-9879-2a1773bc6b46.png">

lopeski:
`        <Marker
        >
          <Image
            source={Img}
          />
        </Marker>`
it works for me

conor909:
@lopeski can you check your versions of react-native?

Akyna:
```
    "react-native": "0.61.5",
    "react-native-maps": "0.26.1",
```
The same...
`Calling bridge.imageLoader is deprecated and will not work in newer versions of RN`

hungnguyentiki:
I have same issue with:
```

"react-native": "0.61.5",
"react-native-maps": "0.26.1",

```

nvuhung:
I have same issue with:

```
"react-native": "0.61.5",
"react-native-maps": "0.26.1",
```

elvispap:
hey guys...do we have any update on this one ? Is there a fix planned to go out ?

vorasudh:
Hello guys, I am using:
`"react-native-maps": "^0.26.1",
"react-native": "0.61.5",`

I am facing the same issue. Though it is just a warning, It gets triggered a lot of times and I wonder if it has effect on performance at later stage (i have not seen any performance issue until now).

altany:
Seems to be fixed in 0.27.0

PR #3125, opened 2019-10-11
replaced deprecated bridge.imageLoader calls with moduleForClass API

brien-crean:

No

#3094 
### How did you test this PR?

I tested on a new react native project running on version 0.61.2 and observed that the deprecation warnings no longer appear after implementing the `moduleForClass` API and tested to make sure that Marker and Map overlay images load correctly after API changes


conor909:
Does anyone know when can we expect this to be released?

PR #3134, opened 2019-10-16
add marker property isAlwaysSelected on iOS

brien-crean:

I couldn't find any PRs that add an always selected option on a marker (iOS). 

It adds a property to the `<Marker />` component called `isAlwaysSelected` which when set to true, allows a user to drag a marker immediately without having to first select it - similar to Android behaviour. This feature is useful when you want the ability to quickly drag markers rather than tap and then drag.

### How did you test this PR?

Affects iOS only.
Tested on the latest version of react native - 0.61.2 on simulator and real device


rborn:
@brien-crean 😻 could you please amend  the docs too ?

Few questions:
- how does it behave on select ? does it work just like before?  are there any drawbacks?
- maybe find a better name 😅? `isPreselected`, `instantDrag` 🤣

brien-crean:
@rborn thanks for the feedback! I just updated the docs and the name to isPreselected 😄

onSelect fires as normal when a marker is tapped for the first time when this is enabled. The only drawback I see is that if you needed to programmatically check whether a marker was selected or not this would always return true.

I have been using this for a while on my fork and I really like it. I can immediately drag a marker without having to focus it first. It's particularly useful when a number of markers are the points of a polygon. I can easily drag markers to reshape the polygon.

PR #3163, opened 2019-10-30
Implement opacity prop for MapOverlay on Android and iOS (Google Maps)

JJStoker:

No.

We needed the ability to control the transparency/opacity of Overlay components. For some reason, the transparency control for GroundOverlay was commented out for [Android](https://github.com/react-native-community/react-native-maps/blob/master/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapOverlay.java#L53) and was never implemented for iOS.

The `transparency` option was implemented as `opacity` as the rest of this project seems to prefer `opacity`, and both platforms implemented it differently as well.

### How did you test this PR?

Tested on an android emulator and iOS simulator using `react-native@0.61.0` and using the branch HEAD `"react-native-maps": "git://github.com/Crop-R/react-native-maps.git#5ee80e8d1e5af9962937d5db578d333fef4d35cb"` as a dependency. We pass an image url to the `image` prop of `MapView.Overlay`.


rborn:
LGTM 

🐽 @christopherdro 

I don't really get why travis is  "yellow" the job seems to be executed with no error 😒

rborn:
@JJStoker thanks for the PR 😻

PR #3235, opened 2020-01-02
fix overlay sizing

bargald:
Yes - PR from https://github.com/react-native-community/react-native-maps/issues/2801
has been merged, but there are still error in it

same as:
https://github.com/react-native-community/react-native-maps/issues/2801

### How did you test this PR?
ran it on iOS and Android deviced


rborn:
❤️

bargald:
@rborn I believe we need to revert this PR, sorry for the inconvenience, from reading the docs now i see that the flipping of the bounds was intentional "(left-top corner, right-bottom corner)". It is wierd that it goes against most geo libraries (first bottom left, than top right). but if it is stated in the docs, this PR goes against it.

Issue #3116, opened 2019-10-08
Need help with project-wide variable values and running example project


zhdanova:
Hi! I'm new to RN development and would like to include react native maps in my first project.

# How can I find values for the project-wide variables?
I did steps from Installation instructions, but got stuck at setting up project-wide variables. My project is created from "react-native init" default template where some variables are already set, but these are not:
 - `supportLibVersion`
 - `playServicesVersion`
 - `androidMapsUtilsVersion`

How can I find their values? I was advised to check included example project, but I wasn't able to run it (see question 2) However, I looked at the source code and didn't find hints on variable values.

# Error when building example project
When trying to run example project I see message `> Task :react-native-maps-lib:compileDebugJavaWithJavac FAILED` and get [following error](https://gist.github.com/zhdanova/a8fc92993901e9f7189cc82b5bc8acfc). Could you please advice how it can be fixed? I'm running the project in Windows OS and building it for Android platform.

Thank you for your patience and time!

yusufdonmez:
this my running project values:

> ext {
>     compileSdkVersion   = 28
>     targetSdkVersion    = 28
>     buildToolsVersion   = "28.0.3"
>     minSdkVersion       = 16
>     supportLibVersion   = "28.0.0"
>     playServicesVersion = "15.0.1" // or set latest version
>     androidMapsUtilsVersion = "0.5+"
> }

zhdanova:
@yusufdonmez, thank you so much for your answer! Could you please tell how did you know values for "supportLibVersion", "playServicesVersion", "androidMapsUtilsVersion"? Are these libraries included somewhere in your project and you checked their versions?

yusufdonmez:
I dont know exactly. those times I search so much things I confuse where I find these values. sorry.

Note: I face a problem with these section

> ext {

> compileSdkVersion = 28
> targetSdkVersion = 28
> buildToolsVersion = "28.0.3"
> minSdkVersion = 16
> supportLibVersion = "28.0.0"
> playServicesVersion = "15.0.1" // or set latest version
> androidMapsUtilsVersion = "0.5+"
> }


and I remove whole the scope. Map running fine now. I advise you a try.

salivanas:
Hey guys. I managed to get the map to work without specifying the `androidMapsUtilsVersion` value at all. May I know what this value means? Because it seems that the module works without it.

cvvspb:
[...]

PR #3316, opened 2020-03-04
Changes to playServicesVersion on installation docs

puremana:
Added a default playServicesVersion and showed how to find latest version in installation docs.

Every time I use this package I have to check what a valid playServicesVersion is otherwise it doesn't compile. I figured I would make it easier for others who have the same issue.

No

#3116 

### How did you test this PR?

I read through it and checked the link works.
As for version 17.0.0, have tested on a real device, simulator, Nexus 2nd Gen, Samsung J5, S5, A30, iPhone XR


rborn:
@puremana 😻

Issue #3282, opened 2020-02-10
error: package com.google.android.gms.location does not exist


dpyeates:
# Bug

 Using the latest package from github, and after following all install instructions (having previous versions working perfectly), I get an error: package com.google.android.gms.location does not exist.

Build fails on line 17 of AirMapManager.java.

This was added in the recent commit "Improved user location on Android (#2993)"

Here is an extract from my app/build.gradle

```
    implementation(project(':react-native-maps')){
        exclude group: 'com.google.android.gms', module: 'play-services-base'
        exclude group: 'com.google.android.gms', module: 'play-services-maps'
    }
    implementation 'com.google.android.gms:play-services-base:17.1.0'
    implementation 'com.google.android.gms:play-services-maps:17.0.0'
    implementation 'com.google.android.gms:play-services-location:17.0.0'
```

Library version: next

rbeer:
Getting the same error trying to run the examples.
Performed steps:
```bash
git clone https://github.com/react-native-community/react-native-maps/
cd react-native-maps/
yarn
yarn run:android
```

Output `yarn run:android`
```
$ ./gradlew installDebug && npm run start:android                                                                                                                                                
Starting a Gradle Daemon (subsequent builds will be faster)
         
> Task :react-native-maps-lib:compileDebugJavaWithJavac                                                                                                                                          
/home/***/react-native-maps/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapManager.java:17: error: package com.google.android.gms.location does not exist
import com.google.android.gms.location.LocationRequest;                                                                                                                                          
                                      ^                                                                                                                                                          
/home/***/react-native-maps/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java:12: error: package androidx.core.view does not exist
import androidx.core.view.GestureDetectorCompat;                                                                                                                                                 
                         ^                                                                                                                                                                       
/home/***/react-native-maps/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java:13: error: package androidx.core.view does not exist                     
import androidx.core.view.MotionEventCompat;                                                                                                                                                     
                         ^                                                                                                                                                                       
/home/***/react-native-maps/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java:73: error: package androidx.core.content does not exist
import static androidx.core.content.PermissionChecker.checkSelfPermission;                                                                                                                       
                                   ^                                                                                                                                                             
[...]

PR #3320, opened 2020-03-06
Fixed fused location source

funkyfourier:

No.

#3282 and #2923

### How did you test this PR?

Set up a test RN project with my fork as a dependency, exported bundled APK to a OnePlus 6 and went outside for a walk. Compared user location continuosly with the Google Maps app.

**Note 1** 

When I try to run *example-android* it fails with the following error. However, when I add my fork to a test RN project it builds fine. I have not succeeded identifying the cause.

```
ERROR: Manifest merger failed : Attribute application@appComponentFactory value=(android.support.v4.app.CoreComponentFactory) from [com.android.support:support-compat:28.0.0] AndroidManifest.xml:22:18-91
	is also present at [androidx.core:core:1.0.0] AndroidManifest.xml:22:18-86 value=(androidx.core.app.CoreComponentFactory).
	Suggestion: add 'tools:replace="android:appComponentFactory"' to <application> element at manifestMerger2158610616111590745.xml:7:5-9:19 to override.
```
**Note 2**

The line `implementation "com.google.android.gms:play-services-location:17.0.0"` needs to have an explicit version number. Using 16.1.0 breaks because of androidx, and version 16.0.0 of play-services-location simply does not exist. I guess it will be sorted out in time, and '17.0.0' can be changed to the safeExtGet value in time.

### DIFF
--- a/lib/android/build.gradle
+++ b/lib/android/build.gradle
@@ -24,8 +24,8 @@ dependencies {
   implementation('com.facebook.react:react-native:+') {
     exclude group: 'com.android.support'
   }
-  implementation "com.google.android.gms:play-services-base:${safeExtGet('playServicesVersion', '16.1.0')}"
-  implementation "com.google.android.gms:play-services-maps:${safeExtGet('playServicesVersion', '16.1.0')}"
+  implementation "com.google.android.gms:play-services-base:${safeExtGet('playServicesVersion', '17.1.0')}"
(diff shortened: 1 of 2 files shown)


rborn:
Thanks for the PR 😻

LGTM 🐽 @christopherdro @alvelig

christopherdro:
Leaving 17.0.0 hardcoded for the location for now because not all play service extensions share the same version.

Issue #3185, opened 2019-11-08
Geojson module not exported?


kylebarron:
# Bug

I ran into the same error as #3033 did with the Heatmap module. I was getting
```
Invariant Violation: Element type is invalid ...
``` 
errors, but when I did
```js
import Geojson from './Geojson';
MapView.Geojson = Geojson
```
as was done in #3033, everything worked.

## Environment info

```bash
info 
  React Native Environment Info:
    System:
      OS: macOS 10.14.6
      CPU: (4) x64 Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz
      Memory: 992.18 MB / 8.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 12.11.1 - /usr/local/bin/node
      Yarn: 1.19.1 - /usr/local/bin/yarn
      npm: 6.13.0 - ~/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 13.1, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
      Android SDK:
        API Levels: 26, 28, 29
        Build Tools: 28.0.3, 29.0.2
        System Images: android-28 | Intel x86 Atom_64, android-29 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.5 AI-191.8026.42.35.5900203
      Xcode: 11.1/11A1027 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: https://github.com/expo/react-native/archive/sdk-35.0.0.tar.gz => 0.59.8 
    npmGlobalPackages:
      react-native-cli: 2.0.1
```

I'm using the latest version of expo for testing (`^35.0.0`) and the `react-native-maps` dependency version is set to `"react-native-maps": "~0.25.0",` in `package.json`.

Library version: ~0.25.0

## Steps To Reproduce

1. Create new project with expo
2. `expo install react-native-maps`.
3. Paste the following code in `App.js`.

```js
import React from "react";
import MapView, { Geojson } from "react-native-maps";
import { StyleSheet, View, Dimensions } from "react-native";
// import Geojson from './Geojson';
// MapView.Geojson = Geojson

const myPlace = {
  type: "FeatureCollection",
  features: [
    {
      type: "Feature",
      properties: {},
      geometry: {
        type: "Point",
[...]

PR #3331, opened 2020-03-11
Geojson component: exported and added types definition

gcoro:

No

#3185 
And also an issue i had personally when importing in a typescript project

### How did you test this PR?

I tested in my project, it's a react native + typescript app, the Geojson component can now be imported without giving any warning/error.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -517,6 +517,21 @@ declare module "react-native-maps" {
   export class Heatmap extends React.Component<MapHeatmapProps, any> {}
 
   // =======================================================================
+  //  Geojson
+  // =======================================================================
+
+  import GeoJSON from 'geojson';
+
+  export interface GeojsonProps {
+    geojson: GeoJSON.GeoJSON;
+    strokeColor?: string;
+    fillColor?: string;
+    strokeWidth?: number;
+  }
+
+  export class Geojson extends React.Component<GeojsonProps, any> {}
+
+  // =======================================================================
   //  Constants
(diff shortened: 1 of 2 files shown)


rborn:
❤️

index.d.ts
```diff
@@ -516,6 +516,21 @@ declare module "react-native-maps" {
 
   export class Heatmap extends React.Component<MapHeatmapProps, any> {}
 
+  // =======================================================================
+  //  Geojson
+  // =======================================================================
+
+  import GeoJSON from 'geojson';
```
SimenB, line 523: needs a dependency on `@types/geojson`, I'm getting errors in CI after this change

image

rborn:
@gcoro can you have a look at that?  Thnx

index.d.ts
gcoro: I can't reproduce this on my project, can you tell me if simply adding `@types/geojson` to `devDependencies` fixes it? Could it also need module `geojson` itself?

SimenB: that would fix it, yeah. I don't use `react-native-maps` at all in my project, I get it via `@types/expo`

SimenB: so I'd like to not add `geojson` either.

If you `import` something, whether it's types or not, you should always have a dependency on it in your own package.json. Maybe a peer dep, but probably a regular dependency

gcoro:
@SimenB @rborn I made another PR to fix it

PR #3342, opened 2020-03-16
Added RCT Convert and CoreLocation libraries

maschad:

No.

As mentioned in the [Lean Core](https://github.com/facebook/react-native/issues/23313) project as well as [here](https://github.com/facebook/react-native/pull/28300#issuecomment-599344895) , the Geolocation libraries are being moved out of react-native core. This would break dependencies used [here](https://github.com/react-native-community/react-native-maps/blob/72b0f77e31d18959e91acb036d5f91ab46e11ef2/lib/ios/AirMaps/AIRMapCircleManager.m#L14) and many other places, thus this PR addresses that.

### How did you test this PR?

This addresses iOS, and given it's an import all previous tests should pass.


shergin:
We also need to change `#import` statement somewhere here in this repo.

maschad:
@shergin we actually have to change it in quite a few files, but I want to know the best way to structure it, currently, I have the `RCTConvert+CoreLocation.h` and `RCTConvert+CoreLocation.m` files in a `Shared` folder and then import them in the respective files from there. What do you think?

rborn:
@maschad @shergin  
How is this influencing the older RN versions (still supported by this module?)
I mean if we merge is a breaking change?

I also see iOS only, android is ok, needs changes, or ?

Thanks 😻

rborn:
@christopherdro ^^

maschad:
Hey @rborn so this came to my awareness because of an [issue](https://github.com/facebook/react-native/issues/20879#issuecomment-592505530) where iOS was still requiring users to update their `Info.plist` despite them not using any geolocation libraries, that was further explained [here](https://github.com/facebook/react-native/issues/28200). There is a breaking change on the [react-native](https://github.com/facebook/react-native) repo but I don't believe it would be here if this was merged, given the _import_ would now be from a library that is in the repo.

rborn:
@maschad but we would need to "force" the usage of the latest RN version  right? (or the one using shared ) or it will be done by RN  ? (Sorry, my objC-foo is minimal 😅)

maschad:
@rborn no you wouldn't need to as React Native would still have the libraries there until [this](https://github.com/facebook/react-native/pull/28326) is merged. I am not strong in Objective-C either so I would like some guidance on how I could test this PR to ensure I am importing the files correctly. But if that does work, then this should be fine to merge as these changes add the requisite files.

rborn:
@maschad great, thank you, maybe @shergin can add some insight? 😻

shergin:
I don't know how to advice in which folder to put new files because I am not a maintainer of the repo so I have no context.

This change is probably already backward-compatible (because it removes a dependency, not introduces it) but to make it even more safe (just in case we don't want to bother with testing) we can just rename the methods to eliminate the possibility of ODR-like violations.

salah-ghanim:
Doing any changes to this libs dependencies that are shared with RN will limit the users of the library to use a matching RN library, this is common in the RN environment and ideally we always support the latest version of RN out of the box.

highlighting RN version compatibility can be done in the release  note

Issue #3344, opened 2020-03-17
Crash with null Location in FusedLocationSource


netshade:
# Bug

In some cases where `getLastLocation` returns `null`, the `null` location provided in the success listener is incorrectly passed to `onLocationChanged`. This then causes a crash later downstream, as seen in this stack trace from a Detox run:

```
   androidx.test.espresso.PerformException: Error performing 'com.wix.detox.espresso.action.detoxsingletap@1305bed click - At Coordinates: 521, 1250 and precision: 16, 16' on view '(with tag value: is "NewReport-Next" and view has effective visibility=VISIBLE)'.
    	at androidx.test.espresso.PerformException$Builder.build(PerformException.java:86)
    	at androidx.test.espresso.base.DefaultFailureHandler.getUserFriendlyError(DefaultFailureHandler.java:87)
    	at androidx.test.espresso.base.DefaultFailureHandler.handle(DefaultFailureHandler.java:59)
    	at androidx.test.espresso.ViewInteraction.waitForAndHandleInteractionResults(ViewInteraction.java:322)
    	at androidx.test.espresso.ViewInteraction.desugaredPerform(ViewInteraction.java:178)
    	at androidx.test.espresso.ViewInteraction.perform(ViewInteraction.java:119)
    	at com.wix.detox.espresso.EspressoDetox.perform(EspressoDetox.java:40)
    	at java.lang.reflect.Method.invoke(Native Method)
    	at org.apache.commons.lang3.reflect.MethodUtils.invokeStaticMethod(MethodUtils.java:443)
    	at org.apache.commons.lang3.reflect.MethodUtils.invokeStaticMethod(MethodUtils.java:405)
    	at com.wix.invoke.types.ClassTarget.execute(ClassTarget.java:23)
    	at com.wix.invoke.types.Target.invoke(Target.java:59)
    	at com.wix.invoke.MethodInvocation.invoke(MethodInvocation.java:35)
    	at com.wix.invoke.MethodInvocation.invoke(MethodInvocation.java:26)
    	at com.wix.invoke.MethodInvocation.invoke(MethodInvocation.java:20)
    	at com.wix.detox.InvokeActionHandler.handle(DetoxActionHandlers.kt:52)
    	at com.wix.detox.DetoxManager$4.run(DetoxManager.java:121)
    	at android.os.Handler.handleCallback(Handler.java:883)
    	at android.os.Handler.dispatchMessage(Handler.java:100)
[...]

PR #3345, opened 2020-03-17
Protect against null Location in FusedLocationSource

netshade:

No

#3344 

### How did you test this PR?

Tested on Android Emulator w/ Detox on a CI server where this was previously occurring. The crash does not recur.


rborn:
@netshade 😻

LGTM @christopherdro 🐽

PR #3358, opened 2020-03-25
Pass camera move's reason to JS callbacks

yairopro:

There is a PR that solves likely the problem #2935, but with a workaround.
The problem we are trying to solve is to know the reason when the region has changed on the map.
Using `onPanDrag`, implemented by the PR #2935, is a non stable workaround. Because will would have to first store somewhere a boolean on the `onPanDrag`, and use it once `onRegionChangeComplete`.

Issue #1620 and #2756.

### How did you test this PR?

I implemented and tested (on a real device) the code for android. The iOS part too has been implemented by my friend. And was tested on an emulator.

### DIFF
--- a/docs/mapview.md
+++ b/docs/mapview.md
@@ -56,8 +56,8 @@ To access event data, you will need to use `e.nativeEvent`. For example, `onPres
 |---|---|---|
 | `onMapReady` |  | Callback that is called once the map is fully loaded.
 | `onKmlReady` | `KmlContainer` | Callback that is called once the kml is fully loaded.
-| `onRegionChange` | `Region` | Callback that is called continuously when the region changes, such as when a user is dragging the map.
-| `onRegionChangeComplete` | `Region` | Callback that is called once when the region changes, such as when the user is done moving the map.
+| `onRegionChange` | (`Region`, `{isGesture: boolean}`) | Callback that is called continuously when the region changes, such as when a user is dragging the map. The second parameter is an object containing more details about the move. `isGesture` property indicates if the move was from the user (true) or an animation (false).
+| `onRegionChangeComplete` | (`Region`, `{isGesture: boolean}`) | Callback that is called once when the region changes, such as when the user is done moving the map.  The second parameter is an object containing more details about the move. `isGesture` property indicates if the move was from the user (true) or an animation (false).
 | `onUserLocationChange` | `{ coordinate: Location }` | Callback that is called when the underlying map figures our users current location (coordinate also includes isFromMockProvider value for Android API 18 and above). Make sure **showsUserLocation** is set to *true*.
 | `onPress` | `{ coordinate: LatLng, position: Point }` | Callback that is called when user taps on the map.
 | `onDoublePress` | `{ coordinate: LatLng, position: Point }` | Callback that is called when user double taps on the map.
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
@@ -349,8 +349,12 @@ public class AirMapView extends MapView implements GoogleMap.InfoWindowAdapter,
       @Override
       public void onCameraMove() {
         LatLngBounds bounds = map.getProjection().getVisibleRegion().latLngBounds;
+
         cameraLastIdleBounds = null;
-        eventDispatcher.dispatchEvent(new RegionChangeEvent(getId(), bounds, cameraMoveReason, true));
+        boolean isGesture = GoogleMap.OnCameraMoveStartedListener.REASON_GESTURE == cameraMoveReason;
+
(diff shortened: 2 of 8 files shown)


yairopro:
To pass [the enum value ](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.OnCameraMoveStartedListener#REASON_API_ANIMATION) to the JS callbacks `onRegionChange` & `onRegionChangeComplete`, I first added the property `reason` to the android `RegionChangeEvent`. And once the event received on the JS thread, I pass a 2nd object parameter, called `details`, to the callback, like `onRegionChange(region, details)`. I build the `details` object before calling the callback. 
For now, `details` has only one property `reason`. It may have other data, which will not be necessary defined in the `Region` object,  in the future.

yairopro:
Because [on iOS the map gives only a boolean, if it's a user gesture or not](https://developers.google.com/maps/documentation/ios-sdk/reference/protocol_g_m_s_map_view_delegate-p#ae1f53e0d056bf4d246356134ec594612), then I adapted the same behaviour for android (it converts the integer to the same boolean) and renamed the property from `reason` to `isGesture`.

rborn:
LGTM @christopherdro 🐽

@yairopro 😻

rborn:
@yairopro I know this is a very old PR 😅 but could you add the extra info in the docs too (the events seem to be missing the `details` param ?
Then I'll merge 🙂

yairopro:
Oh you're right. I've forgotten the documentation.
Let me some time to re-read the PR code.

yairopro:
@rborn  I updated the documentation.
You can edit it too if you wish.

rborn:
😻

yairopro:
🥳 Thanks

rennehir:
Nice feature, this would really help us a lot with our product!

But I noticed that there hasn't been a release since March. @rborn @yairopro do you have any estimates when that might occur? Or should I just point my package.json to a specific commit?

Cheers, great work! 🥳

yairopro:
Hi @rennehir 
I can't say, I am just an outside contributor.

juice49:
This doesn't seem to be working with Apple Maps on iOS. I logged the [`nativeEvent` variable in `_onChange`](https://github.com/react-native-maps/react-native-maps/blob/master/lib/components/MapView.js#L641), and the `isGesture` property is missing:

```
{
  "region": {
    "longitude": -2.7246242753611227,
    "latitude": 54.88447300842492,
    "latitudeDelta": 4.24778519660066,
    "longitudeDelta": 4.548340505366269
  },
  "target": 989,
  "continuous": true
}
```

I get the property as expected on Android, for example:

```
{
  "region": {
    "longitudeDelta": 4.520089812576771,
    "longitude": -2.9489495046436787,
    "latitudeDelta": 3.685225103923095,
    "latitude": 55.22018295561497
  },
  "isGesture": true,
  "continuous": true
}
```

I haven't tested this with Google Maps on iOS.

yairopro:
Hi @juice49 
The flag comes only with google map (iOS only).
I should have say that in the doc, my bad.

juice49:
Thanks @yairopro! I wasn't sure if this should work or not.

rborn:
@juice49 mind a PR with the doc change ? 😍

juice49:
@rborn Absolutely! PRd in #3589. Thank you for your work maintaining this excellent library 🙂.

rborn:
@juice49 merged 😻

juice49:
Thanks @rborn!

RyanTG:
Sorry - earlier today I commented on the commit rather than the issue. I thus deleted the comment, and here it is in a more appropriate place:

Just in case the developers of this commit missed it, some of us are not able to get this to work in version 0.28.0. isGesture is undefined. Any guidance?
https://github.com/react-native-maps/react-native-maps/issues/3818

christopherdro:
Looking through the source it appears the `isGesture` property is only available on google maps. If you are using Apple Maps then the value will be undefined.

RyanTG:
We are using Google Maps on both iOS and Android, and version 0.28.0. It's unclear if we're doing something wrong (and it's working for others) or if there is an issue with the feature.

RyanTG:
@yairopro I just want to say thanks for the `isGesture` feature. Looking back at the issues on this repo, it's been brought up at least half a dozen times over the past few years - so it clearly solves a much-needed issue! It works great, and it saved our butts.

It would be cool to have this functionality on Apple Maps, too. Is there any chance of that? In our particular use case, now that an upcoming version of react-native-maps is going to include a manual theme toggle for Apple Maps, there is feature parity between the two (aside from isGesture).

yairopro:
Hi @RyanTG 
😁 I'm really happy to hear that my work helps. And I really thank you for this warming message.

But unfortunately, I am not able to help you since I am not an iOS developer. sorry
The iOS code in this PR has been added by a colegue

PR #3397, opened 2020-04-20
docs: default value to `userLocationFastestInterval` prop

karimcambridge:

No, not as far as I can tell.

Improves documentation for the `MapView` component.

The default value can be found here https://github.com/react-native-community/react-native-maps/blob/bf032461adc0d843bab60e3c55b226629ca12e74/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapManager.java#L173

Why?

userLocationUpdateInterval which is very similar has it's default value in the docs. This would help new users also.

### How did you test this PR?

No test required.


rborn:
Thank you 😻

PR #3438, opened 2020-05-18
fix overlay bounds confliction from array

isener:

**NO**

Fix for the overlay bounds array conflicts eg: **southern latitude exceeds northern latitude** caused if you put BottomLeft and TopRight coordinates into array. (iOS was already working. This fix only for ANDROID)

(please link the issue here)

### How did you test this PR?

I tested in both Genymotion Emulator;
Android v9.0
ARMv64

and my device running MIUI;
Android v9.0


rborn:
😻

PR #3484, opened 2020-06-26
Added 'useNativeDriver: false' to Animated.timing to suppress warnings in React Native v0.62

liamlows:

No

When using React Native v0.62, a consistent warning will generate when using an animated region and animated marker. The warning (seen below) can be suppressed on iOS by setting the option `useNativeDriver: false` in `AnimatedRegion.js`.

[Issue](https://github.com/react-native-community/react-native-maps/issues/3483)

Warning: `Animated: 'useNativeDriver' was not specified. This is a required option and must be explicitly set to 'true' or 'false'`

### How did you test this PR?

I created an animated marker and then used an animated region when generating the coordinate. I then ran the build on my iPhone 11 Pro Max using xcode on iOS 13.5. The warnings that used to pop up no longer popped up.

### DIFF
--- a/lib/components/AnimatedRegion.js
+++ b/lib/components/AnimatedRegion.js
@@ -162,6 +162,7 @@ export default class AnimatedMapRegion extends AnimatedWithChildren {
         Animated.timing(this.latitude, {
           ...config,
           toValue: config.latitude,
+          useNativeDriver: false
         })
       );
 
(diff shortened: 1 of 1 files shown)


rborn:
😻

EhsanSarshar:
What's the point of setting it to false. only to suppress the warning. Marker.Animated cause a lot of lag in IOS.

PR #3516, opened 2020-07-31
Added polyline props to obtain dashed/dotted lines in Geojson component

Yel42:

No

While the Polyline component already had props for rendering dashed and dotted lines, the Geojson component, despite using Polyline component, didn't have those. So I provided these props (lineDashPhase, lineDashPattern, lineCap, lineJoin and miterLimit) to Geojson.js, so that they can be accessible from the component itself.

### How did you test this PR?

This change has been tested on both Android and iOS, using Google Maps API on an LG Nexus 5 real device (android 6.0.1), and using Apple maps API on an iPhone simulated device, running iOS 13.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -547,6 +547,11 @@ declare module "react-native-maps" {
     strokeColor?: string;
     fillColor?: string;
     strokeWidth?: number;
+    lineDashPhase?: number;
+    lineDashPattern?:number[];
+    lineCap?:'butt'|'round' | 'square';
+    lineJoin?: 'miter'| 'round' | 'bevel';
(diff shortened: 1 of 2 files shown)


christopherdro (approved):

elsurudo:
Thanks @Yel42 !

Code looks really simple, but for some reason I can't get a dotted line to show up (I am on 0.28.0, which includes this change).

Any reason what I might be missing?

```
<Geojson key={`outline-${index}`}
                         geojson={mapFeatureGeojson}
                         strokeColor='green'
                         strokeWidth={2}
                         lineDashPattern={[10, 10]} />
```

PR #3551, opened 2020-09-18
fix: Xcode 12 compatibility

Fetten:
No.

The latest Xcode 12 fails to build when a module does not depend on `React-Core` directly. This change is necessary for all native modules on iOS. For details please see: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

### How did you test this PR?
1. Build an App with Xcode 11.x. Result: The app builds.
2. Build an App with Xcode 12. Result: The app will not build when depending on `React`.
3. Build an App with Xcode 12 with the PR changes applied. Result: The app builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: node_js
 
+install:
+  - yarn
+
 node_js:
-  - "lts/*"
+  - 'lts/*'
 
 cache:
   directories:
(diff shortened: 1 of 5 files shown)


rborn:
LGTM @christopherdro 🐽

rborn:
@Fetten could you please sync and one travis passes I'll merge ? 😻

rborn:
😻

PR #3570, opened 2020-10-01
Remove Travis.yml

christopherdro:
Move to github actions


rborn:
I believe we'll need a mac machine later on if we ever start building for real ? Is this something GH offers? (sorry, not familiar with actions)

PR #3577, opened 2020-10-05
Add key prop to <Marker /> docs

bzrncev:


docs/marker.md
```diff
@@ -23,6 +23,7 @@
 | `stopPropagation` | `Boolean` | false | Sets whether this marker should propagate `onPress` events. Enabling it will stop the parent `MapView`'s `onPress` from being called. **Note**: iOS only. Android does not propagate `onPress` events. See [#1132](https://github.com/react-community/react-native-maps/issues/1132) for more information.
 | `opacity` | `Float` | 1.0 | The marker's opacity between 0.0 and 1.0.
 | `isPreselected` | `Boolean` | false | When true, the marker will be pre-selected. Setting this to true allows the user to drag the marker without needing to tap on it once to focus on it. **Note**: iOS Apple Maps only.
+| `key` | `String` | | By default in iOS, when the position of the markers is changed, they move to the new position with animation. If you want to disable the animation, add a `key` prop with a unique value like `key_${item.longitude}_${item.latitude}`.
```
rborn, line 26: Thanks a lot for your PR 
I'd change the explanation saying something like  "the key is reusing the component instead of creating a new one and therefore there is an animation - default in ios", your comment implies the key is only disabling animation 🤗

rborn (changes requested):

docs/marker.md
bzrncev: Thanks for the note, the description is changed. 👍

rborn:
😻

PR #3579, opened 2020-10-13
TypeScript Fix: Renamed invalid export type MapViewAnimated to Animated

aajahid:

No

It allow typescript to import the `Animated` export. Without it - anyone using typescript won't be able to import the `Animated` export.

### How did you test this PR?

As this change is only applied on the typescript definition - no test is necessary on the functionality of this plugin. 
But you can test the typescript support by simply importing the `Animated` export and transpile using TSC command. Which will allow the typescript import without any issue. 

Tested on latest stable Typescript 4.0.3. Also tested the code working by running the app on a Android device on debug mode.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -298,7 +298,7 @@ declare module "react-native-maps" {
     setIndoorActiveLevelIndex(index:number): void;
   }
 
-  export class MapViewAnimated extends MapView {}
+  export class Animated extends MapView {}
 
   // =======================================================================
   //  Marker


christopherdro:
This was recently changed https://github.com/react-native-maps/react-native-maps/commit/acd5ebafc8f063763e24e271db6d4ad3cd7b26da

@aajahid What's the discrepancy here?

aajahid:
@christopherdro - I assume `MapViewAnimated` was renamed into `Animated`. 
When I tried to use `MapViewAnimated` ( my IDE suggested there is something available based on the typescript ) - In run time I found there no such thing exported from the `react-native-maps` library. 

Hence I renamed it to `Animated` and now I can use the `Animated` export.

Issue #3491, opened 2020-07-03
Add support for the zIndex props in Geojson component.


anusstpl:
# Feature request

Add support for the zIndex props in the Geojson component.

## Why it is needed

While implementing both WMSTile and Geojson component the WMSTiles overlaps on the geoJSON and completely covers the Geojson. If the Geojson has the zIndex prop it will be easily managed.

## Possible implementation

I have checked the native code and they are using the zIndex internally but the prop is not exposed to the JS side. So please anyone add this support. 

I have fixed this issue by a workaround. I have provided the zIndex prop in the lib/components/Geojson.js file with the value relatively higher value. Please look at the code below of the same file. 

### Code sample

```
import React from 'react';
import MapView from './MapView';

export const makeOverlays = features => {
  const points = features
    .filter(
      f =>
        f.geometry &&
        (f.geometry.type === 'Point' || f.geometry.type === 'MultiPoint')
    )
    .map(feature =>
      makeCoordinates(feature).map(coordinates =>
        makeOverlay(coordinates, feature)
      )
    )
    .reduce(flatten, [])
    .map(overlay => ({ ...overlay, type: 'point' }));

  const lines = features
    .filter(
      f =>
        f.geometry &&
        (f.geometry.type === 'LineString' ||
          f.geometry.type === 'MultiLineString')
    )
    .map(feature =>
      makeCoordinates(feature).map(coordinates =>
        makeOverlay(coordinates, feature)
      )
    )
    .reduce(flatten, [])
    .map(overlay => ({ ...overlay, type: 'polyline' }));

  const multipolygons = features
    .filter(f => f.geometry && f.geometry.type === 'MultiPolygon')
    .map(feature =>
      makeCoordinates(feature).map(coordinates =>
        makeOverlay(coordinates, feature)
      )
    )
    .reduce(flatten, []);

  const polygons = features
    .filter(f => f.geometry && f.geometry.type === 'Polygon')
    .map(feature => makeOverlay(makeCoordinates(feature), feature))
    .reduce(flatten, [])
    .concat(multipolygons)
    .map(overlay => ({ ...overlay, type: 'polygon' }));

  return points.concat(lines).concat(polygons);
};

const flatten = (prev, curr) => prev.concat(curr);

const makeOverlay = (coordinates, feature) => {
  let overlay = {
    feature,
  };
  if (
    feature.geometry.type === 'Polygon' ||
    feature.geometry.type === 'MultiPolygon'
  ) {
[...]

PR #3594, opened 2020-10-27
Added zIndex prop on Geojson

WrathChaos:

No, there is no another PR to add this functionality.

#3491 

### How did you test this PR?

- Tested on both Android and iOS Platforms with Google API key.
- I've tested this on both in real device and a simulator.

This simply adds the `zIndex` prop for the `Geojson` component. It lets developers to use multiple Geojson into each other.


christopherdro:
Thank you for the PR. Can you possibly add this prop to the docs here? https://github.com/react-native-maps/react-native-maps/blob/master/docs/geojson.md

WrathChaos:
@christopherdro
Added it, if you want to change the description please write it down, I can change it now :)

PR #3644, opened 2020-12-03
update README.md

Hina-softwareEngineer:

(please answer here)
There are not such  PR.

Replace `marker.latlng` with object `{ latitude , longitude }` in two different code snippets of marker example. Without this, the code looks ambiguous and the meaning of the `marker.latlng` variable remains unclear for beginners. May be beginners assume it an object `{ lat , lng }` instead of `{ latitude , longitude }`.

### How did you test this PR?

(please answer here)
I have changed in the readme.md file. A documentaion change. So, there is no need to test through any code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ import { Marker } from 'react-native-maps';
 ### Rendering a Marker with a custom view
 
 ```jsx
-<Marker coordinate={marker.latlng}>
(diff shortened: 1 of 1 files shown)


christopherdro (approved):

Bilal112:
Great

Issue #3653, opened 2020-12-15
App crash after double top on map on the device without gms


jebai0521:
# Bug report

## Summary

App crash after double top on map, while map not initialized, gms not installed on the device

## Environment info

`react-native info` output:

```bash
react-native: 0.63.4
react: 16.13.1
react-native-maps: 0.27.1

Target Platform:  Android 9
```

## Steps to reproduce

1. show a map view on a device without gms
2. double tap map view

Describe what you expected to happen:

1. no any response

Describe what you actually happens:

1. app crashed

PR #3654, opened 2020-12-15
fix crash of double tap

jebai0521:
fix crash when double tap on a not initialized map view

No.

#3653 

### How did you test this PR?

return if map is null in doubeTap's handler


bahtiyarerden:
Could you merge this PR with master please?

Issue #3057, opened 2019-09-09
Polygon fill color not working properly 


NemanjaManot:
# Bug

I am using Google Maps on iOS and I have Polygons. Before update (_to version 0.18.3. - at the moment I am not able to update to latest version_) everything works properly, but from now fill color gets weird results.
**Sometimes color is ok, sometimes it is not proper**, no rules.
On android everything works well.

## Environment info

React native info output:

```bash
    Binaries:
      Node: 10.16.0 - /usr/local/bin/node
      npm: 6.9.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
      Android SDK:
        API Levels: 26, 28, 29
        Build Tools: 28.0.3, 29.0.1
        System Images: android-28 | Intel x86 Atom_64, android-28 | Google Play Intel x86 Atom, android-29 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5692245
      Xcode: 10.3/10G8 - /usr/bin/xcodebuild
    npmPackages:
      react: ^16.9.0 => 16.9.0 
      react-native: 0.59.10 => 0.59.10 
    npmGlobalPackages:
      react-native-cli: 2.0.1
```

Library version: 0.18.3

## Steps To Reproduce

`1. react-native run-ios`
...

Describe what you expected to happen:

1. Polygon with proper color

## Reproducible sample code

```
export const Polygon = (props) => {
    return (
        <MapView.Polygon
            coordinates={ props.selectedAreas }
            fillColor={ props.fillColor }
            strokeColor={ props.strokeColor }
        />
    )
};

```

wamry:
i just tried circle, it has the same issue

tomasswood:
I had the same issue using **rgba**. Converting to HEX with Alpha worked. Eg `#3F51B570`

kyoyadmoon:
I have the same issue when using Circle component after upgrade react-native-maps to 0.25
I downgrade react-native-maps version to 0.24.2 and it works fine.

Tautotes:
Same issue here

elicleveland12:
[...]

Issue #2867, opened 2019-05-20
Polygon styling props ignored for Google Maps on iOS


benjamin-sweney:
### Is this a bug report?

Yes

### Have you read the [Installation Instructions](https://github.com/react-community/react-native-maps/blob/master/docs/installation.md)?

Yes

### Environment

react-native: >= 0.51.0
react: >= 16.0.0
react-native-maps: 0.24.0

### Steps to Reproduce

1. Download repository
2. Launch example project for iOS 
3. Switch "Use GoolgeMaps?" to on
4. Go to "Update native props"

### Expected Behavior

The shapes would respect the props (fillColor, strokeWidth, strokeColor etc.) of polygons, circles etc. 

### Actual Behavior

The circle has a black stroke and no fill and the polygon has a blue fill and no stroke. 

### Reproducible Demo

The example project in the repository can be used.

<img width="500" alt="Screen Shot 2019-05-20 at 6 34 44 PM" src="https://user-images.githubusercontent.com/32251097/58000803-f96e2200-7b2d-11e9-8d0b-865dda9959be.png">

benjamin-halter:
Note that using `setNatvieProps` on a timeout on mount eg. `setTimeout(() => ref.setNativeProps(this.props), 1000)` will result in the correct behaviour.

rborn:
Not sure but this might be an outdated example?
I'm doing this and it works fine:

~~~
<Polygon
                style={styles.polygon}
                coordinates={this.props.coordinates}
                holes={this.props.holes}
                strokeColor={strokeColor}
                fillColor={fillColor}
/>

Maybe I don't get the issue?
~~~

benjamin-halter:
@rborn strange... Would it be possible for you to check the example project and get that working? I have the exact same issue in my project as I've shown above. Works fine for everything except GoogleMaps on iOS.

rborn:
@benjamin-halter I created a brand new project and all seems to work fine which makes me thing is an issue with the examples or maybe your settings?

https://github.com/rborn/rn_maps_polygons_test

![Screenshot 2019-05-21 at 11 27 25](https://user-images.githubusercontent.com/406409/58085376-e8202680-7bbc-11e9-8c80-744dad37a343.png)

benjamin-halter:
@rborn thanks for that example! I can confirm this works - I noticed that you used pods to install, I however followed the manual instructions - I will try install with pods and report back the results

benjamin-halter:
[...]

PR #3658, opened 2020-12-17
Geojson stroke color, stroke width, fill color, and marker color properties preserved

ddarren:

No

The GeoJson feature of this library does not preserve the stroke width, stroke color, fill color, and marker color properties from the GeoJson supplied. This PR now preserves all those properties and successfully displays map elements with those styling properties. 

While these two issues may not be specifically about GeoJson, much of comments are about the GeoJson styling issue:
#2867 #3057

### How did you test this PR?

This PR was tested with Android and iOS simulators, although the same logic was tested in a patch on real iOS and Android devices. 

Using a GeoJson creator like [https://geojson.io/](https://geojson.io/) - it is easy to create map elements with properties in the GeoJson. From there, it's just a matter of firing up an app that uses the GeoJson you created in a map view.

### DIFF
--- a/docs/geojson.md
+++ b/docs/geojson.md
@@ -5,9 +5,9 @@
 | Prop      | Type | Default                                                | Note |
 | --------- | ---- | ------------------------------------------------------ | ---- |
 | `geojson` |  `GeoJSON`    |  | [Geojson](https://geojson.org/) description of object. |
-| `strokeColor` | `String`     | `#000` | The stroke color to use for the path.  |
-| `fillColor` | `String`     |  | The fill color to use for the path. |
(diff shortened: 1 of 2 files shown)


octagonalsquare:
I tested by just pasting your changes into the Geojson.js file and it seems to have worked. I haven't noticed any more issues. Would love for this to be merged.

ddarren:
@octagonalsquare Was it a polygon by chance? If a polygon overlaps, on Android this can cause the fill color to not come through. I think I'll need the geojson to debug.

rborn (approved):

rborn:
@octagonalsquare please open a new issue if you still have the fill problem

Issue #3631, opened 2020-11-24
onRegionChangeComplete on MapView doesn not have {isGesture} on v0.27.1


Infinitism:
# Bug report

onRegionChangeComplete on MapView doesn not have {isGesture}

## Summary

On version 0.27.1 does not have the {isGesture} property in the downloaded package but lists it in the MapView documentation.

## Environment info

react-native-maps: 0.27.1

Ahmdrza:
Hi, any idea when this will be available in the latest version?

KyleAsaff:
doesn't work for me either even when `provider={PROVIDER_GOOGLE}`. I need to detect when a user pans by dragging. any work arounds or ETA on a fix?

anaumov-work:
@Ahmdrza @KyleAsaff @Infinitism   You can use my workaround - onPanDrag event with a flag variable to detect gestures

KyleAsaff:
> @Ahmdrza @KyleAsaff @Infinitism You can use my workaround - onPanDrag event with a flag variable to detect gestures

I tried to `useState` to set a flag for `onPanDrag` to determine if it's a pan action in `onRegionChangeComplete` then setting it false after `onRegionChangeComplete` completes. it's unreliable - sometimes a `onPanDrag` will trigger an extra time after `onRegionChangeComplete` triggers.

krtejeda:
Not a bug. isGesture is available in the latest branch and can be used if you use 0.28.0. Only thing missing is the type definitions which will be addressed by #3677 .

iners-max:
I have put 0.28.0 in my package.json (and verified in package-lock.json) but the second parameter is undefined

dazulu:
> I have put 0.28.0 in my package.json (and verified in package-lock.json) but the second parameter is undefined

I am also getting back`undefined` for `isGesture` on 0.28.0 (context is onRegionChangeComplete)

![Screenshot 2021-05-10 at 11 30 48](https://user-images.githubusercontent.com/4543447/117638498-2c7aa780-b183-11eb-9e1e-00b4ec49aa9d.png)

rxng8:
it is still unusable in 0.28.0.

vietnt-uet:
@christopherdro This is still reproduced on the latest version.

tomgreener:
Also experiencing this issue in 0.28.0

Seypopi:
I am on 0.28.0. and it is still impossible to use `{isGesture}`

[...]

PR #3677, opened 2021-01-06
Fixed missing typing

Goldziher:

No previous PR

It fixes a missing typing. The onRegionChange and onRegionChangeComplete event handlers receive a second parameter, which appears in the docs and the source code but not in the typings.

### How did you test this PR?

N/A - I read the docs, saw the typing was gone, looked at the source code, saw the docs were ok but the typings were not and fixed these. 

Just code review would be sufficient.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -239,8 +239,8 @@ declare module "react-native-maps" {
 
     onMapReady?: () => void;
     onKmlReady?: (values: KmlMapEvent) => void;
-    onRegionChange?: (region: Region) => void;
-    onRegionChangeComplete?: (region: Region) => void;
+    onRegionChange?: (region: Region, details: { isGesture: boolean }) => void;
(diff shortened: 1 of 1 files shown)


KyleAsaff:
I tested this out and, unfortunately, this doesn't fix this issue for `onRegionChangeComplete`:
https://github.com/react-native-maps/react-native-maps/issues/3631

it gives me:

```
Object {
  "isGesture": undefined,
}

```

I suspect the typing and the issue above should be fixed in the same PR

rborn:
@Goldziher willing to update with the fix for @KyleAsaff ? 🤗

christopherdro:
@KyleAsaff Please make sure you are testing v0.28.0 that includes the changes to include `isGesture`.

christopherdro (approved):

PR #3702, opened 2021-02-03
Add edgePadding support to fitToElements method.

arneson:

There is no previous PR which does the same thing. 

The method fitToElements does not support edge padding. 
It should have the same api as fitToSuppliedMarkers and fitToCoordinates.

### How did you test this PR?

We have tested the changes on iOS and Android using physical devices on both platforms. 

These changes only affect Google Maps and not Apple Maps, as I've also clarified in the readme. 

You can test the new functionality by passing edgePadding to the fitToElements function and see how the markers are no longer aligned with the very edge of the map. Instead it uses the provided padding.

### DIFF
--- a/docs/mapview.md
+++ b/docs/mapview.md
@@ -92,7 +92,7 @@ To access event data, you will need to use `e.nativeEvent`. For example, `onPres
 | `getMapBoundaries` | | `Promise<{northEast: LatLng, southWest: LatLng}>`
 | `setMapBoundaries` | `northEast: LatLng`, `southWest: LatLng` | The boundary is defined by the map's center coordinates, not the device's viewport itself. **Note:** Google Maps only.
 | `setIndoorActiveLevelIndex` | `levelIndex: Number` |
-| `fitToElements` | `animated: Boolean` |
+| `fitToElements` | `options: { edgePadding: EdgePadding, animated: Boolean }` | **Note** edgePadding is Google Maps only
 | `fitToSuppliedMarkers` | `markerIDs: String[], options: { edgePadding: EdgePadding, animated: Boolean }` | If you need to use this in `ComponentDidMount`, make sure you put it in a timeout or it will cause performance problems. **Note** edgePadding is Google Maps only
 | `fitToCoordinates` | `coordinates: Array<LatLng>, options: { edgePadding: EdgePadding, animated: Boolean }` | If called in `ComponentDidMount` in android, it will cause an exception. It is recommended to call it from the MapView `onLayout` event.
 | `pointForCoordinate` | `coordinate: LatLng` | Converts a map coordinate to a view coordinate (`Point`). Returns a `Promise<Point>`.
--- a/example/examples/MapKml.js
+++ b/example/examples/MapKml.js
@@ -28,7 +28,7 @@ class MapKml extends React.Component {
   }
 
   onKmlReady() {
-    this.map.fitToElements(true);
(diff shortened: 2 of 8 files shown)


rborn:
@arneson thanks a lot for this 😻 it seems linter doesn't fully like some spaces there, could you please fix that?
Screenshot 2021-02-03 at 16 14 52

arneson:
For sure! I've removed the spaces now. Hopefully the build passes.

arneson:
I'd forgotten to update the example project, so I added a commit doing that. 
Everything should be OK now. @rborn

rborn:
almost there (some conflict)  😅

arneson:
@rborn 
Yeah, sorry about that. My other PR got merged the very second I tagged you.

arneson:
@rborn 
We should be all good now :)

ravirajn22:
LGTM. 

Note: It's a breaking change we have to release a new major version.

ravirajn22 (approved):

salah-ghanim:
what's needed here to merge this? the change looks clean, does it need more QA?

RobJacobson:
I seem to have been added as a reviewer (not sure how/why),  I'm just getting started with this project and I'm not in a position to give quality QA comments/advice.  Sorry!  If I'm the holdup, could you find another reviewer?

arneson:
@christopherdro

salah-ghanim:
I realized, that instead of introducing a breaking change, a new API would make sense in this case.

simply fitToElements without options and another with the options, that way anyone updating don't have to change their method signiture, can you please implement that @arneson arneson

arneson:
Sure @salah-ghanim, I can take a look. 
To clarify, you'd like the fitToElements function to accept either a boolean or an object as its only parameter?

arneson:
@salah-ghanim 

Looking at it I realised that it would mean that fitToElements would have a different api than fitToSuppliedMarkers, fitToCoordinates etc. That doesn't seem like something that would make sense. 

I'd argue that my original (breaking) change makes the most sense.

salah-ghanim:
@arneson yes agree with you since this will be more consistent with the other apis.
did you test on iOS by any chance? I will test it this week and merge both ways if it looks good.

ravirajn22:
This PR LGTM. Instead of wasting time and making this PR stale, @rborn and @christopherdro I think we can merge this PR and in release notes just add Breaking change.
Also its not a big breaking change, just very minor.

rborn:
@christopherdro thoughts on @ravirajn22 proposal? I kind of agree tbh

christopherdro:
I personally don't see the discussion around this PR as wasted time. This is part of the code review process and discussion should be encouraged. I don't see any harm in having @salah-ghanim doing another review and test before we merge this in.

@arneson There were some similar changes recently merged. https://github.com/react-native-maps/react-native-maps/commit/4741b607fe06b82b24a7b04fc28441dc592944ce Could you take a look at those changes and see if they would apply here?

arneson:
@christopherdro 
Yeah, those changes should be applied here as well. I'll take a look at it. Thanks.

rborn:
@christopherdro sorry I didn't explain better, I was on the same page with the breaking change idea, Code Reviews are never a waste of time :)

salah-ghanim (approved):

salah-ghanim:
@arneson thanks for the contribution now it's on master 🎉


@rborn sorry for the holdup ;)

PR #3704, opened 2021-02-03
Clarify in the documentation that the edgePadding option for fitToSup…

arneson:
Clarify in the documentation that the edgePadding option for fitToSuppliedMarkers is only supported for Google Maps

No other PR does the same thing. 

fitToSuppliedMarkers disregards edgePadding when using Apple MapKit. This is now clarified in the documentation. 

### How did you test this PR?

I've tested on iOS that what I've stated is true. Is disregards edgePadding.


christopherdro:
Isn’t this part of the other PR you opened? I think keeping a single PR is fine here since the changes are related.

arneson:
Hi @christopherdro ! It is actually not. 
This only refers to another method which I did not write. Those changes where part of this 
https://github.com/react-native-maps/react-native-maps/pull/2479
but the changes for Apple MapKit was then reverted by: 
https://github.com/react-native-maps/react-native-maps/commit/1aee143302129f3811f1753155600bd041b5214d

But the documentation was never updated to show that the Apple MapKit support was removed.

christopherdro:
Thanks.

PR #3705, opened 2021-02-04
Update example project

ravirajn22:
1. Updated example project to 0.63.4
2. Updated project structure to build example project from inside example project it self
3. Removed GEM file setup, previously used to build and run example project
4. Other Clean ups not affecting the core library

Note: No changes affect the core library, it affects only the example project that to how we run it.

Check the section on how to run the example project, follow it and try running the project on iOS and android emulator. 

I have tested it on both emulators.

This change should clean up the library and make it easier to upgrade to latest versions as it releases.


rborn:
@ravirajn22 ❤️ 

There are a lot of linting issues, maybe you can have a look 🙏

ravirajn22:
Fixed it

rborn:
😻

christopherdro:
@ravirajn22 Do we need react-native-maps added as a dependency to the example project?

ravirajn22:
No its setup like using react-native-maps directly from the folder

Issue #3725, opened 2021-02-26
Camera `zoom` level rounds to an integer. 


ArshanKhanifar:
# Bug report
## Summary

When I'm using `MapView` with an `initialCamera` prop, the zoom level doesn't get set to the exact zoom level on the `Camera` object that I pass in. It rounds it down. `animateCamera()` works fine though. 

## Environment info
    react-native: 0.63.4
    react: 16.13.1
    react-native-maps: 0.27.1
    Target Platform: Android Google Pixel 4A

`react-native info` output:

```
System:
    OS: macOS 10.15.7
    CPU: (4) x64 Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
    Memory: 650.22 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 15.5.1 - /usr/local/Cellar/node/15.5.1/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.3.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.2, DriverKit 20.0, macOS 11.0, tvOS 14.2, watchOS 7.1
    Android SDK: Not Found
  IDEs:
    Android Studio: 4.1 AI-201.8743.12.41.6953283
    Xcode: 12.2/12B45b - /usr/bin/xcodebuild
  Languages:
    Java: 10.0.1 - /usr/bin/javac
    Python: 3.7.3 - /Users/arshankhanifar/miniconda3/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1
    react-native: 0.63.4 => 0.63.4
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

## Steps to reproduce
1.  create a blank page with just a `MapView` in it, initial camera is defined there:  
```
  e5_camera = {
    "zoom": 19.3294620513916,
    "heading": 330.87255859375,
    "pitch": 0,
    "center": {
      "longitude": -80.54009590297937,
      "latitude": 43.47296776345528
    }
  };

  state = {
    camera: this.e5_camera
  };

  render() {
        return <MapView
          ref={this.map_ref}
          style={StyleSheet.absoluteFillObject}
[...]

PR #3726, opened 2021-02-26
Android - Use double on `setCamera` to fix zoom precision. 

ArshanKhanifar:
No, no other PRs deal with Camera Zoom. 

It is fixing issue #3725 on Android devices. It is pretty simple. 

### How did you test this PR?
* I tested this PR in my own little project. The fix is so minimal that I didn't really see it necessary to create an online sample. If you'd like me to make a sample code, I'd happily do it.
* It's tested on a Google Pixel 4a phone. 
* I haven't tried iOS, this wouldn't affect iOS anyways.

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
@@ -508,7 +508,7 @@ public class AirMapView extends MapView implements GoogleMap.InfoWindowAdapter,
 
     builder.tilt((float)camera.getDouble("pitch"));
     builder.bearing((float)camera.getDouble("heading"));
-    builder.zoom(camera.getFloat("zoom"));
+    builder.zoom((float)camera.getDouble("zoom"));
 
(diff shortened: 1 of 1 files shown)


christopherdro (approved):

Shahriar87:
@christopherdro Hey it looks like this issue was fixed but was never released. We are facing this exact problem and I think this fix would solve it for us. Can this fix be release?

ArshanKhanifar:
@Shahriar87 in the meantime you can directly install from master, though Idk how stable it is:
```
yarn add https://github.com/react-native-maps/react-native-maps/
```

Shahriar87:
Ya, we are also concerned about how stable would that be.

christopherdro:
I will prepare a pre-release of v0.29.0 shortly that will include these changes. However, that is going to include everything that's currently on master. You can always fork the repo and cherry-pick this commit on top of your current version and create your own branch. Then you can update `react-native-maps` to pull from your branch in package.json instead of a published version number.

Issue #3727, opened 2021-03-01
Geojson Point should interactable


WrathChaos:
# Feature request

`Geojson` point should pressable and return its `overlay` value. That point is represented via `Marker`

## Why it is needed

With the GeoJSON, nothing is interactable however I need to interact with the points while using the Geojson.

## Possible implementation

Already implemented by #3728

### Code sample
```
<Geojson
      strokeWidth={4}
      geojson={MyGeoJson}
      strokeColor={'#81AAED'}
      onPress={(overlay: any) => console.log(overlay)} // This is not available in the current version
/>
```

PR #3499, opened 2020-07-14
Update Geojson.js

fattomhk:
The Geojson component not passing `tappable` and `onPress` props to Polygon

### Does any other open PR do the same thing?

No, can't find any.

### What issue is this PR fixing?

No, can't find any.

### How did you test this PR?

Tested on 
- Android(emulator) and iOS(simulator and iphone 11, ios 13.1 & ios 13.5) 
- with Google Map only

I cannot test on real Android device coz I don't have one.

Testing code: 
```
import React from 'react';
import { StyleSheet, View, Dimensions } from 'react-native';
import MapView, { Geojson, PROVIDER_GOOGLE } from 'react-native-maps';
import axios from 'axios';

const { width, height } = Dimensions.get('window');

const ASPECT_RATIO = width / height;
const LATITUDE = 22.3193;
const LONGITUDE = 114.1694;
const LATITUDE_DELTA = 1;
const LONGITUDE_DELTA = LATITUDE_DELTA * ASPECT_RATIO;
const GEOJSON = 'https://www.had.gov.hk/psi/hong-kong-administrative-boundaries/hksar_18_district_boundary.json';

class MapKml extends React.PureComponent {
	constructor(props) {
		super(props);

		this.state = {
			region: {
				latitude: LATITUDE,
				longitude: LONGITUDE,
				latitudeDelta: LATITUDE_DELTA,
				longitudeDelta: LONGITUDE_DELTA
			}
		};
		this.onPress = this.onPress.bind(this);
	}

	componentDidMount() {
		axios
			.get(GEOJSON)
			.then((res) => {
				const { data } = res;
				if (!!data) {
					if (data.type === 'FeatureCollection') {
						const geojsons = data.features.map((originalData) => {
							return {
								type: 'FeatureCollection',
								features: [ originalData ]
							};
						});
						this.setState({
							geojsons
						});
					} else {
						this.setState({
							geojson: data
						});
					}
				}
			})
			.catch((error) => {
				console.log(error);
			});
	}

	onPress({ data, nativeEvent }) {
		console.log(nativeEvent);
		alert(JSON.stringify(data, null, 2))
	}

	render() {
		const { geojson, geojsons } = this.state;
		console.log(geojsons);
		return (
			<View style={styles.container}>
				<MapView
					ref={(ref) => {
						this.map = ref;
					}}
					provider={this.props.provider}
					style={styles.map}
					initialRegion={this.state.region}
				>
					{!!geojson && (
						<Geojson
							tappable={true}
[...]

PR #3728, opened 2021-03-01
feat: Geojson marker has the onPress functionality now

WrathChaos:

I could not find any other PR to do the same thing. 

#3727 

### How did you test this PR?

- I tested this PR on both Android and iOS.
- Tested on both simulator and real devices

### DIFF
--- a/lib/components/Geojson.js
+++ b/lib/components/Geojson.js
@@ -3,51 +3,51 @@ import Marker from './MapMarker';
 import Polyline from './MapPolyline';
 import Polygon from './MapPolygon';
 
-export const makeOverlays = (features) => {
+export const makeOverlays = features => {
   const points = features
     .filter(
-      (f) =>
+      f =>
         f.geometry &&
         (f.geometry.type === 'Point' || f.geometry.type === 'MultiPoint')
     )
-    .map((feature) =>
-      makeCoordinates(feature).map((coordinates) =>
+    .map(feature =>
+      makeCoordinates(feature).map(coordinates =>
         makeOverlay(coordinates, feature)
       )
     )
     .reduce(flatten, [])
-    .map((overlay) => ({ ...overlay, type: 'point' }));
+    .map(overlay => ({ ...overlay, type: 'point' }));
 
   const lines = features
     .filter(
-      (f) =>
+      f =>
         f.geometry &&
         (f.geometry.type === 'LineString' ||
           f.geometry.type === 'MultiLineString')
     )
-    .map((feature) =>
-      makeCoordinates(feature).map((coordinates) =>
+    .map(feature =>
+      makeCoordinates(feature).map(coordinates =>
         makeOverlay(coordinates, feature)
       )
     )
     .reduce(flatten, [])
-    .map((overlay) => ({ ...overlay, type: 'polyline' }));
+    .map(overlay => ({ ...overlay, type: 'polyline' }));
 
   const multipolygons = features
-    .filter((f) => f.geometry && f.geometry.type === 'MultiPolygon')
-    .map((feature) =>
-      makeCoordinates(feature).map((coordinates) =>
+    .filter(f => f.geometry && f.geometry.type === 'MultiPolygon')
+    .map(feature =>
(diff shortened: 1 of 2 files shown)


lib/components/Geojson.js
```diff
@@ -159,6 +159,7 @@ const Geojson = props => {
               image={props.image}
               pinColor={markerColor}
               zIndex={props.zIndex}
+              onPress={() => props.onPress(overlay)}
```
christopherdro, line 162: Do we need to return overlay here or can follow what we do [here](https://github.com/react-native-maps/react-native-maps/blob/e38f5e1e61058711d571fe0029a7dd9c2cf070b2/lib/components/Geojson.js#L176) ?

WrathChaos: Actually, I think we need to return overlay otherwise I cannot find a way to reach which marker did I press. If you suggest a better way of doing that, I can convert it.

christopherdro:
Related to https://github.com/react-native-maps/react-native-maps/pull/3499

lib/components/Geojson.js
christopherdro: Are you using coordinates to determine what marker was pressed? The `onPress` on MapMarker should already be passing the coordinates from the native event back. https://github.com/react-native-maps/react-native-maps/blob/master/docs/marker.md#events  so you could just go with `onPress={props.onPress}`

WrathChaos: I tried that event but I could not get the overlay's data with that event. Simply it does not contain its data. That's why we need to return the pressed `overlay`. 

So, I did not use coordinates to determine what marker is pressed actually we should not do that, it is very redundant because we already have which marker and its data is pressed inside the GeoJSON's logic.

WrathChaos: CleanShot 2021-03-02 at 9 38 34

@christopherdro,
Here is the onPress event object, it simply does not contain anything.

christopherdro: @WrathChaos Thanks for the explanation. I have very limited experience using the Geojson component. @fattomhk Any thoughts here in relation to the changes to the Polygon component? Would it be beneficial to expose overlay there as well?

fattomhk: @christopherdro 
I agree that we need something to identify the clicked polygon like what we already have in maker's `onPress`

BTW,  I found that it is hard to customize the overlays with `Geojson`. In my project, I loop the geojson file and use the `Geojson` component for generating polygons.

WrathChaos: @fattomhk It is really hard to customize and use Geojson with interaction. We cannot even change the size of the marker with the geojson and these are really useful to customise it.

WrathChaos: Also, we need to add the press functionality for Polyline as well. I've made it on this project's other branch but waiting for this to resolve.

christopherdro: @WrathChaos Thoughts on including those changes in this PR?

WrathChaos: @christopherdro Sure.

WrathChaos: @christopherdro @fattomhk Both `Polyline` and `Polygon` has the pressable feature now. Can you guys check it again? 

Btw, we're using this code in production right now. It is working well.

christopherdro:
@WrathChaos This looks great! One last request. Could you update the docs to include the `onPress` prop? https://github.com/react-native-maps/react-native-maps/blob/master/docs/geojson.md

WrathChaos:
@christopherdro I updated the doc as well :)

WrathChaos:
@christopherdro  thank you! When do you think it is available on the new release?

christopherdro:
@WrathChaos It might take some time because we currently have 0.28.0 pending release so these changes will probably land in the release after that. I'd suggest cherry-picking this commit if you need to immediately in your project or patch the package with these changes.

MrHazimAli:
@christopherdro sorry to bug, may I know if we have exact eta for this to be deployed ? :-)

PR #3739, opened 2021-03-10
Use Default Geocoder to convert Coordinates by getting address

mechazod:

Need to leverage use of iOS CLGeoCoder instead using paid third-party for conversion

No active issue, instead adding facility to get address.

### How did you test this PR?

I did create a working branch to integrate with RN project and verify. I'm using only simulator for testing.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -269,6 +269,7 @@ declare module "react-native-maps" {
 
   export default class MapView extends React.Component<MapViewProps, any> {
     getCamera(): Promise<Camera>;
+    getAddress(): Promise<LatLng>;
     setCamera(camera: Partial<Camera>): void;
     animateCamera(camera: Partial<Camera>, opts?: { duration?: number }): void;
     animateToNavigation(
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,6 +8,19 @@ declare module "react-native-maps" {
     ViewProperties
   } from "react-native";
 
+  export interface Address {
+    name: string;
+    thoroughfare: string;
+    subThoroughfare: string,
+    locality: string,
+    subLocality: string,
+    administrativeArea: string,
+    subAdministrativeArea: string,
+    postalCode: string,
+    countryCode: string,
+    country: string,
+  }
+
   export interface Region {
     latitude: number;
     longitude: number;
@@ -269,7 +282,6 @@ declare module "react-native-maps" {
 
   export default class MapView extends React.Component<MapViewProps, any> {
     getCamera(): Promise<Camera>;
-    getAddress(): Promise<LatLng>;
     setCamera(camera: Partial<Camera>): void;
     animateCamera(camera: Partial<Camera>, opts?: { duration?: number }): void;
     animateToNavigation(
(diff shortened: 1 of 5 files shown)


lib/components/MapView.js
```diff
@@ -648,7 +648,29 @@ class MapView extends React.Component {
         this.props.onRegionChange(nativeEvent.region, details);
       }
     } else if (this.props.onRegionChangeComplete) {
-      this.props.onRegionChangeComplete(nativeEvent.region, details);
+      if (Platform.OS === 'ios') {
+        const { latitude, longitude } = nativeEvent.region;
+        return new Promise((resolve, reject) => {
+          this._runCommand('getAddressFromCoordinates', [
```
christopherdro, line 654: I don't think we want to call `getAddressFromCoordinates` every single time the region changes. Can we expose this as a method on Mapview?

mechazod: Alright will make as method props.

christopherdro:
@mechazod Thank you for the PR. Do you have any experience with Java? It'd be nice if we can wire up the same functionality for Android using Geocoder. https://developer.android.com/reference/android/location/Geocoder

lib/components/MapView.js
christopherdro: One other thing that might be worth considering is returning all the results instead of pulling the first one. Although I'm not sure how much data is in each result so it might be overkill if the first returned result is the most relevant.

index.d.ts
```diff
@@ -269,6 +269,7 @@ declare module "react-native-maps" {
 
   export default class MapView extends React.Component<MapViewProps, any> {
     getCamera(): Promise<Camera>;
+    getAddress(): Promise<LatLng>;
```
christopherdro, line 272: The typing for `getAdress is incorrect here. For `getAddress`, I suggest passing in a single `coordinate` object as a param instead of lat/long individually. This will match the other methods we have like `pointForCoordinate`. 

Also, we'll need a create a new type for the payload you are returning on this call. Here's an example of how that might look. I renamed `getAddress` to `addressForCoordinate` for clarity but am open to other suggestions as well.

```
type Address {
  name: string
  throughFare: string
  locality: string
  ...
}
```

`addressForCoordinate(coordinate: LatLng): Promise<Address>`

lib/ios/AirMaps/AIRMapManager.m
```diff
@@ -562,6 +562,32 @@ - (UIView *)view
     }];
 }
 
+RCT_EXPORT_METHOD(getAddressFromCoordinates:(nonnull NSNumber *)reactTag
+                  withLatitude:(double)latitude
+                  withLongitude:(double)longitude
+                  callback:(RCTResponseSenderBlock)callback)
```
christopherdro, line 568: To treat this as a promise we need to use `RCTPromiseResolveBlock ` and `RCTPromiseRejectBlock ` in place of`RCTResponseSenderBlock` https://reactnative.dev/docs/native-modules-ios#promises

christopherdro:
@mechazod Thank you for adding logic for Android. Just left a few more comments for consideration. Once we resolve those we'll just need to add this new method to the docs and we're good to go.

README.md
```diff
@@ -577,6 +577,21 @@ render() {
 }
 ```
 
+### Use of Default Geocoder to get address
+
```
christopherdro, line 581: Sorry, I should've been more clear here. I meant adding the method to the table in docs here: https://github.com/react-native-maps/react-native-maps/blob/master/docs/mapview.md#methods

mechazod: Got it @christopherdro will do.

christopherdro (approved):

Issue #3696, opened 2021-01-29
WMSTile renders tiles much bigger on android than iOS - why?


fewtrees:
# Question

On an iOS app using `<WMSTile>` renders images 

<img width="197" alt="iOS map example" src="https://user-images.githubusercontent.com/2836289/106304374-31aa8b00-6253-11eb-883e-5ac4ea2d5aeb.png">

On an Android app the rendered image is much bigger..

<img width="250" alt="Android map with tile graphic" src="https://user-images.githubusercontent.com/2836289/106303818-7d106980-6252-11eb-98a1-888ca2a745c6.png">

There is NO difference in the tile request, both set tile size to 256..

1. Why the size difference?
2. How could you change it?

Cheers

Rob

gavin-rose:
Hi! Can you send a code snippet or snack?

MaRaSu:
@fewtrees I'm not 100% sure however it looks like iOS is showing tile overlays "pixel to pixel" (ref: [https://developer.apple.com/documentation/mapkit/mktileoverlay/1452108-tilesize?language=objc](url)), while Android is scaling low-res (aka "normal" 256x256 pixel tiles) on high-res displays to the intended physical size on display. This is related to open PR #3740.

Compared to web Javascript map libraries like OpenLayers & Leaflet it seems that both iOS & Android tiled map implementations are very basic and lack many expected features like user defined handling of tile sizes & related scale factors. However I'm new to native mobile development so I might be missing something.

fewtrees:
@MaRaSu I have found that requesting a tile size of 768x768 renders the same on Android as 256x256 does on iOS.. This completely messes up a map server but it does look the same.

MaRaSu:
@fewtrees is your Android screen scale factor 3.0x (ref: [https://developer.android.com/training/multiscreen/screendensities.html](url))? That would explain what happens: iOS renders your tile pixel to pixel without caring about the screen pixel density and Android does what I think is the right default i.e. scales the tile.

I assume that when you say "request" you mean requesting a tile from the WMS tile server.

fewtrees:
@MaRaSu 
The android screen isn't scaled - everything untouched and tried on both an emulator plus physical device..

When you include tileSize property on the <WMSTile> obj set to 768, this gets included in the HTTP request back to the tile server..

i.e.

```
<WMSTile
  urlTemplate={resolvedUrl}
  zIndex={0}
  opacity={1}
  tileSize={768}
[...]

Issue #3388, opened 2020-04-14
Incredibly low resolution with OpenStreetMap, blurry tiles


Vahradrim:
# Bug

Using OpenStreetMap gives low resolution tiles. Tiles and labels are blurry.

## Environment info

> System:
>     OS: Windows 10 10.0.18362
>     CPU: (8) x64 Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
>     Memory: 3.50 GB / 7.95 GB
>   Binaries:
>     Node: 12.16.1 - H:\Programs\NodeJS\node.EXE
>     npm: 6.13.4 - H:\Programs\NodeJS\npm.CMD
>   npmPackages:
>     react: ~16.9.0 => 16.9.0
>     react-native: https://github.com/expo/react-native/archive/sdk-36.0.0.tar.gz => 0.61.4
>     react-native-maps: 0.26.1

## Steps To Reproduce

Use a MapView component with google maps, and change the following to use OpenStreetMap (according to the documentation)

Describe what you expected to happen:
Get the OpenStreetMap tiles in the same resolution than Google Maps

## Reproducible sample code

This code is working perfectly with Google Maps : 

> import MapView, {
>   MAP_TYPES,
>   PROVIDER_DEFAULT,
>   PROVIDER_GOOGLE,
> } from "react-native-maps";
> ...
> <MapView
>           ref={(map) => (currentMap = map)}
>           initialRegion={location}
>           showsUserLocation
>           style={{
>             height: Globals.Layout.window.height,
>             width: Globals.Layout.window.width,
>           }}
>         >
>         </MapView>
You get the correct result :
![Screenshot_20200414_123617_host exp exponent](https://user-images.githubusercontent.com/32706466/79253892-ea811400-7e51-11ea-817c-d193becbb8c6.jpg)

 Change to this to use OpenStreetMap, and you get correct tiles, but everything is blurred and in low resolution :

>  import MapView, {
>   MAP_TYPES,
>   PROVIDER_DEFAULT,
>   PROVIDER_GOOGLE,
> } from "react-native-maps";
> ...
> <MapView
>           ref={(map) => (currentMap = map)}
>           initialRegion={location}
>           showsUserLocation
>           provider={PROVIDER_DEFAULT}
>           mapType={MAP_TYPES.NONE}
>           style={{
>             height: Globals.Layout.window.height,
>             width: Globals.Layout.window.width,
>           }}
>         ><MapView.UrlTile
[...]

Issue #255, opened 2016-05-27
Offline maps support


bunnakal:
Dear all creators, 

Are there any possible ways to use offline maps data (both maps and satellite) with your third-party library?

Kind regards,

vessp:
I would also like to know if map tiles can be downloaded/cached and used while offline.

jayfunk:
Honestly, this is my only reason for not using this lib. I have been using [React Native MapBox](https://github.com/mapbox/react-native-mapbox-gl) because it has offline support for downloading tiles. I saw that you do have snapshot support. When performing a snapshot would react native annotations still be usable while using a snapshot?

allthetime:
I'm also curious about this. I'm wondering how trivial it would be to implement...

After searching around a bit, it looks like google at least doesn't support offline for apps, it is against the TOS, presumably because they are constantly updating and they don't want people displaying stale data with their name on it.

allthetime:
@jayfunk btw, how is react-native mapbox? I have a pretty simple implementation and would really like to have offline capabilities. I don't see it in the docs, do you know if there is a way to draw paths?

churcho:
This would really come in handy for me as well. Any feedback on road map?

devilop3rz:
+1

hernanmateo:
+1

alzalabany:
+1

kazak1377:
+1

ammichael:
+1

Unforgiven-wanda:
+1
This is absolutely essential for a project of mine. The minute is it implemented I will use this module again.

jlyman:
I also need offline support. Looking briefly at the implementations of AirMapUrlTile, would it be possible to pass in a local-referencing URL? For example, similar to [what's being done here](https://mixedreality.uk/2015/03/15/creating-tiled-overlays-with-mktileoverlay-and-swift/) on iOS:

```
//Get the URL template to the map tiles
let baseURL = NSBundle.mainBundle().bundleURL.absoluteString
let urlTemplate = baseURL?.stringByAppendingString("/Map1880/{z}/{x}/{y}.png/")
```

(Disclaimer, I haven't used the native implementations of either platforms, just done Leaflet.js locally using similar approaches.)

[...]

PR #3740, opened 2021-03-10
 Fix for blurry tiles in Android

hussainahmad:

No, but this is the new PR of #2694 . Previous PR is stalled and closed now. In parent PR two issues were fixed but I am not good in swift therefor I didn't added fixed code for second issue(Ability to overzoom on IOS)

URLTile appeared blurrier compared to IOS devices.  

Custom raster tiles appear to be blurrier compared to iOS devices on retina screens (DPI = 2). This patch also fixes this by dynamically creating higher resolution tiles using the tiles from a higher zoom level. Original solution is found here: https://stackoverflow.com/questions/23806348/blurred-custom-tiles-on-android-maps-v2/36696291

### How did you test this PR?
This PR is specific for Android tested in my own live projects in android devices

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapUrlTile.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapUrlTile.java
@@ -1,16 +1,113 @@
 package com.airbnb.android.react.maps;
 
 import android.content.Context;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.graphics.Canvas;
+import android.graphics.Color;
+import android.graphics.Paint;
 
 import com.google.android.gms.maps.GoogleMap;
+import com.google.android.gms.maps.model.Tile;
 import com.google.android.gms.maps.model.TileOverlay;
 import com.google.android.gms.maps.model.TileOverlayOptions;
+import com.google.android.gms.maps.model.TileProvider;
 import com.google.android.gms.maps.model.UrlTileProvider;
 
+import java.io.ByteArrayOutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 
 public class AirMapUrlTile extends AirMapFeature {
+
+  class CanvasTileProvider implements TileProvider {
+    static final int TILE_SIZE = 512;
+
+    public AIRMapUrlTileProvider mTileProvider;
+
+    public CanvasTileProvider(AIRMapUrlTileProvider tileProvider) {
+      mTileProvider = tileProvider;
+    }
+
+    @Override
+    public Tile getTile(int x, int y, int zoom) {
+      byte[] data;
+      Bitmap image = getNewBitmap();
+      Canvas canvas = new Canvas(image);
+      boolean isOk = onDraw(canvas, zoom, x, y);
+      data = bitmapToByteArray(image);
+      image.recycle();
+
+      if (isOk) {
+        Tile tile = new Tile(TILE_SIZE, TILE_SIZE, data);
+        return tile;
+      } else {
+        return mTileProvider.getTile(x, y, zoom);
+      }
+    }
+
+    Paint paint = new Paint();
+
+    private boolean onDraw(Canvas canvas, int zoom, int x, int y) {
+      x = x * 2;
+      y = y * 2;
+      Tile leftTop = mTileProvider.getTile(x, y, zoom + 1);
+      Tile leftBottom = mTileProvider.getTile(x, y + 1, zoom + 1);
+      Tile rightTop = mTileProvider.getTile(x + 1, y, zoom + 1);
+      Tile rightBottom = mTileProvider.getTile(x + 1, y + 1, zoom + 1);
+
+      if (leftTop == NO_TILE && leftBottom == NO_TILE && rightTop == NO_TILE && rightBottom == NO_TILE) {
+        return false;
+      }
+
+      Bitmap bitmap;
+
+      if (leftTop != NO_TILE) {
+        bitmap = BitmapFactory.decodeByteArray(leftTop.data, 0, leftTop.data.length);
+        canvas.drawBitmap(bitmap, 0, 0, paint);
+        bitmap.recycle();
+      }
+
+      if (leftBottom != NO_TILE) {
+        bitmap = BitmapFactory.decodeByteArray(leftBottom.data, 0, leftBottom.data.length);
+        canvas.drawBitmap(bitmap, 0, 256, paint);
+        bitmap.recycle();
+      }
+      if (rightTop != NO_TILE) {
+        bitmap = BitmapFactory.decodeByteArray(rightTop.data, 0, rightTop.data.length);
+        canvas.drawBitmap(bitmap, 256, 0, paint);
+        bitmap.recycle();
+      }
+      if (rightBottom != NO_TILE) {
+        bitmap = BitmapFactory.decodeByteArray(rightBottom.data, 0, rightBottom.data.length);
+        canvas.drawBitmap(bitmap, 256, 256, paint);
+        bitmap.recycle();
+      }
+      return true;
+    }
+
+
+    private Bitmap getNewBitmap() {
+      Bitmap image = Bitmap.createBitmap(TILE_SIZE, TILE_SIZE,
+              Bitmap.Config.ARGB_8888);
+      image.eraseColor(Color.TRANSPARENT);
+      return image;
+    }
+
+    private byte[] bitmapToByteArray(Bitmap bm) {
+      ByteArrayOutputStream bos = new ByteArrayOutputStream();
+      bm.compress(Bitmap.CompressFormat.PNG, 100, bos);
+
+      byte[] data = bos.toByteArray();
+      try {
+        bos.close();
+      } catch (Exception e) {
+        e.printStackTrace();
+      }
+      return data;
+    }
+  }
+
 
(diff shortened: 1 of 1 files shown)


christopherdro:
@hussainahmad Thank you for the PR. @the-habu @softwarefeeder Can you do a code review of this PR if you have a chance?

the-habu:
> @hussainahmad Thank you for the PR. @the-habu @softwarefeeder Can you do a code review of this PR if you have a chance?

Yes, I will do it over the next couple of days. Sounds like a good thing to have.

Thank you for mentioning me. I am still new to contributing to open-source projects like this and would not have dared to take the initiative.

MaRaSu:
@hussainahmad & @the-habu, this change will indeed improve the visual sharpness of map but it's good to understand that it is a tradeoff: when you pull 4 tiles from a higher zoom level you are effectively altering the map style. Map features e.g. text labels (e.g. for street names) become smaller than map style creator intended. 

Most of the time that hack works well enough and tradeoff is thus acceptable, however this is not the recommended way to fix blurriness on high-res displays. As also Google recommends on their SDK ([https://developers.google.com/maps/documentation/android-sdk/tileoverlay](url) the best way to is to supply high-res tiles 512x512 pixels from the tileserver (which enables the rendering back-end to render the map style correctly).

React-native-maps already supports tileSize property (<UrlTile> & <WMSTile>) for iOS and thus enabling use of high-res tiles providers. Therefore I would propose adding same support for Android as well on this PR, rather than continue to keep tileSize hardcoded to 256x256 pixels.

In my humble opinion the "4 tiles to 1" hack should be an option rather than default. However given that it's already default for iOS I guess it's acceptable. But that makes it even more important to support high-res tiles provider so that one can fix this properly whenever high-res tiles are available and only have a fallback to a hack whenever only low-res tiles are available.

fewtrees:
> @hussainahmad & @the-habu, this change will indeed improve the visual sharpness of map but it's good to understand that it is a tradeoff: when you pull 4 tiles from a higher zoom level you are effectively altering the map style. Map features e.g. text labels (e.g. for street names) become smaller than map style creator intended.
> 
> Most of the time that hack works well enough and tradeoff is thus acceptable, however this is not the recommended way to fix blurriness on high-res displays. As also Google recommends on their SDK ([https://developers.google.com/maps/documentation/android-sdk/tileoverlay](url) the best way to is to supply high-res tiles 512x512 pixels from the tileserver (which enables the rendering back-end to render the map style correctly).
> 
> React-native-maps already supports tileSize property ( & ) for iOS and thus enabling use of high-res tiles providers. Therefore I would propose adding same support for Android as well on this PR, rather than continue to keep tileSize hardcoded to 256x256 pixels.
> 
> In my humble opinion the "4 tiles to 1" hack should be an option rather than default. However given that it's already default for iOS I guess it's acceptable. But that makes it even more important to support high-res tiles provider so that one can fix this properly whenever high-res tiles are available and only have a fallback to a hack whenever only low-res tiles are available.

The recommendation from google is breaking their own rules. EPSG900913 recommends tile sizes of 256 and renders the map based on this tile size and zoom level. It's how the google standard works - and they set it.

tileSize isn't currently hardcoded to 256 and you can easily set it to another value. I set it to 768 and now the tile on Android and iOS looks the same.

MaRaSu:
> The recommendation from google is breaking their own rules. EPSG900913 recommends tile sizes of 256 and renders the map based on this tile size and zoom level. It's how the google standard works - and they set it.
> 
> tileSize isn't currently hardcoded to 256 and you can easily set it to another value. I set it to 768 and now the tile on Android and iOS looks the same.

@fewtrees I'm not sure what is your point with there reference to EPSG:9000913? As far as I understand it just defines the projected coordinate system, not quality of tile images, even if the tradition has been 256x256 pixels and later on with higher resolutions screen "retina" aka high-res tiles of 512x512 were invented.

When you say "tileSize isn't currently hardcoded to 256..." can you please be specific in what you mean. I assume you refer to issue #3696 which I understood was about requesting from a WMS tile server 768x768 tiles. If so, please note 2 things: 1) This PR is about UrlTile, not WMSTile - looking at the code they differ significantly with Android WMSTile native version actually using tileSize parameter unlike the URLTile implementation and 2) requesting 768x768 tile from the tile server is not the same as passing tileSize prop to URLTile (or WMSTile) component.

fewtrees:
> > The recommendation from google is breaking their own rules. EPSG900913 recommends tile sizes of 256 and renders the map based on this tile size and zoom level. It's how the google standard works - and they set it.
> > tileSize isn't currently hardcoded to 256 and you can easily set it to another value. I set it to 768 and now the tile on Android and iOS looks the same.
> 
> @fewtrees I'm not sure what is your point with there reference to EPSG:9000913? As far as I understand it just defines the projected coordinate system, not quality of tile images, even if the tradition has been 256x256 pixels and later on with higher resolutions screen "retina" aka high-res tiles of 512x512 were invented.
> 
> When you say "tileSize isn't currently hardcoded to 256..." can you please be specific in what you mean. I assume you refer to issue #3696 which I understood was about requesting from a WMS tile server 768x768 tiles. If so, please note 2 things: 1) This PR is about UrlTile, not WMSTile - looking at the code they differ significantly with Android WMSTile native version actually using tileSize parameter unlike the URLTile implementation and 2) requesting 768x768 tile from the tile server is not the same as passing tileSize prop to URLTile (or WMSTile) component.

@MaRaSu 
Hi, 
You're right about tile size - I think a lot of documentation online has made me take the wrong assumption because it talks in de facto terms about 256x256. It seems amazingly hard for tile servers to use anything else! 

Re: hardcoded 256 - yes, I was talking WMSTile so *my bad* and apologies :-)

MaRaSu:
@fewtrees yes, 256x256 is still incredibly popular given that it looks awful on modern high-resolution displays and mapnik & other popular rendering engines have supported high-res tiles already for a long time. Then again freely available OSM tile servers are not common and private tile servers use high-res much more often. Vector tiles are increasingly a solution as well if you happen to have all map data in vector format.

MaRaSu:
@hussainahmad & @the-habu I did some further testing with the existing implementation. At least Android v8, v10 and v11 seem to utilise retina / high-res tiles (512x512) EVEN if tileSize parameter for Android UrlTileProvider is incorrectly set to 256 (hardcoded in current RNM implementation as I commented above). I.e. Android detects the width / height of the tiles it loads and renders properly the higher resolution tile instead of doing the scaling it does for normal resolution tiles. This is not unusual, e.g. Leaflet JS map library does exactly the same.

This means that the proposed PR would actually make the current implementation worse: when high-res tile source is used the proposed new CanvasTileProvider would still try do its scaling hack. Therefore I strongly oppose implementing this PR as it is proposed now: support for the tileSize property needs to be added first.

@hussainahmad & @the-habu can you also comment if you intend to continue the work with this PR? I have great interest for URLTile capability because of my map application is relying on it and if this PR is stalled like the earlier similar PR #2694 then I might consider starting to work on a new PR addressing the issues outlined above, however obviously I do not want to do any parallel competiting work if this PR keeps on making progress.

hussainahmad:
@MaRaSu Thanks for the detail investigation, I have a shortage of time could you please Update this PR or create a new I will close this. unfortunately now I am not in position to update this PR for time short.  Best Regards

christopherdro:
@MaRaSu Thoughts on the best way to move forward with this PR?

MaRaSu:
@christopherdro my app is depending on RNM, specifically on Tile Overlays rather than Apple or Google Maps. Therefore I do have the interest to contribute to the project. I'm a bit struggling on the priorities (from my perspective): This PR is a bit of nice to have given high-res tiles already render as they should per my comment above and for normal resolution tiles there is another hack in issue #3388 which has side-effect (everything on map becomes scaled "down") but could be used.

Me biggest concern with RNM is actually tile cache logic: with poor cellular signal the Tile Overlays take time to load and at least on iOS the caching logic does not seem be nowhere as effective as for a web app. I have not yet tested LocalTile whether it could be at least a partial solution. Happy to take any pointers or information is someone else has tried to solving this earlier?

Anyways, I hope to circle back soon with some ideas where I could contribute. Need to start learning obj-c & Java somewhere...

MaRaSu:
@christopherdro ok. After the review I ended up with following priorities for improving RNM TileOverlays for real-world use cases:
1) Caching / offline capability - currently native side does minimal caching, with poor data connectivity map has poor UX or is completely useless
2) Scaling zoom above MaxZoom for TileOverlay - currently the tile server needs to serve tiles up to high zoom levels so that user can zoom in deep enough. E.g. Leaflet.js & Openlayers support auto-scaling to provide tiles above "maxNativeZoom" of tile server, greatly optimising caching performance
3) This PR, Android normal tile sources scaling to modern high-res displays.

I'm currently experimenting with 1) to see how far it can be taken with JS side implementation (with LocalTile / URLTile with file:// tiles), looked very initially promising however the problem is Android and especially iOS not updating native tile Overlay as tiles in filesystem become available. Only fix is to re-render LocalTile / URLTile which causes visible glitch on display. If JS side implementation does not provide sufficient solution then native side solution (similar to what was discussed in #255) would be way to go.

I have not done any obj-c nor Android Java so thought about starting a new discussion on these improvements as a new issue or re-opening #255 to continue discussion specifically about cache / offline capability. @christopherdro what would you recommend?

christopherdro:
@MaRaSu Thank you for taking the time. I think it's best to reopen #255 and continue the discussion there.

PR #3744, opened 2021-03-12
Updates `AnimatedRegion` component

jasonmichaels:
This PR doesn't directly address any issues; rather, it's more of a cleanup PR. 

### Changes
- cleans up the `spring` and `timing` methods of the `AnimatedRegion` component
- implements a more readable flow in the constructor when setting `latitude`, `longitude`, `latitudeDelta`, and `longitudeDelta` via the `setInitialValue` method

### Testing
Run the `example` apps to confirm functionality.


lib/components/AnimatedRegion.js
```diff
@@ -119,80 +129,33 @@ export default class AnimatedMapRegion extends AnimatedWithChildren {
...
-        })
-      );
-
+    const animations = [];
+    for (const type of configTypes) {
+      if (config.hasOwnProperty(type)) {
+        animations.push(
+          Animated.spring(this[type], {
+            ...config,
+            toValue: config[type],
```
christopherdro, line 138: There's been some reports about a lot of warnings for `useNativeDriver` and am wondering if we should add it here as well.

jasonmichaels: I can throw that in. First, though, I'll search open issues to document `useNativeDriver` reports and ensure there aren't other places we may need to explicitly pass it where, currently, no configuration for the property exists.

jasonmichaels: @christopherdro Per the types and interfaces declared in `index.d.ts`, I think the `Animated.timing` and Animated.spring` configurations with `useNativeDriver: true` should work, and thus eliminate some in-dev warnings developers see that clog up the Flipper <em>et al</em> dialogs. There may be other opportunities, but I think we can safely wait and see if anyone reports additional excessive warnings. See [this commit](https://github.com/react-native-maps/react-native-maps/pull/3744/commits/64a26e05932fc4c5b141f2f77155dfd39088873e).

christopherdro: I'm a bit nervous about setting the native driver to true here. What if we expose `useNativeDriver` as a prop on config and default it to false? This way we retain the same behavior, reduce the yellow box warnings and expose an option to opt into using the native driver to test perf improvements.

jasonmichaels: OK -- working on that over the next few days, including updating docs.

jasonmichaels: @christopherdro I opted to pass it directly into the `AnimatedRegion` constructor. Also, I merged `master` in, and it looks like there are some `android` files that weren't part of the original `master` checkout I was working with. Thoughts?

christopherdro: Good call. I like the latter where it is part of the param passed into `timing` or `spring`.

christopherdro:
@jasonmichaels I'm not sure why those files int the android directory were creating but I don't think it was from merging in master. Either way, do you mind checkout out those files? I think the only files that need to be changed are `AnimatedRegion.js ` and `index.d.ts`

lib/components/AnimatedRegion.js
christopherdro: @jasonmichaels Any updates here?

lib/components/AnimatedRegion.js
```diff
@@ -8,40 +8,53 @@ if (__DEV__) {
...
+      defaultValues.latitudeDelta
+    );
+    this.longitudeDelta = this.setInitialValue(
+      valueIn.longitudeDelta,
+      defaultValues.longitudeDelta
+    );
     this._regionListeners = {};
   }
 
+  setInitialValue(value, fallback) {
```
miallo, line 52: I am probably a bit nit-picky, but this **function itself** does not really _set_ the initial values, right? I agree, that finding a good name is hard, but maybe something along the lines of `getAnimatedValue` describes a bit better, what it is actually doing? And maybe it can be extracted from the class, since it isn't really coupled to it in any meaningful way and is rather independent?

christopherdro:
@jasonmichaels any updates here ?

christopherdro:
@jasonmichaels can you please give an update to this PR?

lib/components/AnimatedRegion.js
jasonmichaels: @christopherdro Yeah, sorry about the delay here. Started a new gig back in late March, and been somewhat slammed. Kept meaning to jump back onto this. I just pushed [this commit](https://github.com/react-native-maps/react-native-maps/pull/3744/commits/7627ca353fc8a0356cbf3c71a470832478694406), which coerces it from `config.useNativeDriver`. I'll update the docs next.

jasonmichaels:
@christopherdro I updated the PR to infer/coerce `useNativeDriver` from per-`spring`/`timing` calls in the config. Also updated the docs and added notes to example files `AnimatedMarkers` and `AnimatedRegions` indicating that while `useNativeDriver` is set to `true` in said examples, it defaults to `false` if not specified/given.

jasonmichaels:
@miallo I removed the instance method, `setInitialValue` to a standalone function, `getAnimatedValue`. Not sure if there's a better place for that (I'm considering a PR to consolidate constants and common functions).

lib/components/AnimatedRegion.js
christopherdro: No worries. I appreciate you taking the time to follow up. Changes look good. I'll give a test over the weekend.

christopherdro (approved):

PR #3745, opened 2021-03-12
Geojson GeometryCollection Support

tez123z:

(please answer here)
Not to my knowledge

(please link the issue here)
There's not an open issue

### How did you test this PR?
Verified renders with three different data sets - one with GeometryCollection geometry - one without and one with both.

(please answer here)
Google Maps geojson component on both Android and iOS
Tested on both simulators and real devices

### DIFF
--- a/lib/components/Geojson.js
+++ b/lib/components/Geojson.js
@@ -95,6 +95,21 @@ const makeCoordinates = feature => {
   }
 };
 
+const unpackGeometryCollections = features  => {
+  return features.map(f => 
+    f.geometry && 
+    f.geometry.type === "GeometryCollection" ? f.geometry.geometries.map(g => {
+        
+      return {
+        type: 'Feature',
+        properties: f.properties || {},
+        geometry: g
+      }
+
+    }).reduce(flatten, []) : [f]
+  ).reduce(flatten, [])
+}
+
 const doesOverlayContainProperty = (overlay, property) => {
   // Geojson may have 0 for the opacity when intention is to not specify the
   // opacity. Therefore, we evaluate the truthiness of the propery where 0
@@ -143,7 +158,7 @@ const getStrokeWidth = (props, overlay) => {
 };
 
 const Geojson = props => {
-  const overlays = makeOverlays(props.geojson.features);
+  const overlays = makeOverlays(unpackGeometryCollections(props.geojson.features));
   return (
     <React.Fragment>
       {overlays.map((overlay, index) => {
--- a/lib/components/Geojson.js
+++ b/lib/components/Geojson.js
@@ -98,7 +98,8 @@ const makeCoordinates = feature => {
 const unpackGeometryCollections = features  => {
   return features.map(f => 
     f.geometry && 
-    f.geometry.type === "GeometryCollection" ? f.geometry.geometries.map(g => {
+    f.geometry.type === "GeometryCollection" && 
+    f.geometry.geometries ? f.geometry.geometries.map(g => {
         
       return {
         type: 'Feature',
--- a/lib/components/Geojson.js
+++ b/lib/components/Geojson.js
@@ -95,21 +95,25 @@ const makeCoordinates = feature => {
   }
 };
 
-const unpackGeometryCollections = features  => {
-  return features.map(f => 
-    f.geometry && 
-    f.geometry.type === "GeometryCollection" && 
-    f.geometry.geometries ? f.geometry.geometries.map(g => {
-        
-      return {
-        type: 'Feature',
-        properties: f.properties || {},
-        geometry: g
-      }
-
-    }).reduce(flatten, []) : [f]
-  ).reduce(flatten, [])
-}
+const unpackGeometryCollections = features => {
+  return features
+    .map(f =>
+      f.geometry &&
+      f.geometry.type === 'GeometryCollection' &&
+      f.geometry.geometries
+        ? f.geometry.geometries
+            .map(g => {
+              return {
+                type: 'Feature',
+                properties: f.properties || {},
+                geometry: g,
+              };
+            })
+            .reduce(flatten, [])
+        : [f]
+    )
+    .reduce(flatten, []);
+};
 
 const doesOverlayContainProperty = (overlay, property) => {
   // Geojson may have 0 for the opacity when intention is to not specify the
@@ -150,7 +154,7 @@ const getColor = (props, overlay, colorType, overrideColorProp) => {
 
 const getStrokeWidth = (props, overlay) => {
   if (props.hasOwnProperty('strokeWidth')) {
-    return props['strokeWidth'];
+    return props.strokeWidth;
   }
   if (doesOverlayContainProperty(overlay, 'stroke-width')) {
     return overlay.feature.properties['stroke-width'];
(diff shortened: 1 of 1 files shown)


tez123z:
Added support for GeometryCollection feature type on Geojson component via unpackGeometryCollections function.
Function filters for GeometryCollection type then maps each geometry as separate feature.

christopherdro:
@fattomhk @WrathChaos Does this apply to your implementation with Geojson? Can you help out with a code review?

WrathChaos:
@tez123z I really like the idea but I just want to ask you that do we really need to do it inside the component? Don't you think that the dev should handle it on their side for the GeoJSON data?

tez123z:
@WrathChaos Your call. I ran into this when doing some conversions from kml/kmz files to Geojson. At a glance on geojson.org it appears that GeometryCollection shouldn’t be supported but it is mentioned here - https://tools.ietf.org/html/rfc7946#section-3.1.8 - I think it could prove useful to have coverage on this geometry type.

WrathChaos:
@tez123z Don't get me wrong but I think we should do this logic outside of the library. Before using GeoJSON, we should use proper data to use it. Of course, this is just my idea. I certainly use this utility function before using the GeoJSON but I think everyone should handle themselves. 

@christopherdro What do you think?

tez123z:
Right so wouldn’t GeometryCollection being included in The GeoJSON Specification qualify as “proper data”? Link here - https://tools.ietf.org/html/rfc7946#section-3.1.8 - I feel like I’m missing something, can you help me understand?

> @tez123z Don't get me wrong but I think we should do this logic outside of the library. Before using GeoJSON, we should use proper data to use it. Of course, this is just my idea. I certainly use this utility function before using the GeoJSON but I think everyone should handle themselves.
> 
> @christopherdro What do you think?

WrathChaos:
> @WrathChaos Right so wouldn’t GeometryCollection being included in The GeoJSON Specification qualify as “proper data”? Link here - https://tools.ietf.org/html/rfc7946#section-3.1.8 - I feel like I’m missing something, can you help me understand?
> 
> > @tez123z Don't get me wrong but I think we should do this logic outside of the library. Before using GeoJSON, we should use proper data to use it. Of course, this is just my idea. I certainly use this utility function before using the GeoJSON but I think everyone should handle themselves.
> > @christopherdro What do you think?

@tez123z `GeometryCollection` can be proper data for the `GeoJSON` but I believe that it should not handle itself, the devs should handle it on their own side.  I mean they should handle the `unpackGeometryCollections` function on their own side. I don't see why `GeoJSON` should handle it. Please explain it to me if I misunderstood or missed something.

tez123z:
> > @WrathChaos Right so wouldn’t GeometryCollection being included in The GeoJSON Specification qualify as “proper data”? Link here - https://tools.ietf.org/html/rfc7946#section-3.1.8 - I feel like I’m missing something, can you help me understand?
> > > @tez123z Don't get me wrong but I think we should do this logic outside of the library. Before using GeoJSON, we should use proper data to use it. Of course, this is just my idea. I certainly use this utility function before using the GeoJSON but I think everyone should handle themselves.
> > > @christopherdro What do you think?
> 
> @tez123z `GeometryCollection` can be proper data for the `GeoJSON` but I believe that it should not handle itself, the devs should handle it on their own side. I mean they should handle the `unpackGeometryCollections` function on their own side. I don't see why `GeoJSON` should handle it. Please explain it to me if I misunderstood or missed something.

Ah ha! Ok - thanks for the clarification. Yes that makes sense. Although the geojson prop value is described as "Geojson description of object." so with GeometryCollection being a valid description - one might think the Geojson function would handle it. Maybe it could be an option? I have see the geometry quite often in my experience with Geojson.

WrathChaos:
> > > @WrathChaos Right so wouldn’t GeometryCollection being included in The GeoJSON Specification qualify as “proper data”? Link here - https://tools.ietf.org/html/rfc7946#section-3.1.8 - I feel like I’m missing something, can you help me understand?
> > > > @tez123z Don't get me wrong but I think we should do this logic outside of the library. Before using GeoJSON, we should use proper data to use it. Of course, this is just my idea. I certainly use this utility function before using the GeoJSON but I think everyone should handle themselves.
> > > > @christopherdro What do you think?
> > 
> > 
> > @tez123z `GeometryCollection` can be proper data for the `GeoJSON` but I believe that it should not handle itself, the devs should handle it on their own side. I mean they should handle the `unpackGeometryCollections` function on their own side. I don't see why `GeoJSON` should handle it. Please explain it to me if I misunderstood or missed something.
> 
> Ah ha! Ok - thanks for the clarification. Yes that makes sense. Although the `geojson` prop value is described as "Geojson description of object." so with `GeometryCollection` being a valid description - one might think the `Geojson` function would handle it. Maybe it could be an option? I have seen the `GeometryCollection` geometry quite often in my experience with Geojson.

Actually, I thought the same, maybe we can export this function to make the devs use it? What do you think about that?

tez123z:
@WrathChaos Works for me - so you'd prefer it be an export? Or should it be a property?

christopherdro:
@tez123z @WrathChaos I'm leaning towards having this logic live outside this library with the intention of trying to keep things as lean as possible. However, if you feel strongly about including it I think providing it as an exported function can be useful.

tez123z:
@christopherdro @WrathChaos Did y'all have a chance to check out the link I provided - https://tools.ietf.org/html/rfc7946#section-3.1.8 ? Is the intention with this function to support all valid Geojson? Or just `Point`, `LineString`, `Polygon`, `MultiPoint`, `MultiLineString`, and `MultiPolygon`? https://geojson.org/ reads as though these are the only valid geometries but the The GeoJSON Specification (RFC 7946) mentions `GeometryCollection` (link above) https://tools.ietf.org/html/rfc7946

WrathChaos:
> @christopherdro @WrathChaos Did y'all have a chance to check out the link I provided - https://tools.ietf.org/html/rfc7946#section-3.1.8 ? Is the intention with this function to support all valid Geojson? Or just `Point`, `LineString`, `Polygon`, `MultiPoint`, `MultiLineString`, and `MultiPolygon`? https://geojson.org/ reads as though these are the only valid geometries but the The GeoJSON Specification (RFC 7946) mentions `GeometryCollection` (link above) https://tools.ietf.org/html/rfc7946

I got this but still, this logic can be handled on the outside of the library.  The library itself does not need to handle it inside of it.. 
I agree with @christopherdro, libraries should keep lean. If you think it is really necessary we can make it an exportable function.

christopherdro:
@tez123z I did take a look at the RFC and can see that a `GeometryCollection` is part of the spec but the argument being made here is the API in this library is designed to simply display geometry objects on a map vs. implementing GeoJson spec. Where do we draw the line of what is supported or not? Should we be taking `Antimeridian Cutting` into account as well then?

Not parsing GeometryCollection in this lib doesn't make it invalid, it's just something that this library doesn't support. Also seeing that there are no issues created in the past to address this makes me lean more towards excluding this logic from this library.

@tez123z I appreciate you taking the time to open this PR, and @WrathChaos for chiming in and contributing to the discussion but I am going to go ahead and close this PR and leave it up to the developer to unpack the collection.

tez123z:
Well bummer - almost got my first open source PR!

WrathChaos:
Keep going @tez123z :) It's a good start!

fattomhk:
wow, you guys got a great discussion on Geojson.
I think following RFC spec is always a good practise. The problem we are facing is if we want more control to the overlays, developers need to do the `unpackGeometryCollections` themselves.

maybe an optional`renderOverlay(type, data, ...anymore)` function?

PR #3762, opened 2021-03-24
iOS add support for disabling userLocation callout

salah-ghanim:
NO

option to disable user location callout on iOS

### How did you test this PR?

manually on iOS device / simulator


christopherdro (approved):

gianpaj:
Is this userLocation callout?

I'm trying to see how to disable this "user avatar" current location pin/marker. It appears after you press on the blue circle of your location.

I'm on a previous version

mzanini:
Interesting. Like @gianpaj said, this does not seem to disable the user location callout. Even passing `userLocationCallout={false}` as an argument to the `<MapView>` element, the callout still pops up when clicking on the blue dot that represents the user's location. Was able to reproduce on a simulator and on real device, running the application through Expo with the Expo Go app.

christopherdro:
@mzanini these changes are included in current prerelease v0.29 while expo is on the latest stable release of 0.28. I’ll be releasing 0.29 as a stable version soon.

Issue #3769, opened 2021-04-01
Custom (dynamic) markers bugs & performance issues


eyalyoli:
Bug report

## Summary
There is no stable way to make dynamic custom markers! There are weird bugs in the current implementation, the docs are lacking and the only way that really works has very bad performance issues which make it unusable!

## Environment info
Using Google maps for android and Apple maps for iOS.
```
"react-native-maps": "0.27.1",
"react": "16.13.1",
"react-dom": "16.13.1",
"react-native": "0.63.3",
```

`react-native info` output:

```bash
System:
    OS: macOS 10.15.6
    CPU: (6) x64 AMD Ryzen 5 3500X 6-Core Processor             
    Memory: 40.27 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.7.0 - /usr/local/bin/node
    Yarn: Not Found
    npm: 7.7.5 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.4, DriverKit 20.2, macOS 11.1, tvOS 14.3, watchOS 7.2
    Android SDK:
      API Levels: 28, 29
      Build Tools: 28.0.3, 29.0.2, 29.0.3, 30.0.0
      System Images: android-28 | Google Play Intel x86 Atom_64
      Android NDK: 21.3.6528147
  IDEs:
    Android Studio: Not Found
    Xcode: 12.4/12D4e - /usr/bin/xcodebuild
  Languages:
    Java: 14.0.1 - /usr/bin/javac
    Python: 2.7.16 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.3 => 0.63.3 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

## Steps to reproduce

There are 3 possible ways to implement custom markers (I've tested them all):
1. using `image` prop of `Marker` - 
```javascript
    <Marker
      coordinate={coordinate}
      image={require("./images/map_pin.png")}
    />
```
2. using `icon` prop of `Marker` - 
```javascript
    <Marker
      coordinate={coordinate}
      icon={require("./images/map_pin.png")}
    />
```
[...]

PR #3771, opened 2021-04-01
Better docs for custom markers

eyalyoli:

Nope

#3769 

### How did you test this PR?

No code just docs... But all of instructions were tested on a private project

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,21 +117,26 @@ import { Marker } from 'react-native-maps';
 </MapView>
 ```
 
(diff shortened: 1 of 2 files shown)


rossjcooper:
@eyalyoli thanks for updating the docs for this, in Android Studio, which settings should we use to generate the images for the map icons?

Screenshot 2022-04-01 at 10 10 19

eyalyoli:
I don't know, but i think you'll get the same results when using a launcher or actionbar icons

PR #3789, opened 2021-04-14
URLTile tile caching

MaRaSu:

No.

This is a first step in addressing the issue #255 offline maps support: provide tile caching capability for URLTile component on native side for default map providers on both iOS and Android. When caching is enabled with the new tileCachePath property tiles are loaded with urlTemplate as usual and then saved in cache (directory) and further tile loading is done from the cache. 

All cache management logic i.e. storage space limits, cache purge, refreshing old / expired cache tiles needs to be on JS side in client implementation i.e. no cache management is provided by this PR.

Yet I believe even this minimal implementation will solve the foundational problem in the tile caching capability which cannot be effectively solved on JS side (see last few comments in issue #255). 

### How did you test this PR?

NB! This is a draft PR, not yet ready for merging. My request is for a code review by people who have experience with iOS / Android native dev and preferably RNM code base. My own experience with native side development is limited to learn-as-you-go approach with this PR.

Code is currently tested with my own RN map app and works in normal conditions. As a next step I will implement a new view for the example project to facilitate testing. After additional testing is completed I will also remove excess logging from the code.

iOS implementation is less risky since without enabling the cache it uses the "old" tile loading logic. Android implementation is higher risk since even without enabling the cache it uses new logic for loading tiles, a custom TileProvider implementation rather than system provided UrlTileProvider.


MaRaSu:
Example is now done and I have tested with it the implementation: 

- On iOS & Android, both emulator and real devices

- Functionality both with and without the tile caching enabled

- Stress testing with no & poor network connectivity & recovery of network connectivity: performance seems to be on par with previous implementation and system provided exponential back-off recovery seems to be working ok and both old and new implementation are equally "stupid" in recovering from a loss of connectivity i.e. they try to re-connect all pending tile fetches instead of prioritising what is visible currently in device screen (like e.g. LeafletJS does on web).

Logging is still in PR and probably should be removed before merging as RNM does not seem to do much normal logging? Still it's good to have for anyone who wants to test the code before final merging is done.

example/examples/CacheURLTiles.js
```diff
@@ -0,0 +1,114 @@
...
+          mapType={this.mapType}
+          style={styles.map}
+          initialRegion={region}
+        >
+          <UrlTile
+            urlTemplate="http://c.tile.stamen.com/watercolor/{z}/{x}/{y}.jpg"
+            zIndex={-1}
+						// Empty tileCachePath = defaults to Documents-directory on iOS, 
+						// on Android empty string does not activate tileCachePath at all
+						tileCachePath={''}
```
felipecsl, line 55: this doesn't look correct. by reading the method `getTileFilename()`, I can see it will return a non-null filename if `tileCachePath` is set to empty string. Also the different behavior between iOS and Android to me seems surprising to me. The behavior should be consistent across platforms I assume

MaRaSu: @felipecsl I agree ideal would be to have consistency across platforms. Even if in RNM it does not seem be the standard, even for features were underlying platform support is similar e.g. tileSize has not been supported for Android even if it exists in platform the same way as in iOS (this PR fixes that).

However my intent for supporting empty string for tileCachePath was to have native code use a default directory (app documents): this is now implemented for iOS, however for Android empty strings in properties do not seem to get to the URLTile at all, therefore caching is not enabled and the problem that you refer to in `getTileFilename()` does not actually happen. I'm new to RN, RNM and native side coding on both platform - so I was hoping one of the reviewers could guide in how to solve this empty string thing on Android so that I can then add some code to set the default directory correctly.

MaRaSu:
`tileCacheMaxAge`property implemented for automatic refreshing of cached tiles. With this I think the native side now has the minimum yet sufficient for most use cases implementation enabling rest of offline & caching logic to be implemented on the JS side by client code.

MaRaSu:
@rborn & @christopherdro what are your thoughts on moving this PR forward? PR is in use in my app which is in production and working well, though naturally that is testing only one particular way of using the tile cache mechanism.

christopherdro:
@MaRaSu Thank you for the follow up. I’ll have some time next week to test and do another review. In the meantime can you remove all debug logs you added? Log.d and NSLog

MaRaSu:
@christopherdro thanks for circling back on this. Now all logging removed except for one error case.

MaRaSu:
@christopherdro I'm now working on the next version of this, with following new functionality:
- Android high resolution tiles (based on PR #3740 however controlled with a property and proper handling for tileSize-property). NB iOS Mapkit already does this automatically whether you want it or not...
- LeafletJS-style maximumNativeZ which allows zooming in even if tile source does not support tiles at said zoom levels, also great for offline-mode & caching to manage growth of cache size
- Offline-mode property which uses only cache and furthermore scales lower zoom tiles if tiles at desired zoom levels are not available in cache, therefore allowing for much better offline user experience
- Refactoring implementation and integration with WMSTile (current version only on URLTile)

I have already implemented all new functionality for Android, except finalised WMSTile integration. Now starting with iOS.

Since this PR has not been moving forward one option is to wait until I finish the iOS version and after that I deploy it to production in my app which uses all these features and thus does some real world testing. I can then do a new PR if there is still interest to merge this to RNM. The downside of that is that there is going to be a lot more code to review.

MaRaSu:
@christopherdro I have not implemented new features listed above and have done first round of testing & fine-tuning with my app. Changes were substantial enough for both iOS & Android, and therefore I do not think it makes sense to completely replace existing implementations as it would be high risk.

On iOS it was possible to add logic which only uses new TileProvider code when new properties (for my new features) are used, otherwise it uses old TileProvider logic. For Android it probably is possible as well but looking at the RNM code I'm not quite sure how the logic works for AirMapUrlTile instance: whenever it is added as an Overlay to Map it cannot be changed anymore - and removing & re-adding looks complicated (in iOS-version there was earlier code for that which I re-used). Based on my quick testing on Android it looks like AirMapUrlTile properties are set before adding as an Overlay to Map - however I'm not sure if that is guaranteed.

Can you advise on the above - the alternative logic is to create a new UrlCachedTile & WMSCachedTile sub-components for MapView for new code. Quite a bit of repetitive code needed for this...Let me know what works best, ideally I would like to settle on this soon otherwise code what I need for my own app and what could be merged back to RNM start to deviate too much.

MaRaSu:
@christopherdro, I had finally enough time to finish also example code and testing. This code has been in use now for several weeks in my RN app with MAU +3k which has provided additional testing. 

During the implementation I run into many finer details of iOS and Android TileOverlay & TileProvider implementations e.g. Android only using a limited number of threads for GetTile (of TileProvider) calls - which then required the change of network fetch logic from synchronous to WorkManager driven (to allow also for repeated retries to GetTile method). This is to say that the code is a result of rather substantial experimentation & testing, rather than "first attempt" at developing the functionality.

This PR now implements new properties to both URLTile + WMSTile, for both Android and iOS (iOS only AirMaps!):
- maximumNativeZ (*): Zooming past maximumNativeZ will scale tiles in device, this enables more efficient caching and zooming in further what is available from the tile server
- doubleTileSize (only Android): This is the same as in PR #3740 i.e. for tileSize 256(x256) tiles it fetches 4 higher zoom tiles to create a 512x512 tile which looks better on high-resolution displays, however impacts negatively some aspects of map styling esp. text label size - NB! iOS does this automatically, without any settings to disable the scaling.
- tileCachePath (*): Enables Caching and Offline tiles from the directory. Automatically saves fetched tiles to directory and it tiles have been prefetched then uses those tiles
- tileCacheMaxAge: Enables automatic refreshing of tiles with "stale-while-refresh" logic i.e. when a stale (expired) tile is served a background refresh fetch is triggered. Serving stale tiles allows for offline / poor network coverage use.
- offlineMode: Turns on offline-mode where tiles are only served from cache, with no attempt to fetch (or refresh cache) from network to save battery and network bandwidth. If tiles at requested zoom levels are not found then tiles in lower zoom levels are tried and if found scaled appropriately - search goes 4 zoom levels deep. This improves user experience whenever tile cache is not fully populated with tiles at all needed zoom levels.

Properties marked above with (*) will activate new functionality. Without those properties old URLTile & WMSTile functionality is used, which should greatly reduce the risk that is coming from all the new functionality this PR has implemented for caching & scaling.

I have left the logging in the code. If you want to test with example-code logging is very useful to see what code does. I can then remove the logging before the merge (assuming we get there).

MaRaSu:
BTW, for getting the example-folder project to compile on latest XCode (12.5.1) had to follow tip in [https://stackoverflow.com/questions/66189325/xcode-throws-atomic-notify-oneunsigned-long-is-unavailable/67314652#67314652](https://stackoverflow.com/questions/66189325/xcode-throws-atomic-notify-oneunsigned-long-is-unavailable/67314652#67314652)

shealan:
This is a very exciting development. I hope this gets included in the official release soon.

MaRaSu:
@christopherdro & @rborn can you please let me know if there is more I can do to facilitate merging of this PR. Otherwise it would be useful for me to know if there simply is no capacity or interest to merge it, in that case I can just work on my own fork of RNM without putting in the extra effort of generalizing the improvements I'm developing for the needs of my own app.

christopherdro:
Hi @MaRaSu. I plan on including this in the next pre-release but would like to have more code review done on the native side. I'm no expert but will try and take a closer look soon. Pinging @felipecsl @alvelig @salah-ghanim to see if they have any bandwidth to review.

example/examples/CacheURLTiles.js
```diff
@@ -0,0 +1,128 @@
...
+            // 5) With tileCacheMaxAge too
+            // 6) With offlineMode=true too - test zoom in to test scaling of lower zoom level tiles to higher zoom levels
+            //
+            maximumNativeZ={15}
+            // For testing activate different tile cache paths, examples below
+            // work for simulator / emulator testing
+            // This is for iOS simulator
+						//tileCachePath="/Users/suomimar/Library/Developer/CoreSimulator/tiles"
+            // This is for Android simulator
+            tileCachePath="/data/user/0/com.airbnb.android.react.maps.example/files/tiles"
```
christopherdro, line 67: @MaRaSu Are we going to rely developer to figure out the location of where the file should be saved? It might make sense to configure the file path on the native side.

MaRaSu: @christopherdro yes, the current implementation expects developer to use react-native-fs or expo-file-system or something similar to get a correct path.

In earlier version (that @felipecsl commented above) for iOS there was automatic configuration for iOS app Documents directory if` tileCachePath = ""`, however in Android empty string for property never got through to the native code. 

I'm no expert in RN mechanisms for passing properties and difference between iOS & Android so I did not dive deeper into resolving that. A workaround could be naturally to add yet another property to indicate native side to assign the file path but that does not feel elegant.

christopherdro: I think leaving it up to the developer to specify the file path makes sense but am wondering if we should throw an error if a file path is invalid or not specified. Is there any use of the feature if the tiles can't be saved?

MaRaSu: @christopherdro few thoughts on this:
- if `tileCachePath` nor `maxNativeZ` is not specified then old UrlTile and WMSTile implementations are used and they work as normal (without caching obviously). For this case in iOS there is just few lines of new code in the execution path before "falling back" to old implementation, in Android there is slightly more but the principle is the same - the purpose here is 2 fold: ensure full compatibility with existing clients of RNM as well as de-risk this new implementation: new code only gets activated with new properties.
- RNM does not seem to throw errors easily e.g. old implementation of URLTile nor WMSTile do not throw errors if absolutely mandatory property URLTemplate is missing - therefore I did not consider throwing errors for any errors in properties either e.g. for malformed / invalid `tileCachePath,` however some defensive code has been built in.

christopherdro: @MaRaSu Thank you for the explanation. I was able to successfully test on iOS and will try and test Android over the weekend. Would you be willing to update the docs for our tile components following the same format as the [others](https://github.com/react-native-maps/react-native-maps/blob/master/docs/).

MaRaSu: @christopherdro good to hear about the progress with testing. Let me know if any questions arise from testing or code review.

I pushed a commit for the new Docs page for `URLTile`and `WMSTile`. Might benefit from some additional text explaining or referencing some link for map tiles basics, as reading through RNM issues log gives me a feeling that many developers just want to enable (free) OpenStreetMap while do not really understand the basics of map tiling and tileservers.

christopherdro: Yea I noticed that as well. What you put together is great. Thanks again!

kalebheitzman:
This works perfectly in an app I'm building. Thanks for the incredible functionality!

christopherdro (approved):

shealan:
Thank you to everyone involved and especially @MaRaSu for all your hard work. I will be rolling this functionality into some pretty high profile apps in the next few weeks. I appreciate all the hard work you put into this feature.

christopherdro:
@kalebheitzman @shealan I'm going to include these changes in the next pre-release which I will put together in the next week or so. Please report back here with any issues you may come across so we can try and address them before shipping these changes in a stable release.

@MaRaSu Thank you again for your contributions and patience throughout the process.

shealan:
@christopherdro good stuff. I'll report back if I encounter anything notable.

kalebheitzman:
I've only encountered one odd thing but it was was easily fixed. When specifying the cache folder name, you have to add a trailing slash or the folder doesn't get created correctly. 

```
<WMSTile
    urlTemplate="https://kygisserver.ky.gov/arcgis/rest/services/WGS84WM_Services/KY_Data_Tiles_PointCloud_WGS84WM/MapServer/export?dpi=96&transparent=true&format=png32&layers=show%3A0&bbox={minX}%2C{minY}%2C{maxX}%2C{maxY}&bboxSR=102100&imageSR=102100&size={width}%2C{height}&f=image"
    tileSize={256}
    opacity={0.5}
    tileCachePath={`${FileSystem.documentDirectory}tiles/5dem_grid/`}
/>
```

If I use ```${FileSystem.documentDirectory}tiles/5dem_grid/``` then the tiles get placed in the 5dem_grid folder. If I leave off the trailing slash and use ```${FileSystem.documentDirectory}tiles/5dem_grid``` then the tiles get placed in the tiles folder and 5dem_grid does not get created. This is a problem if you are caching multiple tilesets and forget the trailing slash because tiles then begin to overwrite one another.

MaRaSu:
@kalebheitzman Hmmm, that is interesting issue. Just re-checked my code and directory path handling looked safe for both iOS & Android. I also tested with example app on both iOS & Android simulators `tileCachePath` both with & without trailing slash - both worked fine. Can you elaborate a bit how whether problem was in iOS or Android, simular or real device etc. that would help to narrow down the issue.

kalebheitzman:
@MaRaSu it was iOS simulator specifically. I haven't tested it on an iOS device. The problem was weird.

For example, if you had the following folder structure ```/tiles/set1/set2``` without the trailing slash, only ```/tiles/set1``` would get created. With the trailing slash, all of the folders get created. This is all from an Expo managed app.

MaRaSu:
@kalebheitzman thx for info, will check again and see if I can reproduce. BTW how did you integrate with Expo? I ejected my Expo app earlier this year when I started tinkering with this development fir RNM.

kalebheitzman:
@MaRaSu I haven't ejected the project. I'm just following the latest documentation to init a new app and install the react-native-maps dependency. I was referencing your branch when I ran into the issue but now am on the master branch of this repo.

MaRaSu:
@kalebheitzman thanks for providing detailed information on the issue, with that I managed to identify the bug: it was in the handling of file path references of type `file:///xxx/yyy`, which Expo creates rather than using normal directory path references. RFC 1808 defines how those are handled with and without the trailing slash, resulting in the slightly odd behaviour that you observed.

@christopherdro I will create a new PR with a fix to the issue @kalebheitzman found. Will submit soon after I do some testing first.

kalebheitzman:
Hey @MaRaSu thanks for all the incredible work on this. This may be a bug or it might be an implementation issue on my end. I'm not sure. I went to test my app in the field today and caching didn't seem to work like I expected (and that may be because I'm thinking about caching differently than being offline). Here is some sample code that does work like I expect when I flip an offline switch in my app. Does your code automatically adjust to read locally when offline or should I be implementing that on my end?

```
export default ({ active, offline }: MapProps) => {

	if (!active) return null

	if (offline) {
		console.log(`Going offline`)
		return(
			<UrlTile
				urlTemplate={`${FileSystem.documentDirectory}tiles/foxtown/{z}/{x}/{y}.png`}
				tileSize={256}	
				opacity={0.5}
				maximumZ={18}
				zIndex={50}
			/>
		)

	}

	console.log(`going online`)
	return(
		<UrlTile
			urlTemplate="https://geoserver.foxtownboulders.com/geoserver/gwc/service/tms/1.0.0/foxtown:foxtown-mbtiles@OSMTILE@png/{z}/{x}/{y}.png?flipY=true"
			tileSize={256}	
			tileCachePath={`${FileSystem.documentDirectory}tiles/foxtown/`}
			opacity={0.5}
			maximumZ={18}
			zIndex={50}
			offlineMode={offline}
		/>
	)
}
```

kalebheitzman:
Hey @MaRaSu sad news... It doesn't look like tile caching is actually working from an expo built app and deployed to iOS. It works great in the simulator. I just created a debug build to see if files were being written to storage and they are not. I'm using expo-file-system so I'm going to back to their docs to see if I've missed something, but it seems odd it works while running metro server on both the Simulator and Device, but not when natively built. 

Maybe I need to eject my code and go from there? I'll report back what I find.

christopherdro:
@kalebheitzman you will need to eject to test because there were native changes made for this update.

kalebheitzman:
@christopherdro that did it. Works fine now. Ejected and built via XCode.

MaRaSu:
@kalebheitzman yep, that was the reason I asked from you how did you get this new RNM version working with an app using managed Expo workflow 😀   From what I had read from Expo blog & docs they have now in preview new capability which is enabling building of custom clients ("Expo Go"), allowing developer to use custom native code instead of what comes pre-built into Expo Go (e.g. in their latest SDK they are using RNM version v.0.28.0 as @christopherdro pointed out). BTW, Expo SDK43 is probably coming out soon, great if new RNM version could be made available to it with URLTile & WMSTile caching functionality.

shealan:
Any update on this being included in an official release?

danloiterton:
Hi @MaRaSu, thanks for all this work - there's some lovely features here. I've pulled v30.0.0 to try out the tile caching stuff, which seems to be working pretty well. 

I noticed the new "maximumNativeZ" prop on the UrlTile component also, but I can't seem to get that to work. I have a tile service that only goes to zoom 20, so I set maximumNativeZ={20}. But when I zoom to around 21, all the tiles disappear. From the docs, it sounds like the tiles from zoom 20 should be reused and scaled up, correct?

I'm testing with Google Maps on iOS - is it designed to work in that environment? I've tried it with and without the "tileCachePath" prop.

MaRaSu:
Hi @danloiterton, on iOS new functionality (caching, maximumNativeZ etc.) is only implemented on Apple Maps (aka "Airmaps"). Porting to AirGoogleMaps would be a good job for someone...

PR #3801, opened 2021-04-28
Fix import/local conflict in typings

renchap:
When using Typescript in strict mode, it fails with this error:

```
error TS2440: Import declaration conflicts with local declaration of 'Animated'.
```

It is caused by the typings importing `Animated` from `react-native` while defining a local `Animated` class.
Using an import alias fixes the issue.

No

None, I directly opened the PR instead of an issue

### How did you test this PR?

We are using this updated typings in our app and TS compiler is no longer complaining.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,7 +1,7 @@
 declare module 'react-native-maps' {
   import * as React from 'react';
   import {
-    Animated,
+    Animated as RNAnimated,
     ImageRequireSource,
     ImageURISource,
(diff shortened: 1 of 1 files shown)


stephen776:
Ran into this same issue. Would be great to get this merged ASAP

renchap:
In the meantime you can use the `--skipLibCheck` typescript option so errors in libraries files are not reported for your project.

PR #3827, opened 2021-05-27
Include missing information on MapEvent definition

CrisTofani:

No

Typescript definition file is missing informations such: 
- `action`
- `id`

Which are instead sent by AirMap bridge module when pressing inside the map. 

Using typescript would cause error at compile time even if the information is provided at run time.

### How did you test this PR?

Android/iOS with both Google and Apple Maps

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -46,12 +46,24 @@ declare module 'react-native-maps' {
     y: number;
   }
 
+  export type EventActionType =
+    | 'marker-press'
+    | 'polygon-press'
+    | 'polyline-press'
+    | 'callout-press'
+    | 'press'
+    | 'long-press'
+    | 'overlay-press'
+    | undefined;
+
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -46,12 +46,24 @@ declare module 'react-native-maps' {
...
 
+  export type EventActionType =
+    | 'marker-press'
+    | 'polygon-press'
+    | 'polyline-press'
+    | 'callout-press'
+    | 'press'
+    | 'long-press'
+    | 'overlay-press'
+    | undefined;
```
christopherdro, line 57: Is there ever a case where the action will be undefined ?

CrisTofani: Yep @christopherdro , actually the value results `undefined` when pressing outside of any specific element (marker, polygon or any other)

Issue #3830, opened 2021-05-30
Improve installation documentation


gahfy:
# Feature request

## Why it is needed

The installation document is written in a way that you cannot avoid sharing your API Key in the repository if you want to share your code

## Possible implementation

PR to be made

### Code sample

N/A

### PR

#3831

PR #3831, opened 2021-05-30
Avoid sharing API key in repository

gahfy:
…git repository

No PR is doing the same

#3830

### How did you test this PR?

I just followed the instructions I wrote in my own code

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -38,17 +38,26 @@ The app's `Info.plist` file must contain a `NSLocationWhenInUseUsageDescription`
 
 ### Enabling Google Maps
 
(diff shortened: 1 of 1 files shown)


gahfy:
My english is not perfect. So it may be reworded if needed

christopherdro:
I think it's best to leave it up to the developer on how keys should be pulled and would rather keep the installation steps to a minimum. I do like the idea though of discouraging users so maybe we can keep this section here with the notice.
https://github.com/react-native-maps/react-native-maps/pull/3831/files#diff-8b042e3f94ca5c59a7cd990b950aec0073ea84fe811e7b22be51158d7b180d56R76-R81

Issue #3854, opened 2021-06-16
[0.29.0] - Android crashes when addressForCoordinate returns empty list


owinter86:
# Bug report

## Summary

The new addressForCoordinate will crash on android with `java.lang.IndexOutOfBoundsException: Index: 0, Size: 0`
when moving the map to coordinates that tigger an empty list result from the geocoder.

## Environment info

`react-native info` output:

```bash
# paste your react, react-native, react-native-maps versions here
```

## Steps to reproduce

1. Using the new RN pre-release (0.29.0) and the new `addressForCoordinate` method, zoom all the way out and move around near the bearing sea on android

Describe what you expected to happen:

1. Invalid coordiantes or empty results should not hard crash the app but return throw and get caught in the JS code.

Describe what you actually happens:

1. The app will crash with java.lang.IndexOutOfBoundsException: Index: 0, Size: 0

christopherdro:
Related https://github.com/react-native-maps/react-native-maps/issues/3850

Issue #3850, opened 2021-06-16
New addressForCoordinate always throws on iOS


owinter86:
# Bug report

## Summary

I have been testing the recent Pre-release and notice that the new addressForCoordinate will always throw on iOS, I beliebe the conditional here is incorrect

https://github.com/mechazod/react-native-maps/blob/5efd3534e43ba21bffd2e64d946c7a9580b390b7/lib/ios/AirMaps/AIRMapManager.m#L575-L578

as it will throw if `coordinate != nil` which is any time coordinates are correctly added.

Below is a patch package diff solving the issue with what I believe is the correct conditional

```diff
diff --git a/node_modules/react-native-maps/lib/ios/AirMaps/AIRMapManager.m b/node_modules/react-native-maps/lib/ios/AirMaps/AIRMapManager.m
index 726885f..c5ce4ff 100644
--- a/node_modules/react-native-maps/lib/ios/AirMaps/AIRMapManager.m
+++ b/node_modules/react-native-maps/lib/ios/AirMaps/AIRMapManager.m
@@ -575,9 +575,9 @@ - (UIView *)view
         if (![view isKindOfClass:[AIRMap class]]) {
             reject(@"Invalid argument", [NSString stringWithFormat:@"Invalid view returned from registry, expecting AIRMap, got: %@", view], NULL);
         } else {
-            if (coordinate != nil ||
-                ![[coordinate allKeys] containsObject:@"latitude"] ||
-                ![[coordinate allKeys] containsObject:@"longitude"]) {
+            if (coordinate != nil &&
+                (![[coordinate allKeys] containsObject:@"latitude"] ||
+                ![[coordinate allKeys] containsObject:@"longitude"])) {
                 reject(@"Invalid argument", [NSString stringWithFormat:@"Invalid coordinate format"], NULL);
             }
             CLLocation *location = [[CLLocation alloc] initWithLatitude:[coordinate[@"latitude"] doubleValue]
```

christopherdro:
Ping @mechazod

mechazod:
@owinter86 I intentionally set that condition to check each statement, because some might pass invalid data for `coordinate`. I don't see any issue here. Can you point out issue will encounter of the existing condition?

owinter86:
@mechazod if coordinate is not nil it will throw, that's the issue. It will never not throw if passing in coordinate data.

mechazod:
[...]

PR #3855, opened 2021-06-16
Fix crashes related to getAddressFromCoordinates for android and iOS

owinter86:

No

#3850 and #3854

### How did you test this PR?

After testing the pre-release, I have found these two main issues listed above, which are solved with the above commit

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
@@ -214,6 +214,7 @@ public class AirMapModule extends ReactContextBaseJavaModule {
                   geocoder.getFromLocation(coordinate.getDouble("latitude"),coordinate.getDouble("longitude"),1);
           if (list.isEmpty()) {
             promise.reject("Can not get address location");
+            return;
           }
           Address address = list.get(0);
 
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
@@ -212,6 +212,9 @@ public class AirMapModule extends ReactContextBaseJavaModule {
         try {
           List<Address> list =
                   geocoder.getFromLocation(coordinate.getDouble("latitude"),coordinate.getDouble("longitude"),1);
+          if (list.isEmpty()) {
+            promise.reject("Can not get address location");
+          }
           Address address = list.get(0);
 
           WritableMap addressJson = new WritableNativeMap();
--- a/lib/ios/AirMaps/AIRMapManager.m
+++ b/lib/ios/AirMaps/AIRMapManager.m
@@ -575,7 +575,7 @@ RCT_EXPORT_METHOD(getAddressFromCoordinates:(nonnull NSNumber *)reactTag
         if (![view isKindOfClass:[AIRMap class]]) {
             reject(@"Invalid argument", [NSString stringWithFormat:@"Invalid view returned from registry, expecting AIRMap, got: %@", view], NULL);
         } else {
-            if (coordinate != nil ||
+            if (coordinate == nil ||
                 ![[coordinate allKeys] containsObject:@"latitude"] ||
                 ![[coordinate allKeys] containsObject:@"longitude"]) {
                 reject(@"Invalid argument", [NSString stringWithFormat:@"Invalid coordinate format"], NULL);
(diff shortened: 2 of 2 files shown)


lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
```diff
@@ -212,6 +212,9 @@ public void execute(NativeViewHierarchyManager nvhm)
         try {
           List<Address> list =
                   geocoder.getFromLocation(coordinate.getDouble("latitude"),coordinate.getDouble("longitude"),1);
+          if (list.isEmpty()) {
+            promise.reject("Can not get address location");
```
christopherdro, line 216: Should we keep the error message the same across both platforms?

christopherdro:
ping @mechazod

lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java
christopherdro: Also, it might be wise to check for `longitude & `latitude` properties like we do on iOS alongside the check for an empty list.

owinter86: I believe that check is already happening earlier on, and won't get to this point unless valid coordinate data has been provided.

mechazod: @owinter86 I agree, can you add handler if empty response of `geocoder` https://github.com/owinter86/react-native-maps/blob/fix/geocoder/lib/ios/AirMaps/AIRMapManager.m#L602 with similar message. BTW, this is good to have. 👍

christopherdro: A yea I see the logic here: https://github.com/react-native-maps/react-native-maps/blob/a0e3b393c3d5b4fb3bac03c7a4e4b427889624b3/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java#L205-L208

@mechazod It looks like the conditional you had set up on Android is what @owinter86 is proposing on iOS. Maybe it was just an oversight?

mechazod: Nice @christopherdro good catch.

christopherdro:
@owinter86 Could you add an `else` to the conditional [here](https://github.com/react-native-maps/react-native-maps/blob/a0e3b393c3d5b4fb3bac03c7a4e4b427889624b3/lib/ios/AirMaps/AIRMapManager.m#L588-L602) and reject the promise with the same error message as you do on Android?

lib/ios/AirMaps/AIRMapManager.m
```diff
@@ -599,6 +599,8 @@ - (UIView *)view
                             @"countryCode" : [NSString stringWithFormat:@"%@", placemark.ISOcountryCode],
                             @"country" : [NSString stringWithFormat:@"%@", placemark.country],
                         });
+                    } else {
+                        reject(@"Invalid argument", [NSString stringWithFormat:@"Can not get address location"], NULL);
```
christopherdro, line 603: Can we update the message here to match android? “ Can not get address location”

owinter86: Both android and iOS have the same `error.message` and that is `Can not get address location`, however it seems that through this library only iOS errors have an `error.code` in this case is `Invalid argument` which may not suit.

owinter86: For reference both of these errors will look like this on the JS side.

android
```
{
  "code": "EUNSPECIFIED",
  "message": "Can not get address location",
  "userInfo": null,
  "nativeStackAndroid": []
}
```
ios
```
{
  "code": "Invalid argument",
  "message": "Can not get address location",
  "domain": "RCTErrorDomain",
  "userInfo": null,
  "nativeStackIOS": [
    "0   Wayke                               0x000000010115e9f0 RCTJSErrorFromCodeMessageAndNSError + 120",
    "1   Wayke                               0x00000001010f5a08 __41-[RCTModuleMethod processMethodSignature]_block_invoke_2.129 + 152",
    "2   Wayke                               0x00000001015d9aa0 __72-[AIRMapManager getAddressFromCoordinates:coordinate:resolver:rejecter:]_block_invoke_2 + 188",
    "3   libdispatch.dylib                   0x0000000180100580 _dispatch_call_block_and_release + 24",
    "4   libdispatch.dylib                   0x0000000180101d44 _dispatch_client_callout + 16",
    "5   libdispatch.dylib                   0x000000018010f2f4 _dispatch_main_queue_callback_4CF + 956",
    "6   CoreFoundation                      0x0000000180365d1c __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 12",
    "7   CoreFoundation                      0x000000018036025c __CFRunLoopRun + 2448",
    "8   CoreFoundation                      0x000000018035f3bc CFRunLoopRunSpecific + 572",
    "9   GraphicsServices                    0x000000018afdd70c GSEventRunModal + 160",
    "10  UIKitCore                           0x00000001843f03d0 -[UIApplication _run] + 964",
    "11  UIKitCore                           0x00000001843f51ac UIApplicationMain + 112",
    "12  Wayke                               0x00000001007f0a10 main + 108",
    "13  libdyld.dylib                       0x0000000180224554 start + 4"
  ]
}
```

PR #3920, opened 2021-08-16
GeoJSON onPress Crash Fix and Marker Customization Options

WrathChaos:

No other PR has the same thing

No issue right there, directly a fix and a new feature.

### How did you test this PR?

Tested on both iOS and Android with both simulators and real devices.

# What to do
## Bug Fix
This PR fixes the `GeoJSON`'s onPress undefined crash fix.

CleanShot 2021-08-16 at 6 52 10@2x

---------

## New Feature 

- GeoJSON has the default Marker for the `Point` however there are no `customization` options for the `GeoJSON`. Now it has the new prop option: `markerChildren` which receives any UI component for the new Marker Image.

### **Example:**

Big yellow Point markers are the default one and we could not customize its style in any way. Small-green Point markers are the new `Image` or `FastImage` any other UI component which can be customized by the dev itself.

CleanShot 2021-08-16 at 6 58 42@2x

### DIFF
--- a/lib/components/Geojson.js
+++ b/lib/components/Geojson.js
@@ -153,7 +153,7 @@ const Geojson = props => {
         const strokeWidth = getStrokeWidth(props, overlay);
         if (overlay.type === 'point') {
           return (
-           <Marker
+          <Marker
               key={index}
               coordinate={overlay.coordinates}
               image={props.image}
--- a/lib/components/Geojson.js
+++ b/lib/components/Geojson.js
@@ -153,7 +153,7 @@ const Geojson = props => {
         const strokeWidth = getStrokeWidth(props, overlay);
         if (overlay.type === 'point') {
           return (
-          <Marker
+            <Marker
               key={index}
               coordinate={overlay.coordinates}
               image={props.image}
--- a/lib/components/Geojson.js
+++ b/lib/components/Geojson.js
@@ -1,7 +1,171 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import Marker from './MapMarker';
 import Polyline from './MapPolyline';
 import Polygon from './MapPolygon';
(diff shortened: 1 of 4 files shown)


christopherdro:
Thank you for the PR. Can you update the docs and TS types to reflect the changes?

lib/components/Geojson.js
```diff
@@ -159,8 +159,10 @@ const Geojson = props => {
               image={props.image}
               pinColor={markerColor}
               zIndex={props.zIndex}
-              onPress={() => props.onPress(overlay)}
-            />
+              onPress={() => props.onPress && props.onPress(overlay)}
+            >
+              {props.markerChildren}
```
christopherdro, line 164: Thoughts on using `component` for the prop name?

WrathChaos: Only `component` not referring to the `Marker` maybe `markerComponent` ?

christopherdro: makerComponent sounds good

christopherdro:
@WrathChaos I'd like to include these changes in v0.29 before releasing it as a stable version. Can you give me an idea of when you can finish up the last set of changes to get this merged? Thanks!

christopherdro (approved):

WrathChaos:
> @WrathChaos I'd like to include these changes in v0.29 before releasing it as a stable version. Can you give me an idea of when you can finish up the last set of changes to get this merged? Thanks!
> 
> Update: Docs, types and example have been updated in [3c37082](https://github.com/react-native-maps/react-native-maps/commit/3c37082087b161568449500263e491cb17408633)

Hello @christopherdro,
Really sorry for very late response. I was in vacation, could not take care of this PR. Thank you so much for the contribution :)

Issue #3927, opened 2021-08-22
Unable to install in new react-native project via npm due to incorrect peer dependencies


kirkryan:
# Bug report

## Summary

When using react-native 0.64.2 and trying to install react-native-maps via npm you will receive a peerDependency error due to incorrect version formats specified in the package.json for react-native-maps.

## Environment info

`react-native info` output:

```bash
"react": "17.0.1",
"react-native": "0.64.2",

```

## Steps to reproduce

1. create new react-native project via npx react-native init <projectname>
2. install as per instructions npm install react-native-maps --save-exact. I noticed that npm is actually behind the current GitHub version so also used npm install react-native-maps@0.29.2 to manually attempt installation of the latest version.

Describe what you expected to happen:

1. expect npm to complete successfully with the usual "added 829 packages, and audited 830 packages in 1m"

Describe what you actually happens:

1. npm error
2. 
![image](https://user-images.githubusercontent.com/18603029/130356474-f545137c-5120-4023-99b7-dda688b382d4.png)

## Reproducible sample code

Run the steps above.

PR #3928, opened 2021-08-22
fix peerDependencies to allow react-native-maps installation via npm …

kirkryan:
…using react native latest

No

https://github.com/react-native-maps/react-native-maps/issues/3927

### How did you test this PR?

I forked the repo, updated the semver logic in the package.json and was able to install react-native-maps into a project that had existing dependencies and avoiding npm dependency hell.

You may also confirm this fix by installing from my forked repo: 
`npm install https://github.com/kirkryan/react-native-maps`

This has been testing in xCode 12.5.1 with the iOS simulator 14.4 iPhone 12

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "mapkit"
   ],
   "peerDependencies": {
-    "react": "^16.0.0 || ^17.0.0 || ^18.0.0",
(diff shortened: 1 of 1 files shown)


kirkryan:
My bad..

Issue #3936, opened 2021-09-03
WMSTile and URLTile Opacity


kalebheitzman:
# Bug report

## Summary

There seems to be no way to add transparency to URLTiles or WMSTiles. Setting opacity seems to do nothing on URLTile or WMS from the tile servers I'm pulling from. I've tried apply opacity with the the style prop as well as styled components to no avail.

## Environment info

react-native: 0.63.4
react: 16.13.1
react-native-maps: git+ssh://git@github.com/react-native-maps/react-native-maps.git#master
Target Platform: iOS 14.5 Simulator/iPhone 11 Apple Maps

`react-native info` output:

```bash
System:
    OS: macOS 11.5.2
    CPU: (4) x64 Intel(R) Core(TM) i5-7267U CPU @ 3.10GHz
    Memory: 27.10 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.0 - ~/.nvm/versions/node/v14.16.0/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.17.0 - ~/.nvm/versions/node/v14.16.0/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.10.2 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    Android SDK: Not Found
  IDEs:
    Android Studio: 4.1 AI-201.8743.12.41.7199119
    Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
  Languages:
    Java: Not Found
    Python: 2.7.16 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1
    react-native: ~0.63.4 => 0.63.4
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

## Steps to reproduce

1. git clone git@github.com:jamfire/boulders-react-native.git
2. cd boulders-react-native
3. yarn install
4. yarn start
5. open new terminal tab
6. yarn run ios

Describe what you expected to happen:

1. Expect URLTile and WMSTile to have css applied with opacity factor based on opacity settings

Describe what you actually happens:

1. Opacity doesn't take effect

christopherdro:
@MaRaSu any ideas ?

[...]

PR #3938, opened 2021-09-04
iOS impl for URLTile & WMSTile opacity property

MaRaSu:

No.

#3936

### How did you test this PR?

Updated example code and tested with iOS simulator.


christopherdro (approved):

kalebheitzman:
Amazing! This works perfectly. Tested on iOS simulator and on device.

kalebheitzman:
Ran into a bug. Opacity works fine when using Metro on an iOS device. When I ran my build through ```expo build:ios``` opacity doesn't seem to be working. Not sure why it would work when running on the metro server but not the actual build...

kalebheitzman:
Disregard my last comment. After ejecting from Expo transparency works fine. Maybe that's something to add to the docs. I've only been working with Expo for a couple of weeks and still trying to figure it all out.

PR #3944, opened 2021-09-08
[iOS] Add heading to onUserLocationChange

### LINKED ISSUE #3924: Map crash on eas build product in android
Hi,

We have been using this package for a while now both in development and production with expo build and it has worked as expected.

Recently we migrated to eas, for the sole purpose of cutting down the size of the app. 

After eas build and submit are implemented without any errors.

The map functions as should in iOS, but on android, it lags for a few seconds and crashes.

Sentry does not log any errors for this crash.

We have use the expo run:android command and there’s no error message being displayed and the crash still happens.

Has any one experienced this in the past?

### LINKED ISSUE #3934: onUserLocationChange callback heading undefined on iOS AppleMaps
# Bug report

## Summary

`onUserLocationChange` callback in iOS MapKit is not returning the heading (its always undefined).
I went through the code and I believe the issue is [in this line of code here](https://github.com/react-native-maps/react-native-maps/blob/master/lib/ios/AirMaps/AIRMapManager.m#L1021)

I am not sure how to change it tho I'm not very experiences with iOS development. (maybe we can add development manual to this repo on how to setup and contribute to it and i'd be happy to assist?)

## Environment info

`react-native info` output:

```bash
      react-native: 0.65.1
      react: 17.0.2
      react-native-maps: 0.28.0 
     target: iOS (AppleMaps)
```

## Steps to reproduce

1. Setup a plain MapView
2. Ensure you have `showUserLocation ={true}`
3. pass a function to `onUserLocationChange` and inspect the callback to get the heading

Describe what you expected to happen:

The heading in the `onUserLocationChange` callback should NOT return `undefined` 

## Reproducible sample code

simple EXPO example here https://snack.expo.dev/@eyadal/react-native-maps-heading-undefined (run on iPhone of course)

christopherdro:
No

Closes #3934

### How did you test this PR?

Added `onUserLocationChange` callback for iOS to the example app and confirmed heading output matches Google Maps

### DIFF
--- a/example/examples/EventListener.js
+++ b/example/examples/EventListener.js
@@ -78,14 +78,6 @@ class EventListener extends React.Component {
   }
 
   render() {
-    // Events that are dependent on
-    let googleProviderProps = {};
-    if (this.props.provider === PROVIDER_GOOGLE) {
-      googleProviderProps = {
-        onUserLocationChange: this.recordEvent('Map::onUserLocationChange'),
-      };
-    }
-
     return (
       <View style={styles.container}>
         <MapView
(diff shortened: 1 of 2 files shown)


kalebheitzman:
So this is returning the heading/course now in the location callback. On iOS maps there's an option to display a course indicator with the blue dot that shows location on screen. Does react-native-maps support this or is that something else that i need to code myself?

christopherdro:
@kalebheitzman it should. Please take a look at the docs for mapview, otherwise open a separate ticket. `showsUserLocation` Prop might be what you’re looking for

PR #3964, opened 2021-09-22
Feature: GeoJSON Marker Title Prop & A bit Refactor

WrathChaos:

No other PR has the same thing

No issue right there, directly a fix and a new feature.

### How did you test this PR?

Tested on both iOS and Android with both simulators and real devices.

# What to do

## New Feature 

- `GeoJSON` has the new prop option: `title` which shows a default `Callout` with the title.
- A bit of code refactoring for props. It has a de-construction for better code maintainability.


WrathChaos:
Note: Wrong feature branch PR. Sorry

Issue #3906, opened 2021-07-31
A MapViewProps interface attribute is out of date


hmartiins:
# Bug report

## Summary

A MapViewProps interface attribute is out of date

## Environment info

```
expo: ~41.0.1
react: 16.13.1
react-native-maps: 0.27.1
```

## Steps to reproduce

1. create a reference variable and assign the MapViewProps to its type (example: ```const refMap = useRef<MapViewProps>(null);```;
2. when trying to access the current.region.latitude attribute value it returns undefined;
3. however, when I try to access the attribute value using .current .__lastRegion.latitude, then yes it returns the correct value.

```tsx
console.log(refMap.current.region.latitude) // Return undefined
console.log(refMap.current.__lastRegion.latitude) // Return latitude
```

#### Describe what you expected to happen:

I was expecting the refMap.current.region.latitude attribute to return the MapView's latitude to me, but it returned undefined, and when I logged in refMap.current, then yes I realized that inside it the right attribute would be refMap. .current__lastRegion.latitude

#### Describe what you actually happens:

When using refMap.current.__lastRegion.latitude, my Eslint returns a typing error, because in the index.d.ts file _lastRegion is not defined, only region that is defined there. As you understand it, the __lastRegion attribute should be region according to the MapViewProps type file

![2021-07-30-225342_1366x768_scrot](https://user-images.githubusercontent.com/51277667/127725783-d42a1b0a-5015-4767-95c6-2d85e3d3d257.png)

## Reproducible sample code

Excerpt from where the error is in the code:

```tsx
import React, { ReactElement, useRef, useState } from 'react';
import { View } from 'react-native';

import MapView, { Marker, MapViewProps } from 'react-native-maps';

export function App(): ReactElement {
const refMap = useRef<MapViewProps>(null);
const [viewHeight, setViewHeight] = useState<number>();
const [viewWidth, setViewWidth] = useState<number>();

async function onMapChange() {
  // Here the object is undefined
  const latitude = refMap.current.region.latitude;
  // This correct code would be below, however, I get the error "Property '__lastRegion' does not exist on type 'MapViewProps'."
  // const zoom = refMap.current.__lastRegion.latitude;
}
```

Snack link: [https://snack.expo.dev/taj_4vzbV](https://snack.expo.dev/taj_4vzbV)

Simon-TechForm:
A couple of comments on this:

[...]

PR #3967, opened 2021-09-23
fix: add __lastRegion to MapView class declaration

Simon-TechForm:

No

#3906

### How did you test this PR?

Verified that __lastRegion is present on a MapView ref by printing its content, as well as looking through the source code of MapView.js.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -334,6 +334,8 @@ declare module 'react-native-maps' {
     pointForCoordinate(coordinate: LatLng): Promise<Point>;
     coordinateForPoint(point: Point): Promise<LatLng>;
     setIndoorActiveLevelIndex(index: number): void;
+
+    __lastRegion?: Region;
   }
 
   export class Animated extends MapView {}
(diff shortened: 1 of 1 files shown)


carlosgallo0 (approved):

christopherdro:
@Simon-TechForm Should we move forward with setting `__lastRegion` as private?

Simon-TechForm:
> @Simon-TechForm Should we move forward with setting `__lastRegion` as private?

Absolutely. I noticed that correcting mismatched types was a recurring event in this repo, and while debugging some things / familiarizing myself with the code, I became quite frustrated by the fact that the source code isn't typed, so I actually began doing a typescript rewrite. So if there's interest in transitioning to typescript (I hope there is), I'm looking to submit an initial pull request for it, hopefully within the next few days, but possibly not until next week. This is also the reason why I've been quiet on here for the last few weeks. The time I had to spent on this project have been used on doing the rewrite.

christopherdro:
@Simon-TechForm There is def interest in transitioning to TS and I appreciate you taking the time to start that process.

Issue #1132, opened 2017-03-15
IOS OnMapPress called when press the marker


felangga:
This problem using Google Maps and doesn't occurred on Android.

WernlingTan:
In my case, that is same.
Who can help me?

saadmutahar:
@hysan 
The same error is occurred in latest version of maps.
When I press the marker, it triggers the onPress of parent Maps also. Can you please have a look for this.

I am uses these versions. 

"react-native": "0.59.8",
"react-native-maps": "0.25.0"

Thanks

saadmutahar:
It's fixed the issue for me, As I forgot to add the stopPropagation event on marker click. 

Thanks.

mmamani93:
Hi there, I am trying to create a custom behavior when any part of the map is pressed:
In case a Marker is pressed, I want to change the color of the marker. And the color turns back to normal when any other part of the map is pressed. This is working fine in iOS but not in Android.
Im not english native so Im having some difficulties to understand this thread completely, does it mean that `onPress` events will not be triggered in Android at any cost?
Is there any workaround to handle Markers press for both platforms?

Thanks in advance!

PR #3985, opened 2021-10-20
Typescript rewrite

Simon-TechForm:

No

None

### How did you test this PR?

Source code evaluation and manual testing.

### Misc comments

A full transition to TS.

This took a quite a while to do, so reviewing probably will as well.

This fixes numerous (I stopped counting) TS errors, mostly in callback types, and probably also introduces some (I did rigorous examination of every component, but I surely missed something), and as such, this _might_ introduce new errors.
Regardless, this should lead the way for a new code base which doesn't require manual examination of the d.ts file and proptypes each time the source code is changed.

We could also look into generating docs from JSDoc, to further reduce the sources of documentation? I have no personal experience with this, but reducing the different parts of documentation that needs manual work would be great.

The documentation currently uses `@platform iOS` and `@platform Android` without further description, but as iOS supports both Google Maps and Apple Maps, this isn't enough. This pull request seeks to address this as well, adding a thorough platform description to each component prop.

Please feel free to chime in regarding type placement and so on. This probably needs some refactoring.

I still need to do some testing on this as well, but the more the merrier, so I decided to push this now.

Testing:
1. Clone this pull request
2. Run the newly added build script, eg: yarn build
3. copy and replace dist/index.js, dist/index.d.ts, dist/types.js, dist/types.d.ts & dist/components folder into your node_modules/react-native-maps

BREAKING CHANGES:
1. drops supports for react-native < 0.63.4
2. Remove animateToNavigation, animateToCoordinate, animateToViewingAngle and animateToBearing.
3. Removes legacy iOS snapshot method
4. Components are no longer re-exported under MapView.


Simon-TechForm:
Still need to handle #1132 somehow.

On iOS, every event can basically be of any event type.

christopherdro:
> I get what you are trying to achieve, but I don't really think it's possible. As soon as you rename the files from .js to .ts, git(hub) will track it as though you deleted all the .js files and added all the .ts files.

This might of been a thing in the past but Github now tracks the file change and shows the diff as normal. Take a look at this diff on a quick test I did. https://github.com/react-native-maps/react-native-maps/compare/test-ts?expand=1

christopherdro:
@Simon-TechForm Any updates here? I think another reason for updating the existing files is that it would reduce the number of steps to test. Ideally, we could just check out your branch and run the example or test existing applications by referencing your branch instead of the latest release.

Simon-TechForm:
> @Simon-TechForm Any updates here? I think another reason for updating the existing files is that it would reduce the number of steps to test. Ideally, we could just check out your branch and run the example or test existing applications by referencing your branch instead of the latest release.

I had a couple of busy weeks that didn't allow me to work on this, but I'm expecting to be able to look at this again within this week. I will get back to you.

Simon-TechForm:
@christopherdro can you specify how you would like to handle gitIgnore, ts outDir etc.? I've added a crude solution for now, but I'm not at all accustomed to having .ts files in the lib/dist folder.

lib/components/AnimatedRegion.ts
```diff
@@ -26,7 +29,7 @@ const defaultValues = {
 
 let _uniqueId = 1;
 
-const getAnimatedValue = (valueIn, fallback) => {
+const getAnimatedValue = (valueIn: unknown, fallback: number) => {
```
jasonmichaels, line 32: Technically, this isn't `unknown`. This would be more

```js
type TValueIn = number | Animated.Value;

const getAnimatedValue = (valueIn?: TValueIn, fallback: number) => { ... };
```

This is why we conditionally use the `fallback`.

Simon-TechForm: You are correct. I didn't make it optional however, so that we can preserve the order of the parameters.

Simon-TechForm:
See #4057

PR #4002, opened 2021-11-08
[android] Fix applicationId to com.example

marqroldan:
This was causing an "error type 3" when attempting to do `npm run android`

n/a

n/a

### How did you test this PR?

Android
Simply do `npm run android` after installing the required dependencies

### DIFF
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -124,7 +124,7 @@ android {
     compileSdkVersion rootProject.ext.compileSdkVersion
 
     defaultConfig {
-        applicationId "com.airbnb.android.react.maps.example"
+        applicationId "com.example"
         minSdkVersion rootProject.ext.minSdkVersion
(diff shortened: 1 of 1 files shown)


christopherdro (approved):

isidoro98:
Hey @marqroldan @christopherdro This change seems to be breaking the example app. Likely because the package name is used to restrict the usage of the API key. When using this new applicationId the example app maps are blank, but when using the older one it works.

christopherdro:
@isidoro98 yea you’ll have to update the package name that’s restricted or turn it off. The change in this PR was required for me as well to get the example app to run

isidoro98:
> @isidoro98 yea you’ll have to update the package name that’s restricted or turn it off. The change in this PR was required for me as well to get the example app to run 

Shouldn't this be added to the example app readme? So that anyone trying to run it can change it for it to work?

PR #4039, opened 2021-12-16
Update index.d.ts

JimTeva:
l.83 & l.100: useNativeDriver is required in type 'AnimationConfig'.

l.103-->107: Incorrect use of class Animated instead of RNAnimated.

I've checked PR https://github.com/react-native-maps/react-native-maps/pull/3985 but could see the same modifications, so no.

None as I didn't opened an issue for that but encountered this Type errors in my project.

### How did you test this PR?

I wasn't able to test but changes are minor.
However it will introduce the need to add prop useNativeDriver as a requirement in the modified functions.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -80,7 +80,7 @@ declare module 'react-native-maps' {
     easing?: (value: number) => number;
     duration?: number;
     delay?: number;
-    useNativeDriver?: boolean;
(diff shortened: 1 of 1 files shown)


JimTeva:
@rborn

christopherdro (approved):

Issue #3991, opened 2021-10-27
New Google Maps SDK Renderer


Borduhh:
# Feature request

## Why it is needed

All devices on Android 5.0 (Lollipop) or later with 2 GB or more of data storage will use the new renderer. All other devices will continue to use the legacy renderer until further notice. SDK Progressive Rollout beginning March 2022.

## Possible implementation

Upgrade Maps SDK to the new renderer (https://developers.google.com/maps/documentation/android-sdk/renderer)

anirudhabhurke:
+1 👍

salah-ghanim:
I will take this issue, it looks simple to implement and for now I would just expose a prop to choose the renderer type.

the only issue I see is that the SDK demo shows initialization on application create, so it might be too late if we do it via prop (when view is loaded).

pke:
Is there any action required at all here? This repo still uses Maps SDK 5.1 so its not affected by Google's changes.
Or am I missing something here?

SuhilMohammad-rogers:
Everytime i add a new marker on the MapView, the MapView seems to rerender, it looks like its zooming in and out. i wonder if thats happening because this package is currently using the legacy renderer

mak12:
Any timeline for this implementation ? current implementation isn't accepting `playServicesVersion:18.0.0` from project level configurations.

siddharth-kt:
any update on this ?

christopherdro:
@Borduhh @siddharth-kt @pke @mak12 @dishantwalia Please take a look at this open PR #4055. 
It would be very helpful if you can test the proposed changes with your app and report back with any issues you come across.

Simon-TechForm:
The new renderer can be now be enabled starting from [v0.31.0-beta.2](https://github.com/react-native-maps/react-native-maps/releases/tag/v0.31.0-beta.2) which is released on the beta channel of npm. Props to @isidoro98.

PR #4045, opened 2021-12-21
[android]: New Google Maps SDK Renderer

isidoro98:

No

This PR provides access to the new Google Maps renderer on Android, as raised on #3991.

The new renderer can be enabled using the android prop `useNewRenderer`. The only thing to bear in mind is that the application will use the renderer of the first map that is used in your application, regardless of what this prop is past that first render. i.e. If the first map that shows in your application uses the latest renderer all other maps will use it. If the first map that shows in your application you uses the legacy renderer all other maps will use that one.

An alternative approach to this PR would be to have a method that you can call on your App.js to set the renderer at app start time. I've implemented both approaches but decided to go with the prop in the end. Ultimately this is a design decision and would like to get your feedback.

### How did you test this PR?

Tested it on a Samsung Galaxy Note 9. Also tested in on emulators that don't meet [google requirements](https://developers.google.com/maps/documentation/android-sdk/renderer) to verify that the legacy renderer was used as fallback.

### DIFF
--- a/docs/mapview.md
+++ b/docs/mapview.md
@@ -12,6 +12,7 @@
 | `mapPadding` | `EdgePadding` |  | Adds custom padding to each side of the map. Useful when map elements/markers are obscured.
 | `paddingAdjustmentBehavior` | 'always' \| 'automatic' \| 'never' | 'never' | Indicates how/when to affect padding with safe area insets (`GoogleMaps` in iOS only)
 | `liteMode` | `Boolean` | `false` | Enable lite mode. **Note**: Android only.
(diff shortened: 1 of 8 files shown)


isidoro98:
@christopherdro thinking of changing this to be a function that users can call when they first start the app. The prop is a bit misleading because the use of the new renderer can only be set once.

I've already implemented it and tested it. Is ready on my local, what do you think? Which approach you think will be better?

```javascript
// User would call this on their App.js
// What you think about the function name? Any other suggestion?
useLatestRenderer();
```

Issue #245, opened 2016-05-22
Map size can't be 0. Most likely layout has not yet been occurred.


jaysson:
Occassionally I get this error:
![screenshot_2016-05-22-23-58-08](https://cloud.githubusercontent.com/assets/1563862/15456694/774c068e-2095-11e6-81a7-d5fc482e950e.png)

But this does not happen always.

christopherdro:
@jaysson Could you provide an example? My guess is it's some sort of race condition.
Is it possible it possible that your trying to set the original region of the app with fetched data.

A simple fix would be to make sure you have those coordinates before rendering the map.

jaysson:
Yes, I am fetching the user location and animating the map to that region.

```
<MapView
                    style={ styles.map }
                    initialRegion={{
                        latitude: 12.97162345,
                        longitude: 77.59462345,
                        latitudeDelta: 0.0461,
                        longitudeDelta: 0.02105
                    }}
                    ref={(gMap) => this.googleMap = gMap}
                ><MapView.Marker draggable coordinate={this.state.coordinates} onDragEnd={this.markerDragged}/>
```
In constructor:
```
navigator.geolocation.getCurrentPosition(
            (position) => {
                Geocoder.geocodePosition({lat: position.coords.latitude, lng: position.coords.longitude}).then((results)=> {
                    let result = results[0];
                    Address.saveFromMap(result);
                    this.setState({
                        addressText: result.formattedAddress,
                        address: result
                    })
                });
                this.setState({
                    coordinates: {
                        latitude: position.coords.latitude, longitude: position.coords.longitude
                    }
                });
                this.googleMap.animateToRegion({
                    latitude: position.coords.latitude,
                    longitude: position.coords.longitude,
                    latitudeDelta: 0.02305,
                    longitudeDelta: 0.01075
                })
            },
            (error) => alert(error),
            {timeout: 60000, maximumAge: 10000}
        );
```

adam-jakiela:
+1

iiitmahesh:
@jaysson have you solved this. I have same issue..!

manikantag:
For now I m setting the width & height. 

```
< MapView style={{width: Dimensions.get('window').width, height: Dimensions.get('window').height}}>
```

This only requires if I've `MapView` in the route navigation directly. If I render it on any event, it is not required. So, I think we need either the render the map delayed or keep the dimensions as above. `flex:1` is not working, though.

nonameolsson:
[...]

PR #4047, opened 2021-12-24
Skip calling newLatLngBounds if non positive height or width

frankitox:

There's an already merged related PR, [#1521](https://github.com/react-native-maps/react-native-maps/pull/1521). Which is related to the issue #245. The initial PR did not fix the problem for me.

Avoid throwing an exception by avoiding a call to `newLatLngBounds`. The initial error only appears to me just before navigating with react-nativation and disabled animations.

### How did you test this PR?

I tested with an emulator and just editing the .java file (change only affects Android).

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
@@ -749,9 +749,7 @@ public class AirMapView extends MapView implements GoogleMap.InfoWindowAdapter,
 
       //fix for https://github.com/react-native-maps/react-native-maps/issues/245,
       //it's not guaranteed the passed-in height and width would be greater than 0.
-      if (width <= 0 || height <= 0) {
-        map.moveCamera(CameraUpdateFactory.newLatLngBounds(boundsToMove, 0));
-      } else {
+      if (width > 0 && height > 0) {
         map.moveCamera(CameraUpdateFactory.newLatLngBounds(boundsToMove, width, height, 0));
       }
 


lib/android/src/main/java/com/airbnb/android/react/maps/AirMapView.java
```diff
@@ -749,9 +749,7 @@ public void updateExtraData(Object extraData) {
 
       //fix for https://github.com/react-native-maps/react-native-maps/issues/245,
       //it's not guaranteed the passed-in height and width would be greater than 0.
-      if (width <= 0 || height <= 0) {
-        map.moveCamera(CameraUpdateFactory.newLatLngBounds(boundsToMove, 0));
-      } else {
+      if (width > 0 && height > 0) {
```
christopherdro, line 752: I'm not quite sure I follow how this fixes the issue. It looks like the call to move the camera will be ignored altogether. What value are you getting for width and height?

frankitox: The problem is that if `width` or `height` is 0 then `CameraUpdateFactory.newLatLngBounds` throws an exception. So my idea is simply skip the call to moveCamera in that case. It does fix it on my case (width and height are 0), but I don't know if it breaks somebody else's workflow.

frankitox: This is the exception
image

christopherdro: Have you taken a look at some of the Stack Overflow posts that discuss the same issue? https://stackoverflow.com/questions/47726904/error-using-newlatlngboundslatlngbounds-int-map-size-cant-be-zero-usi

christopherdro: https://lifesaver.codes/answer/error-using-newlatlngbounds-int-int-int)-map-size-cant-be-0

frankitox: I didn't! I've only checked #245. Well those two links try 2 solutions
1. Setting width/height or some css to force a non zero size map (not an option for what I'm doing)
2. Using `onMapReady`, this one I think won't work in my case, because the map starts with a proper size then because of some external factor (navigation hides it) I get an empty map, and that's when the error is triggered.

I will give a shot to 2, but it would be nice to just not throw in that particular case.

The vars `height` and `width` refer to the map height and width? That's what I'm assuming.

christopherdro: Yes most likely, or the parent container. If neither solution works I suggest using patch-package to make the required change locally for your app.

PR #4055, opened 2022-01-03
feat(android): add support for new google maps renderer

isidoro98:

Yes, #4045. But this PR takes a different approach. On #4045 The new renderer is set using a prop of `MapView`. This approach might be misleading as the renderer can only be set once in our application. Thus changing the value of this prop doesn't modify the renderer that our application will continue to use.

Instead this PR allows the user to opt in to the new renderer by calling `enableLatestRenderer` on the entry file of their application. 

This function can also be called at any other point to return the renderer that is being used, but it won't change it after the first map has been rendered.

#3991 

### How did you test this PR?

Tested on a Samsung Galaxy Note 9


christopherdro:
@isidoro98 Thank you for this. I definitely like this approach much better. I was gonna suggest to exposing the option on the native side but this might be a better approach since some users might be using this within a managed Expo workflow.

salah-ghanim (approved):
Thanks @isidoro98 this approach definitely have some advantages. and better than the old suggestion.


with that being said an environment variable or a javascript config (in package.json if possible) would be a better implementation. since it's awkward to add such a function call at the top of the project / main and to expect other contributers to understand the need for it.


but it's a decent solution, and hopefully google will provide better apis in the future releases.

salah-ghanim (approved):

salah-ghanim:
@@christopherdro feel free to merge

fakeheal:
I would suggest renaming `useLatestRenderer` to `enableLatestRenderer` or something else that doesn't start with `use...`, because in React context it can be a bit misleading (mainly to IDEs).

salah-ghanim:
I agree with @fakeheal improvements suggestions
1. preferLatestRenderer is a much better name
2. on iOS a resolved promise should be returned, that way we have consistent API regardless of the platform.

isidoro98:
> I agree with @fakeheal improvements suggestions
> 
> 1. preferLatestRenderer is a much better name
> 2. on iOS a resolved promise should be returned, that way we have consistent API regardless of the platform.

Regarding 2, at the moment iOS would return undefined, what would you want it to return? Undefined makes the most sense imo.

gillpeacegood:
I am trying this and had to change the android build.gradle 

```
android {
  compileSdkVersion safeExtGet('compileSdkVersion', 28)

   compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
```

Mhd-Emad:
Hi @isidoro98 
Thank you so much for your effort on this,
I tried your code and I got this error:

> error: lambda expressions are not supported in -source 7
>          MapsInitializer.initialize(context, MapsInitializer.Renderer.LATEST, (MapsInitializer.Renderer renderer) -> {
> (use -source 8 or higher to enable lambda expressions)

Can you advise, please?

isidoro98:
> Hi @isidoro98 
> 
> Thank you so much for your effort on this,
> 
> I tried your code and I got this error:
> 
> 
> 
> > error: lambda expressions are not supported in -source 7
> 
> >          MapsInitializer.initialize(context, MapsInitializer.Renderer.LATEST, (MapsInitializer.Renderer renderer) -> {
> 
> > (use -source 8 or higher to enable lambda expressions)
> 
> 
> 
> Can you advise, please?

Could you check your Java version? Looks like you're using Java 7, but need Java 8

botondus:
Any ETA on getting this merged in and released?

satya-imaginea (dismissed):

satya-imaginea:
Would someone please approve this PR, which need to be used by June 22 as per recent mails.

<blockquote type="cite" style="color: rgb(34, 34, 34); font-family: Arial, Helvetica, sans-serif; font-size: small; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><div dir="ltr">

Some of your projects use Maps SDK for Android. Try the new upgrade to test the improved renderer and other features.Hello Google Maps Platform Customer,We're writing to remind you that a major update of Maps SDK for Android (version 18.0.x), available via Google Play services, features a new map renderer that will become the default renderer for Android devices* through a progressive rollout, starting in June 2022 at the earliest.Once the new renderer becomes the default, your apps will use the new renderer whether or not they were built against Maps SDK for Android version 18.0.x.Update to the new map renderer (SDK latest version: 18.0.2) now to test and deploy the new SDK and renderer with your apps before the new renderer becomes the default for Android devices*.We’ve improved how we serve tiles to your apps and render the map on client devices with the new SDK. Therefore, Android devices will benefit from:Reduced network load, processing demand, and memory consumption.Improved gesture handling for better animations, plus smoother panning and zooming.More fluid transitions and clearly positioned map labels.A more stable and improved user experience for future map innovations on Android.The new SDK also includes the general availability of Cloud-based maps styling on mobile. Usage of Cloud-based maps styling for a map in Maps SDK for Android is billed with the Dynamic Maps SKU. For more information on pricing and included features, see our Cloud Customization documentation.*All devices on Android 5.0 (Lollipop) or later with 2 GB or more of data storage will use the new renderer. All other devices will continue to use the legacy renderer until further notice.What do I need to do?To anticipate and avoid issues with your apps we advise you to:Specify version 18.0.x in your dependencies (example).Specify the new renderer in your code (instructions).Test, build, and release a new version of your app based on the new renderer, before the beginning of the new SDK progressive rollout in June 2022. See our New Map Renderer documentation for instructions on how to use the new renderer.If you encounter any issues in your apps with the new SDK or the new renderer, please check our Support Options for Maps SDK for Android, and consider pausing the apps’ deployment to users until the issues are resolved.

ruanblimaig:
Does this PR lack something to be approved?

I tested this update, but I had problems using it on android. It gave an error related to the java version, even with a version greater than 8. To fix it I needed to add the following code snippet in android/build.gradlew:

`compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }`

Simon-TechForm:
We experienced a couple of things while testing the new renderer (would love to hear feedback from anybody else regarding this):

- Labels on the map (e.g. city names) are POI's and triggers the `onPoiClick` callback.
- The `onPoiClick` callback takes precedence over `onMapClick` and doesn't propagate.
- The `onPoiClick` callback takes precedence over `onMarkerClick` (At least for customMarkerViews) and doesn't propagate.

These are breaking changes (for us) as some markers are practically unpressable when rendered on top of labels and `onMapClick` isn't triggered when clicking the map on top of labels.
Not registrering the `onPoiClickListener` (natively) seems to allow the event to propagate to `onMapClick` and `onMarkerClick`.

Whether expected behaviour or not, this should most likely be documented in the PR.

phinfqasia:
Are there any concerns that we need to do for the testing? I'm quite sure that PR needs to merge for anyone that currently using this library.

Simon-TechForm:
@phinfqasia Could you test if you experience the same as I mentioned in the comment just above yours? :)

Simon-TechForm:
@isidoro98 @christopherdro I'm happy with this approach of enabling the renderer as well, and enabling it seems to be working as expected. I'm okay merging this, if we can do the following:

- Update the play services dependencies to the latest versions as of today (there's been a couple of releases since this PR)
- Update the doc with the new rollout date, which has been changed to June instead of March
- Update docs to include a warning that there’s a risk that enabling the new renderer might cause behavioural changes and maybe link to the comment I made above? Or even better, rephrase and include the comments directly in the docs

@isidoro98 would you be willing/able to update the PR with these changes?

Simon-TechForm (approved):

dishantwalia:
@Simon-TechForm when it will be available as an official release, any tentative date?

stockhuman:
> I tested this update, but I had problems using it on android. It gave an error related to the java version, even with a version greater than 8. To fix it I needed to add the following code snippet in android/build.gradlew:

```gradle
android {
  compileSdkVersion safeExtGet('compileSdkVersion', 28)

  defaultConfig {
    minSdkVersion safeExtGet('minSdkVersion', 21)
    targetSdkVersion safeExtGet('targetSdkVersion', 27)
  }
  compileOptions {
    sourceCompatibility JavaVersion.VERSION_1_8
    targetCompatibility JavaVersion.VERSION_1_8
  }
}
```

I too had to hack in this to `lib/android/build.gradle`, perhaps it's worth including?

fullstack0103:
hi, everyone
How can I implement new renderer in react native project?
Thanks

dishantwalia:
> hi, everyone How can I implement new renderer in react native project? Thanks

To opt in to the new renderer add the following code in your entry file (e.g. App.js):

import { enableLatestRenderer } from 'react-native-maps';

enableLatestRenderer();

https://github.com/react-native-maps/react-native-maps/blob/4a9fdf5f0d52519ce8b17f15fd35b81cde197701/docs/installation.md#using-the-new-google-maps-renderer

PR #4084, opened 2022-01-31
fix(ios): enable isPreselected for default marker

NavigilOMantere:

There is no existing PR for this.

#4085 

### How did you test this PR?

1. Create a draggable marker with the default pin:
```
import { Marker } from 'react-native-maps';
<Marker draggable isPreselected />
```
2. Build the react-native app for iOS, run it on the device (iPhone 8).
3. Confirm that the pin is now draggable immediately, without an extra tap to select.

Without this PR, the pin would require an extra press to select before dragging.


Simon-TechForm:
@NavigilOMantere can you update your branch to fix the conflicts? :) When it's fixed I will be happy to test it out and get it merged.

abehoffman:
Can this be reopened please?

PR #4096, opened 2022-02-13
feat: migrate to deprecated-react-native-prop-types

Simon-TechForm:

No.

`Image.propTypes`, `Text.propTypes`, `TextInput.propTypes`, `ColorPropType`, `EdgeInsetsPropType`, `PointPropType`, and `ViewPropTypes` from react-native are deprecated, and starting from react-native@0.68 using them will print a warning. They have already been removed in the main branch, so react-native@0.69 likely won't include them at all.

See: [react-native@0.68 changelog](https://github.com/facebook/react-native/blob/66419ed8f5be4e0efca4e0f261d902607408f29c/CHANGELOG.md#changed)

Suggested solution is to use the `deprecated-react-native-prop-types` package instead, which this PR does.

### How did you test this PR?

Installed react-native-maps in a react-native@0.68.0-rc.1 project, resulting in a lot of warning messages. After these changes, the warnings are no longer printed and the library continue to work as expected.

### DIFF
--- a/lib/components/Geojson.js
+++ b/lib/components/Geojson.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import Marker from './MapMarker';
 import Polyline from './MapPolyline';
 import Polygon from './MapPolygon';
-import { ColorPropType } from 'react-native';
+import { ColorPropType } from 'deprecated-react-native-prop-types';
 
 const propTypes = {
   /**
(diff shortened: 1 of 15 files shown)


Simon-TechForm:
@christopherdro could you look into this?

christopherdro (approved):

Issue #4112, opened 2022-03-09
[iOS] - add details.isGesture prop to onRegionChangeComplete callback


9r4ik:
# Feature request

## Why it is needed

**onRegionChangeComplete** note from [mapview doc](https://github.com/react-native-maps/react-native-maps/blob/master/docs/mapview.md)
```
 isGesture property indicates if the move was from the user (true) or an animation (false). Note: isGesture is supported by Google Maps only. 
 ```
Now iOS MapView with MapKit provider  trigger **onRegionChangeComplete** multiple times while user pan/zoom/pinch map.

iOS MapView with google provide trigger **onRegionChangeComplete** ones when user remove fingers from screen

Without this feature developers could not detect end of user interaction

---
#3805 #3818 
## Possible implementation

With this feature developers could detect end of user interaction

### Code sample

```js
onRegionChangeComplete={(new_region, details) => {
          if (details?.isGesture) {
            console.log('fetch new data with new region');
          }
        }}
```

aliwaqar981:
It would be great if this can be implemented for iOS MapView. Eagerly looking for this

RyanTG:
I don't think this should be marked stale just yet. There's an open PR with a (potential) solution.

eduardorfreitas93:
PR https://github.com/react-native-maps/react-native-maps/pull/4272 with a solution

PR #4113, opened 2022-03-09
add isGesture prop to onRegionChangeComplete callback on iOS MapKit p…

9r4ik:
…rovider

no

#4112

### How did you test this PR?

Test on 3 iOS devices and 1 simulator


9r4ik:
You need to declare **OnPanDrag** in **MapView** props otherwise {"isGesture": true} do not trigger

RyanTG:
Eagerly awaiting this functionality, so that we aren't forced to use google maps on iOS devices.

Issue #4149, opened 2022-03-28
Android 12 app crash with 0.31.0-beta.3 version - cannot find symbol variable Renderer


vkozinec:
Currently trying to run maps on android 12 and getting this error. Using standard Idk 11 from android studio. Does anyone have luck with building the app on android 12?

```
error: cannot find symbol
        MapsInitializer.initialize(context, MapsInitializer.Renderer.LATEST, (MapsInitializer.Renderer renderer) -> {
                                                           ^
  symbol:   variable Renderer
  location: class MapsInitializer
```

![Screenshot 2022-03-28 at 16 51 31](https://user-images.githubusercontent.com/570613/160425434-d2b23e24-384d-4888-a5b1-6767c3bc62d0.png)

## system info

```
System:
    OS: macOS 12.2.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 95.45 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.15 - /usr/local/bin/yarn
    npm: 8.5.5 - /usr/local/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.11.3 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 15.4, macOS 12.3, tvOS 15.4, watchOS 8.5
    Android SDK: Not Found
  IDEs:
    Android Studio: 2021.1 AI-211.7628.21.2111.8193401
    Xcode: 13.3/13E113 - /usr/bin/xcodebuild
  Languages:
    Java: Not Found
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2
    react-native: 0.67.2 => 0.67.2
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

## build.gradle
```
        buildToolsVersion = "30.0.2"
        minSdkVersion = 23
        compileSdkVersion = 31
        targetSdkVersion = 31
        ndkVersion = "21.4.7075529"
        playServicesVersion = "17.0.0"
        androidMapsUtilsVersion = "2.3.0"
```

## react native

```
"react-native": "0.67.2",
"react-native-maps": "0.31.0-beta.3",
```

Simon-TechForm:
Hi @vkozinec.

[...]

PR #4155, opened 2022-03-30
Readme fix googlepayservices

salah-ghanim:
No

https://github.com/react-native-maps/react-native-maps/issues/4149

### How did you test this PR?
No, this is updated based on Google Maps Docs
https://developers.google.com/android/guides/setup

Some questions you might want to think about:

Which platform (eg. Android/iOS) & Maps API (eg. Google/Apple) does this change affect, if any?

**Android**


Simon-TechForm:
Doing a bit of cleaning up - I believe the relevant changes from this is in #4150? :)

Issue #4160, opened 2022-04-01
Fails to compile using Expo EAS Build with 0.31.0-beta.2


jsjexpert:
# Bug report

## Environment info

`react-native info` output:

```bash
      react-native: "git+https://github.com/expo/react-native/archive/sdk-44.0.0.tar.gz"
      react: 17.0.1
      react-native-maps: 0.31.0-beta.2
```

## Steps to reproduce

1. Using Expo SDK 44
2. in App.tsx add the following lines to allow new Maps SDK 18 renderer to load:
`import { enableLatestRenderer } from "react-native-maps";
enableLatestRenderer();`
3. Replace Expo managed react-native-maps@0.29.4 with react-native-maps@0.31.0-beta.2
4. Run `eas build:android` to begin 

Describe what you expected to happen:

1. Expected successful build!

Describe what you actually happens:

Error during Expo EAS compilation:
Task :react-native-maps:compileReleaseJavaWithJavac FAILED
w: Detected multiple Kotlin daemon sessions at build/kotlin/sessions
[stderr] /home/expo/workingdir/build/node_modules/react-native-maps/lib/android/src/main/java/com/airbnb/android/react/maps/AirMapModule.java:368: error: lambda expressions are not supported in -source 1.7
[stderr]         MapsInitializer.initialize(context, MapsInitializer.Renderer.LATEST, (MapsInitializer.Renderer renderer) -> {
[stderr]                                                                                                                  ^
[stderr]   (use -source 8 or higher to enable lambda expressions)
[stderr] 1 error
[stderr] FAILURE: Build failed with an exception.
[stderr] * What went wrong:
[stderr] Execution failed for task ':react-native-maps:compileReleaseJavaWithJavac'.

Build failed an no binary produced.

Looks like the lambda/arrow function is not supported with Expo's build process.

Full gradle log from Expo.dev
`Running './gradlew :app:bundleRelease' in /home/expo/workingdir/build/android
Downloading https://services.gradle.org/distributions/gradle-6.9-all.zip
Unzipping /home/expo/.gradle/wrapper/dists/gradle-6.9-all/dooywd8nv05k16orzxge2b1bs/gradle-6.9-all.zip to /home/expo/.gradle/wrapper/dists/gradle-6.9-all/dooywd8nv05k16orzxge2b1bs
[...]

PR #4178, opened 2022-04-11
fix(android): remove use of lambda expression

Simon-TechForm:

No

 #4160

### How did you test this PR?

Ran an android project containing the changes.


salah-ghanim:
@Simon-TechForm I see androidx.annotation used twice in the source code only ... I would remove it and reduce dependencies for this project instead of adding more of it.

Simon-TechForm:
@salah-ghanim I'm okay streamlining the use of javax.annotation vs androidx.annotation, but in that case the proper solution imo would be to replace all uses of javax.annotation instead of androidx.annotation, as androidx.annotation is the default, and we would be fighting the Android Studio IDE going forward if use javax.annotation.

salah-ghanim (approved):
LGTM

PR #4182, opened 2022-04-11
feat(android): migrate to androidx.annotation

Simon-TechForm:

No

None

### How did you test this PR?

Ran a project with the proposed changes on Android

### DIFF
--- a/android/src/main/java/com/airbnb/android/react/maps/AirMapCalloutManager.java
+++ b/android/src/main/java/com/airbnb/android/react/maps/AirMapCalloutManager.java
@@ -1,5 +1,7 @@
 package com.airbnb.android.react.maps;
 
+import androidx.annotation.Nullable;
+
 import com.facebook.react.common.MapBuilder;
 import com.facebook.react.uimanager.LayoutShadowNode;
 import com.facebook.react.uimanager.ThemedReactContext;
@@ -7,8 +9,6 @@ import com.facebook.react.uimanager.ViewGroupManager;
 import com.facebook.react.uimanager.annotations.ReactProp;
 
 import java.util.Map;
-
-import javax.annotation.Nullable;
 
 public class AirMapCalloutManager extends ViewGroupManager<AirMapCallout> {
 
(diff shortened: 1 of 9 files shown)


salah-ghanim:
@Simon-TechForm unifying is definitely a step better than using two libraries for nullability.

but from my point of view I don't see the need to use nullability annotation at all, it make sense if this library exposes java APIs and then it hints to the library users what to expect from the library APIs.

But since all of this is abstracted via JS, I would simply remove all nullability annotations and make this lib just a bit leaner.

what do you think ?

Simon-TechForm:
@salah-ghanim I respectfully disagree :)

Annotations are extremely useful in the Java code for static code analysis, even when the methods aren't exposed as an API. They aren't only for third parties, but also for whoever might need to edit the block of code that includes the annotation.

Take the new renderer callback as an example:
```java
MapsInitializer.initialize(context, MapsInitializer.Renderer.LATEST, new OnMapsSdkInitializedCallback() {
  @Override
  public void onMapsSdkInitialized(@NonNull MapsInitializer.Renderer renderer) {
    Log.d("AirMapRenderer", renderer.toString());
    promise.resolve(renderer.toString());
  }
});
```

Removing the `@NonNull` annotation, how would the developer/static code analysis know that it's safe to call `toString()` on the variable? For all we know, it could be `null` and should be checked with `if(renderer != null)` before being accessed.

On top of that, Android Studio will even add warnings to the code if you remove the annotations. In the example above it will warn that `Not annotated parameter overrides @NonNull parameter` and suggest to add the annotation. So removing the annotations would add warnings all over the java code base.

Last but not least, importing the annotations wouldn't leave any meaningful impact on the size of the library. In fact, I think you will struggle to find a react-native project that doesn't already include androidx.annotation :)

salah-ghanim (approved):
@Simon-TechForm very convening arguments, if the RN template app is using it, I don't see why we should do less here.

I'm always a bit holding back / trying to reduce when it comes to complex dependencies setup and seeing the issues with google play services versions that are submitted I thought less dependencies mean less conflicts means less issues to handle.

PR #4239, opened 2022-05-15
fix(android): wrong coordinates on pressing polyline

ramsingit:
Fix for wrong coordinates on pressing polyline

No

This change will fix the wrong coordinates sent while pressing the polyline on Android platform

### How did you test this PR?

I have tested it on Android platform with the real device. The change is specific to Android, so haven't tested it on iOS


ramsingit:
@Simon-TechForm Hi, can you check and approve this PR if possible?

ramsingit:
@Simon-TechForm Please review and approve the changes. Thanks

Simon-TechForm:
LGTM, seems like it should have been fixed alongside #3966.

PR #4241, opened 2022-05-18
feat: cleanup & ts

Simon-TechForm:
**PLEASE DON'T SQUASH MERGE THIS - THE INDIVIDUAL COMMITS ARE IMPORTANT FOR THE CHANGELOG**

This pr includes a full ts-rewrite and a couple of breaking changes that made sense to make while rewriting the codebase. Please have a look through the commits for a full changelog.

This will mark the change to v1.0.0.

As doing a ts rewrite stalled a couple of times earlier due to the workload required to review it, I suggest we make good use of our beta channel, and release the changes on there. When the beta release is published, I suggest pinning an issue asking for people to help try out the new version.

@christopherdro thoughts? :)


tomasmozeris:
Hey, thanks for supporting this library. I was checking iOS initialRegion fix for google maps spotted few issues, used `example/examples/DefaultMarkers`:
- initialRegion is focused after first interaction is made
- having initialRegion and marker on map initialRegion acts like region prop, prevents to any map movement


https://user-images.githubusercontent.com/5544385/171189155-5ebd4a93-8f62-45e7-8145-2e65b6d187d8.mov

Simon-TechForm:
@tomasmozeris I appreciate you taking the time to test out the changes. I've pushed a commit that should fix the problem you are describing, that's included in the release mentioned above :)

PR #4267, opened 2022-06-09
fix(types): export types that were present before v1

hsource:

No

Many types are exported by the 0.31.1 of the library, but because we're now directly creating the `index.d.ts` from the actual Typescript, they're no longer exported.

This PR fixes this by:

1. Extracting all the types used by Props and callable functions into a `MapView.types.ts` and exporting those types
2. Exporting MapViewProps
3. Exporting all `sharedTypes.ts`
   - I moved the helper type `Modify` into a `sharedTypesInternal.ts` that's not exported
4. Exporting all the Props and inner components. The inner component types are useful for `React.createRef` and `useRef`

### How did you test this PR?

This is a pure typing change with no functionality change. I tested this by:

1. Creating a build of the package and installing it using `yarn add wanderlog/react-native-maps#407b2526384c6336f00ae094b518a34db9164a2c`
2. Build it and run it with the build of our app
   - We had to change a reference of MapEvent to PoiClickEvent, but otherwise it imported just fine

### DIFF
--- a/src/Geojson.tsx
+++ b/src/Geojson.tsx
@@ -10,14 +10,14 @@ import {
   Polygon,
   MultiPolygon,
 } from 'geojson';
-import Marker, {Props as MarkerProps} from './MapMarker';
-import {Props as PolygonProps} from './MapPolygon';
-import {Props as PolylineProps} from './MapPolyline';
+import Marker, {MapMarkerProps as MarkerProps} from './MapMarker';
+import {MapPolygonProps as PolygonProps} from './MapPolygon';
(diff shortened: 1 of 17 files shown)


Simon-TechForm (approved):

Simon-TechForm:
Hi @hsource 👋 

Thank you for putting this together. LGTM. If you could bring the branch up to date, I'm more than happy to merge this.

hsource:
@Simon-TechForm rebased and should be ready to merge

Simon-TechForm (approved):

Issue #4303, opened 2022-07-04
Wrong coordinates when pressing a polyline on iOS


filiphorvei:
### Summary

When pressing a polyline on Android, the coordinates of the first point in the polyline was sent to JavaScript in the onPress event. However, this was changed in PR #4239. I propose we make the same change on iOS so the different platforms are in sync.

### Reproducible sample code

```TSX
import MapView, { Polyline } from 'react-native-maps';

export default function App() {

    const coordinates = [
        { latitude: -74.0082, longitude: 40.7191 },
        { latitude: -74.0069, longitude: 40.7270 },
        { latitude: -73.9956, longitude: 40.7255 },
        { latitude: -73.9969, longitude: 40.7185 }
    ];

    function onPolylineClick(event) {
        let firstCoordinate = event.nativeEvent.coordinate;
        console.log(firstCoordinate === coordinates[0]);
        // -> always true on iOS, false on Android
    }

    return (
        <MapView style={{ flex: 1}}>
            <Polyline
                coordinates={coordinates}
                tappable={true}
                onPress={onPolylineClick} />
        </MapView>
    );
}
```

### Steps to reproduce

1. Create a map with a polyline.
2. Press the polyline.

### Expected result

The behaviour should be the same across platforms

### Actual result

On Android, the place pressed on the polyline is used. On iOS, the first coordinate is used.

### React Native Maps Version

0.31.1

### What platforms are you seeing the problem on?

iOS (Apple Maps)

### React Native Version

0.68.2

### Are you using Expo?

No

### Device(s)

All iOS devices

PR #4304, opened 2022-07-04
fix(ios): wrong coordinates on pressing polyline

filiphorvei:

No. This PR is essentially the iOS version of #4239

#4303

### How did you test this PR?

The change affects Apple Maps on iOS.

Test plan:
1. Create a map with a polyline.
2. Press the polyline.
3. The event object should contain the coordinates of the tap location, like the behaviour on Android.


Simon-TechForm:
Sounds reasonable :) Did you look into how Google Maps behaves on iOS?

PR #4272, opened 2022-06-14
add isGesture to onRegionChangeComplete iOS

eduardorfreitas93:
add isGesture prop to onRegionChangeComplete callback on iOS MapKit

### Does any other open PR do the same thing?

https://github.com/react-native-maps/react-native-maps/pull/4113

### What issue is this PR fixing?

https://github.com/react-native-maps/react-native-maps/issues/4112

### How did you test this PR?

Test on 1 iOS(15.5) devices and 1 simulator

PR #4322, opened 2022-07-13
add isGesture to onRegionChangeComplete iOS

eduardorfreitas93:
Reopen this PR by adding isGesture to onRegionChangeComplete callback on iOS MapKit

https://github.com/react-native-maps/react-native-maps/pull/4113
https://github.com/react-native-maps/react-native-maps/pull/4272

https://github.com/react-native-maps/react-native-maps/issues/4112

### How did you test this PR?

Tested in production and 1 simulator iOS


eduardorfreitas93:
[RyanTG](https://github.com/RyanTG) If they don't approve this PR you can try using
https://javascript.plainenglish.io/live-npm-package-patching-with-patch-package-b8f30d47779c
https://gist.github.com/eduardorfreitas93/eb1f37784a936b4bd618ffd209fb72db

Simon-TechForm:
Hi @eduardorfreitas93 👋. Thank you for opening a new PR after the refactor.

As far as I can tell, this PR adds`isGesture` to the `onRegionChange` with a value that's always false? That seems undesirable :) If you could look into that, I will be happy to review this again :)

eduardorfreitas93:
@Simon-TechForm Hi, it adds isGesture true in the return of onRegionChangeComplete, but I can see if I implement it in onRegionChange too

Simon-TechForm:
It's not a necessity to implement the `isGesture` functionality for `onRegionChange` for me to merge this. But we can't have it return `isGesture=false` even when it should be true :)

eduardorfreitas93:
With this implementation I return isGesture true when the user makes the gesture to navigate the map, when the navigation on the map is automatic, for example zoom or recenter it has to return false, at least that was the functionality I needed

ios/AirMaps/AIRMapManager.m
```diff
@@ -964,8 +964,10 @@ - (void)mapView:(AIRMap *)mapView regionDidChangeAnimated:(__unused BOOL)animate
 
     // Don't send region did change events until map has
     // started rendering, as these won't represent the final location
-    if (mapView.hasStartedRendering) {
-        [self _emitRegionChangeEvent:mapView continuous:NO];
+    if (mapView.hasStartedRendering && animated) {
+        [self _emitRegionChangeEvent:mapView continuous:NO isGesture:NO];
+    } else if (mapView.hasStartedRendering && !animated) {
+        [self _emitRegionChangeEvent:mapView continuous:NO isGesture:YES];
```
Simon-TechForm, line 970: Since `animated` is used, the `__unused` macro should be removed from it.

Simon-TechForm (changes requested):

eduardorfreitas93:
@Simon-TechForm I made the changes

Simon-TechForm:
@eduardorfreitas93 I'm afraid this approach isn't going to work either. If you swipe the map (meaning the region will keep changing for a bit after you release your gesture), `isGesture` will change to false even though it should be true.

I'm afraid that I'm too occupied atm to help you out with a solution, but I would suggest putting together a set of test cases that you evaluate when coming up with a new approach, e.g.:

| Case isGesture                           | Expected onRegionChange             | Actual onRegionChange          | Expected onRegionChangeComplete | Actual onRegionChangeComplete | Result |
|------------------------------------------|-------------------------------------|--------------------------------|---------------------------------|-------------------------------|--------|
| animateToRegion                          | false                               |                                | false                           |                               |        |
| animateToRegion intercepted by gesture   | false until interception, then true |                                | true                            |                               |        |
| Drag map                                 | true                                |                                | true                            |                               | Good   |
| Swipe map                                | true, also when gesture is released | true while gesture, then false | true                            | false                         | Bad    |
| Swipe map intercepted by animateToRegion | True until interception, then false |                                | false                           |                               |        |

These are just off the top of my head, you could come up with more/others. Also make sure to check out Google Maps to match the behaviour.

Simon-TechForm:
I'll close this for now and re-open when we have a working solution :)

wmonecke:
I think this is urgently needed. Have you guys found a workaround?

RyanTG:
I don’t have a workaround. I created a “Discussion” for this here: https://github.com/react-native-maps/react-native-maps/discussions/4440

But it’s not gaining traction.

wmonecke:
@RyanTG Awesome, thank you!

samih-sghier:
This would be so useful haven't found a workaround

Issue #4130, opened 2022-03-16
App crash on ios 15.4


anija:
# Bug report

## Summary
App started to crash with ios 15.4

## Environment info

react-native: 0.66.4
react: 17.0.2
react-native-maps: 0.30.1

Target Platform: iOS - 15.4 Simulator/iPhone 13
Using GoogleMaps

`react-native info` output:

System:
    OS: macOS 12.2.1
    CPU: (8) arm64 Apple M1
    Memory: 1.02 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.2 - ~/.nvm/versions/node/v16.13.2/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v16.13.2/bin/yarn
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.2/bin/npm
    Watchman: 2022.01.24.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.2 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 15.4, macOS 12.3, tvOS 15.4, watchOS 8.5
    Android SDK: Not Found
  IDEs:
    Android Studio: 2020.3 AI-203.7717.56.2031.7678000
    Xcode: 13.3/13E113 - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.12 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2 
    react-native: 0.66.4 => 0.66.4 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

## Steps to reproduce

1. Create a component with:
```import React, { useEffect, useState } from 'react';
import { StyleSheet } from 'react-native';
import MapView, { PROVIDER_GOOGLE } from 'react-native-maps';

<MapView
  provider={PROVIDER_GOOGLE}
  style={StyleSheet.absoluteFill}
/>
```
2. Display the component
3. After a few seconds, the app crashes

Describe what you expected to happen:

1. The app don't crash

Everything works fine when provider is not set.

Crash report: [crash_report.txt](https://github.com/react-native-maps/react-native-maps/files/8268928/crash_report.txt)

Simon-TechForm:
I'm unable to reproduce this. Is this happening in a clean react native project?

[...]

PR #4334, opened 2022-07-18
feat: allow metal renderer to be enabled

jbrodriguez:
No

(please answer here)

On iOS 15.4, the google provider makes the application crash

[(please link the issue here)
](https://github.com/react-native-maps/react-native-maps/issues/4130)

### How did you test this PR?
The changes don't work yet, it's asking that the call should be done on the main thread, but I'm not experienced in those matters, so I'm opening the PR hoping someone can take a look.

Simulator Screen Shot - iPhone 12 Pro Max - 2022-07-18 at 11 01 30

### DIFF
--- a/ios/AirGoogleMaps/AIRGoogleMapManager.m
+++ b/ios/AirGoogleMaps/AIRGoogleMapManager.m
@@ -480,6 +480,11 @@ RCT_EXPORT_METHOD(setIndoorActiveLevelIndex:(nonnull NSNumber *)reactTag
   }];
  }
 
+RCT_EXPORT_METHOD(setMetalRendererEnabled:(BOOL)enabled)
+{
+  [GMSServices setMetalRendererEnabled:enabled];
+}
+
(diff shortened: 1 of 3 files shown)


Simon-TechForm:
hi @jbrodriguez 👋. I appreciate you taking the time to put this together.

Coincidentally I've been testing out the metal renderer for a while, although for performance reasons, not because of crashes. Either way, I'm happy that you provided the link to the google issue tracker, where Google officially recommends switching to the Metal renderer, as that makes me confident that we should make the change as well.

As enabling the metal renderer shouldn't be a breaking change, I believe that we should simply always enable it. I'd prefer not exporting more platform specific functions, and from a maintenance point of view, I'm dreading having to handle yet another case each time an issue is opened.

Would you be up for opening a new PR that simply enables the metal renderer?

jbrodriguez:
@Simon-TechForm it was easier than i had anticipated, opened https://github.com/react-native-maps/react-native-maps/pull/4336.

closing this one.

PR #4336, opened 2022-07-18
feat(ios-google): enable metal renderer

jbrodriguez:
No

This issue addresses https://github.com/react-native-maps/react-native-maps/issues/4130, based on discussions in https://github.com/react-native-maps/react-native-maps/pull/4334

### How did you test this PR?
This change affects iOS only.

As stated by Google, [they won't fix iOS crashes with the OpenGL renderer](https://issuetracker.google.com/issues/221643638#comment11).

Tests were performed adding the line on a production app and validating that the app didn't crash, setting `provider` to `PROVIDER_GOOGLE`


Simon-TechForm:
Excellent :) A couple of comments:

The change will become part of the release notes, so I don't believe that it's necessary to add a note to the installation docs.

Did you test whether the Google Maps SDK is happy that `[GMSServices setMetalRendererEnabled:YES]` is called multiple times? E.g. if a consumer is unmounting screens when navigating. It's most likely fine, but I would like to be sure :)

jbrodriguez:
Hi simon, I didn't test the setting the renderer multiple times, because we don't have that scenario in our app.

Having said that, if you suggest another place for that call, I can move it.

Maybe we can leave this config up to the user, that is call `[GMSServices setMetalRendererEnabled:YES]` in `didFinishLaunchingWithOptions` ? Haven't tried it :)

Simon-TechForm:
I'm fine keeping it where it is - the less user setup required, the better.

I will do some tests myself and then we should be good to merge this.

Simon-TechForm (approved):

PR #4347, opened 2022-07-22
feat(geojson): support for tracksViewChanges in Geojson component

jcuffney:

No.

- [x] allows for consumers to set `tracksViewChanges` for all makers via the `Geojson` component
- [x] allows for consumers to override the default `tracksViewChanges` on a per marker basis with a value set in their "point" at `properties.tracksViewChanges`

### How did you test this PR?

- This change affects both Android and iOS


Simon-TechForm:
Looks good :)

Can I have you add the prop to the [api docs](https://github.com/react-native-maps/react-native-maps/blob/master/docs/geojson.md) as well?

src/Geojson.tsx
```diff
@@ -198,11 +211,13 @@ const Geojson = (props: GeojsonProps) => {
     <React.Fragment>
       {pointOverlays.map((overlay, index) => {
         const markerColor = getColor(color, overlay, 'marker-color');
-
+        const pointOverlayTracksViewChanges =
+          overlay?.feature?.properties?.tracksViewChanges || tracksViewChanges;
```
Simon-TechForm, line 215: Optional chaining should only be needed on the `properties` key.

jcuffney:
> Looks good :)
> 
> Can I have you add the prop to the [api docs](https://github.com/react-native-maps/react-native-maps/blob/master/docs/geojson.md) as well?

yep!

docs/geojson.md
```diff
@@ -18,6 +18,7 @@
 | `onPress` | `Function`     |  | returns the selected overlay value with the onPress functionality |
 | `markerComponent` | `React Node`     |  | Component to render in place of the default marker when the overlay type is a `point` 
 | `title` | `string`     |  | The title of the marker. This is only used if the <Marker /> component has no children that  are a `<Callout />` 
+| `tracksViewChanges` | `Boolean` | | Sets whether this marker should track view changes. It's recommended to turn it off whenever it's possible to improve custom marker performance. This is the default value for all point markers in your geojson data. It can be overriden on a per point basis by adding a `trackViewChanges` property to the `properties` object on the point.
```
Simon-TechForm, line 21: Can we add the default value as well? `tracksViewChanges` defaults to `true`.

jcuffney: yep! - just added that in.

Simon-TechForm: Hmm, I don't see any changes? 🤔 Did you push your commit? :)

Simon-TechForm: I'll push a doc update after merging this, so we can move forward with the PR.

Simon-TechForm (approved):

PR #4417, opened 2022-08-23
feat: refactored podspec

pinpong:
Does any other open PR do the same thing?
No

What issue is this PR fixing?
Fixed auto link warning message

How did you test this PR?
Tested on IOS & Android devices with example project

PR #4446, opened 2022-09-17
fix(ios): add header paths required for use_frameworks compilation

mikehardy:

There is a similar PR doing a full reorganization of the iOS structure at #4417 , but maintainer @Simon-TechForm mentioned it was a breaking change, and indeed it is quite wide-ranging. So it is stalled.

Whereas this PR is laser-focused on a backwards-compatible / frameworks-or-not-frameworks fix for users that need `use_frameworks!` in their Podfile. I'm laser-focused on that myself because I maintain react-native-firebase and this is a pain point for my users.

The change itself is from @billnbell and I'm just carrying it into a PR with hope it is considered

In my experience most use_frameworks fixes are like this - just header path meddling and it works (that's the case for react-native core, and hermes so far as big examples, there are other smaller ones as well)

This PR fixes the case where `use_frameworks!` is in the Podfile, without breaking the non-use-frameworks case

### How did you test this PR?

I will defer to @billnbell on this one, he indicates he did a bunch of testing :-). Care to comment William?

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -9,9 +9,13 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    env:
+      HUSKY: 0
     steps:
       - name: Check out repository code
         uses: actions/checkout@v3
+        with:
+          persist-credentials: false
 
       - name: Setup node
         uses: actions/setup-node@v3
@@ -27,6 +31,6 @@ jobs:
 
       - name: Release
         env:
-          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          GITHUB_TOKEN: ${{secrets.CI_GITHUB_TOKEN}}
           NPM_TOKEN: ${{secrets.NPM_TOKEN}}
         run: yarn release
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -1,0 +1,15 @@
+{
+  "plugins": [
+    "@semantic-release/commit-analyzer",
+    "@semantic-release/release-notes-generator",
+    [
+      "@semantic-release/changelog",
+      {
+        "changelogTitle": "# Changelog"
+      }
+    ],
+    "@semantic-release/npm",
+    "@semantic-release/git",
+    "@semantic-release/github"
+  ]
+}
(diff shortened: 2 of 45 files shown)


billnbell (approved):
I tested this in 0.65.5, 0.70.0 and 0.70.1 releases with use_frameworks! enabled. Both worked in debug and release modes on iOS for react-native-maps v1.3.1. I also tested it with react-native-firebase v9.6.0 using use_frameworks. Hermes is enabled, and Flipper is off.
I also did not enable $RNFirebaseAsStaticFramework = true in the Podfile. I commented that out actually.

```
use_frameworks! :linkage => :static
```

I also added to the Podfile:

```
  rn_maps_path = '../node_modules/react-native-maps'
  pod 'react-native-google-maps', :path => rn_maps_path

  # Note you don't need to add RNBootSplash or react-native-app-auth if unused
  $static_library = [
   'React',
   'RNBootSplash',
   'Google-Maps-iOS-Utils',
   'GoogleMaps',
   'react-native-maps',
   'react-native-google-maps',
   'react-native-app-auth',
   'React-hermes'
  ]

pre_install do |installer|
      Pod::Installer::Xcode::TargetValidator.send(:define_method, :verify_no_static_framework_transitive_dependencies) {}
      installer.pod_targets.each do |pod|
        bt = pod.send(:build_type)
        if $static_library.include?(pod.name)
          puts "Overriding the build_type to static_library from static_framework for #{pod.name}"
          def pod.build_type;
            Pod::BuildType.static_library
          end
        end
      end
      installer.pod_targets.each do |pod|
        bt = pod.send(:build_type)
        puts "#{pod.name} (#{bt})"
        puts "  linkage: #{bt.send(:linkage)} packaging: #{bt.send(:packaging)}"
      end
  end
```

Make sure you `pod deintegrate`, removed Podfile.lock to clear it all. I also `rm -rf ~/Library/Developer/Xcode/DerivedData` to make sure it is cleared. Then pod install.

Seems very limited scope since we are just added a few Header search paths so it can find the .h headers when you use use_frameworks!.

I thought this would be helpful for others struggling to get it to work.

Simon-TechForm:
Seems harmless to me, so I'm happy to merge this :)

Simon-TechForm (approved):

MarcoBrugali255:
Hi, i don't understand, with the 1.3.2-beta.1 I need to edit podfile like @billnbell solution?

With this solution, maps work, but debugger and fast refresh not work

NateJackman:
> billnbell

took me a while to find this solution! thanks

Issue #4110, opened 2022-03-08
[IOS] onRegionChangeComplete is not firing up at all


aliwaqar981:
# Bug report

## Summary

onRegionChangeComplete is not firing up at any point on ios.

## Environment info

      react-native: 0.66.4
      react: 17.0.2
      react-native-maps: ^0.30.1

    Target Platform: iOS - 14.5 Simulator/iPhone 12,

    MapProvider: AppleMaps

`react-native info` output:

System:
    OS: macOS 11.2.3
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 271.91 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - /usr/local/bin/node
    Yarn: 1.22.17 - ~/.yarn/bin/yarn
    npm: 8.3.1 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.10.2 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    Android SDK:
      API Levels: 23, 24, 25, 26, 28, 29, 30
      Build Tools: 23.0.1, 26.0.1, 28.0.3, 29.0.2, 29.0.3, 30.0.2
      System Images: android-28 | Intel x86 Atom_64
      Android NDK: Not Found
  IDEs:
    Android Studio: 4.1 AI-201.8743.12.41.6953283
    Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_252 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2 
    react-native: 0.66.4 => 0.66.4 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

## Steps to reproduce

1. Set up new react native project
2. Install react-native-maps
3. Follow all instructions given in installation.md file
4. Try to fire onRegionChangeComplete on ios

Describe what you expected to happen:

1. When scroll on map ends then onRegionChangeComplete should fire

Describe what you actually happens:

1. After I stop scrolling map onRegionChangeComplete is not firing as it's getting fired on android.

## Reproducible sample code

```
[...]

PR #4478, opened 2022-10-11
fix: ios onRegionChange stops firing up when animateCamera is called …

kappu72:
No

[IOS] onRegionChangeComplete is not firing up at all
(please link the issue here) #4110

### How did you test this PR?
On IOS, calling animate camera repeatedly, it never stops firing the event as before
Create a map, crete a flow of animateCamera called repeatedly, and print the result of onRegionChangeComplete
tested on simulator iPhone 13 iOs 15.4 and real device iPhone Xr iOs 15.5

### DIFF
--- a/ios/AirMaps/AIRMap.h
+++ b/ios/AirMaps/AIRMap.h
@@ -51,7 +51,7 @@ extern const NSInteger AIRMapMaxZoomLevel;
 @property (nonatomic, assign) MKCoordinateSpan pendingSpan;
 
 
-@property (nonatomic, assign) BOOL ignoreRegionChanges;
+@property (nonatomic, assign) NSInteger ignoreRegionChanges;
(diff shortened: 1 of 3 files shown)


WillyRamirez:
Can a maintainer please look at this?

bawb89:
Hi @Simon-TechForm and @christopherdro, who is reviewing PR's on this lib currently? Could perhaps one of you be so kind as to review this one? We have a lot riding on this bug getting solved. Thanks!!

WillyRamirez:
@kappu72 Thanks again for the PR. I forked the branch from your PR and I receive the following error in my app trying when opening the screen that contains the map:

`View config getter callback for component AIRGoogleMap must be a function (received undefined)`

Is this something you're familiar with?

Issue #4458, opened 2022-09-22
Apple Maps URLTile shouldReplaceMapContent not respected iOS 16


stdavis:
### Summary

I'm using a custom `URLTile` provider that has labels baked into it. Everything looks good in the expo dev client. But after I built it, the apple maps labels show up.

expo dev client:
<img width="414" alt="image" src="https://user-images.githubusercontent.com/1326248/191782298-22e574cd-b49f-41ea-829a-fe9da42972ff.png">

built version:
<img width="414" alt="image" src="https://user-images.githubusercontent.com/1326248/191782573-7eada1d3-0221-4605-83cd-2077c2206c9b.png">

### Reproducible sample code

```TSX
import { Platform } from 'react-native';
import MapView, { MAP_TYPES, UrlTile } from 'react-native-maps';

export default function Map() {
  return (
    <MapView
      initialRegion={{
        latitude: 38.38576,
        longitude: -112.660201,
        latitudeDelta: 0.1,
        longitudeDelta: 0.05,
      }}
      loadingEnabled={true}
      mapType={Platform.select({ android: 'none', ios: MAP_TYPES.STANDARD })}
      maxAltitude={650}
      maxZoomLevel={18}
      minZoomLevel={5}
      pitchEnabled={false}
      rotateEnabled={false}
      showsBuildings={false}
      showsCompass={false}
      showsMyLocationButton={false}
      showsPointsOfInterest={false}
      showsScale={false}
      toolbarEnabled={false}
    >
      <UrlTile
        doubleTileSize={true}
        minimumZ={3}
        shouldReplaceMapContent={true}
        tileSize={384}
        urlTemplate='https://discover.agrc.utah.gov/login/path/april-select-minute-alpha/tiles/hybrid_basemap/{z}/{x}/{y}.jpg'
        zIndex={-5}
      />
    </MapView>
  );
}
```

### Steps to reproduce

Open this map in an expo dev client and notice that there are no labels.

Run a build and notice that the labels appear.

Please note that this is a temporary base map URL that will expire in 4 weeks since this is licensed imagery.

### Expected result

I expected the dev version (no labels) to match the prod version.

### Actual result

The prod version shows the apple maps labels.

### React Native Maps Version

1.3.1

### What platforms are you seeing the problem on?

iOS (Apple Maps)

### React Native Version

0.69.5

### What version of Expo are you using?

SDK 46

### Device(s)

iPhone 11 (iOS 16.0)

### Additional information

[...]

PR #4549, opened 2022-12-04
fix for URLTile shouldReplaceMapContent

cixio:

no

https://github.com/react-native-maps/react-native-maps/issues/4458

### How did you test this PR?

iPhone Simulator + iPhone 14 Pro

```
      <MapView
        style={{width:"100%", height:"100%"}}
        showsScale={true}
        initialCamera={{
          center: {
            latitude: 52,
            longitude: 10,
          },
          heading: 5,
          altitude: 26900,
        }}
        provider={PROVIDER_DEFAULT}
        mapType={"standard"}
      >
        <UrlTile
          urlTemplate={"https://b.tile.openstreetmap.de/{z}/{x}/{y}.png"}
          maximumZ={19}
          shouldReplaceMapContent={true} 
        />  
      </MapView>
```

Result: Apple Maps are not longer in Background of the map and labels are not longer visible above the UrlTile-Map.


stdavis:
This solves #4458 for me. Thanks, @cixio! 🙌

Simon-TechForm:
@cixio thank your for putting this together and thank you @stdavis trying it out and providing feedback 👍

@stdavis is iOS 15 and iOS 16 behaving identically when trying out this change? I'm a little confused as to why this started acting up in iOS 16 without any release notes from Apple 😕 (`MKOverlayLevelAboveLabels` was added in iOS 7 after all...)

I'm considering whether we would need to add this as an option instead of forcing it on every consumer. While using `MKOverlayLevelAboveLabels` might be desired if using tiles that includes labels and POI's, some consumers might rely on these actually being shown all the time.

stdavis:
> is iOS 15 and iOS 16 behaving identically when trying out this change?

Yes. It broke at 16 for me.

> I'm considering whether we would need to add this as an option instead of forcing it on every consumer.

I think that you may be on to something. My tiles have labels baked in, but I can see users who are using plain imagery wanting the labels to show up on top.

Simon-TechForm:
> I think that you may be on to something. My tiles have labels baked in, but I can see users who are using plain imagery wanting the labels to show up on top.

Exactly. I don't think this can be merged as is.

If either of you are able to work on a solution that provides this setting as a prop, I'll be happy to provide feedback and review :)

stdavis:
> If either of you are able to work on a solution that provides this setting as a prop, I'll be happy to provide feedback and review

I took a stab at it: https://github.com/react-native-maps/react-native-maps/pull/4725

cixio:
Still relevant in combination with https://github.com/react-native-maps/react-native-maps/pull/4725

salah-ghanim:
@cixio thank you for this pull request, I'm trying to merge it but I see somethings need to be addressed first:

1. I suggested to @stdavis an alternative implementation and waiting on his feedback, so any changes regarding the tiles / layers order need to wait on that discussion because him and @monholm highlighted a good point regarding other users of the library.

2. I see that you fixed a crash for removing a layer on android, can you please submit a pull request specifically for that issue, that way it can be quickly merged regardless of this discussion.

3. I see that you added a note regarding an issue of auto-tracking user, but I don't see the change for it in in the code.

PR #4553, opened 2022-12-07
ios fix

infoSamyak:

Not sure about other PRs

iOS build fails as the import paths are relative in AirGoogleMap.m instead of absolute so fixed that issue 

### How did you test this PR?

This takes effect on iOS. Tested on both real device and simulator.


.gitignore
```diff
@@ -44,5 +44,3 @@ yarn-error.log
 # VSCode workspace settings
 .vscode
 
-# ts outDir
-lib/
```
WillyRamirez, line 48: I think lib should not live in the repository, it gets added in the workflow through `yarn build` before deploy to the NPM registry

infoSamyak: Yes
You are right about that but until the PR is merged I am using these changes as a fork. Now since it's a fork the yarn build does not work and therefore I had to push the lib folder.
Can you provide any suggestions on how to proceed with this?

FrikkieSnyman: Use a different branch containing these changes instead of using your PR?

econnerty:
I tried installing this fork and got `AppDelegate.mm:7:9: fatal error: 'GoogleMaps/GoogleMaps.h' file not found
#import <GoogleMaps/GoogleMaps.h>`

Any ideas? It seemed to get further.

Jessen94:
> I tried installing this fork and got `AIRGoogleMapPolygon.m normal arm64 objective-c com.apple.compilers.llvm.clang.1_0.compiler (in target 'react-native-google-maps' from project 'Pods') (1 failure)`
> 
> Any ideas? It seemed to get further.

There is a bug on that issue reported here
https://github.com/react-native-maps/react-native-maps/issues/4563

PR #4574, opened 2022-12-28
onRegionChangeComplete() callback is called infinitely

Bilal-Abdeen:
https://github.com/react-native-maps/react-native-maps/issues/846#issuecomment-1210079461

(please answer here)
No

(please link the issue here)
https://github.com/react-native-maps/react-native-maps/issues/846

### How did you test this PR?

(please answer here)
PR is for adding more info on the readme file.


RyanTG:
Is it worth noting that this fix only works if you're using Google Maps? `isGesture` isn't available on Apple Maps.

Bilal-Abdeen:
> Is it worth noting that this fix only works if you're using Google Maps? `isGesture` isn't available on Apple Maps.

You are right. I added this note. Thanks.

MadeinFrance (dismissed):

Issue #4552, opened 2022-12-07
RCTConvert+AirMap.h module not found react native maps


rupam999:
### Summary

**I was getting the below error**
RCTConvert+AirMap.h module not found react native maps

So, I added the below lines in my pod file

```
$static_framework += [
    'react-native-maps',
    'react-native-google-maps',
    'Google-Maps-iOS-Utils',
    'GoogleMaps',
    'RNPermissions',
    'Permission-LocationWhenInUse',
    'Permission-Notifications',
    'react-native-paste-input',
    'vision-camera-code-scanner',
    'VisionCamera'
  ]

pre_install do |installer|
    Pod::Installer::Xcode::TargetValidator.send(:define_method, :verify_no_static_framework_transitive_dependencies) {}
        installer.pod_targets.each do |pod|
            if $static_framework.include?(pod.name)
                def pod.build_type;
                Pod::BuildType.static_library # >= 1.9
            end
        end
    end
  end
```

**After that I am getting the below error**

Any Idea how solve this issue

error Failed to launch the app on simulator, An error was encountered processing the command (domain=NSPOSIXErrorDomain, code=3):
Application launch for 'org.reactjs.native.example.BiroCabs' did not return a valid pid nor a launch error.
No such process

### Reproducible sample code

```TSX
require_relative '../node_modules/react-native/scripts/react_native_pods'
require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'

$RNFirebaseAsStaticFramework = true

platform :ios, '13.0'
install! 'cocoapods', :deterministic_uuids => false
use_frameworks! :linkage => :static

target 'App' do
  $static_framework = []
  config = use_native_modules!

  # Flags change depending on the env values.
  flags = get_default_flags()

  # React Native Maps dependencies
  rn_maps_path = '../node_modules/react-native-maps'
  pod 'react-native-google-maps', :path => rn_maps_path

  $static_framework += [
    'react-native-maps',
    'react-native-google-maps',
    'Google-Maps-iOS-Utils',
    'GoogleMaps',
    'RNPermissions',
    'Permission-LocationWhenInUse',
    'Permission-Notifications',
    'react-native-paste-input',
    'vision-camera-code-scanner',
    'VisionCamera'
  ]

  use_react_native!(
    :path => config[:reactNativePath],
    # Hermes is now enabled by default. Disable by setting this flag to false.
    # Upcoming versions of React Native may rely on get_default_flags(), but
[...]

Issue #4492, opened 2022-10-25
[Regression] 'RCTConvert+AirMap.h' file not found


BrodaNoel:
### Summary

This issue is quite similar to https://github.com/react-native-maps/react-native-maps/issues/3597

Happening now with Expo Dev Clients

I've updated some libraries in `Expo`. One of those is `expo-firebase-analytics`.
The new version of `expo-firebase-analytics` (https://docs.expo.dev/versions/latest/sdk/firebase-analytics/) implies adding this config:

```
{
  "expo": {
    "plugins": [
      [
        "expo-build-properties",
        {
          "ios": {
            "useFrameworks": "static"
          }
        }
      ]
    ]
  }
}
```

As you can see in the last comments in https://github.com/react-native-maps/react-native-maps/issues/3597, some other people are having the same issue after adding that ` "useFrameworks": "static" ` config.

### Reproducible sample code
It is actually a matter of dependencies.

Add these dependencies:

```
"expo-build-properties": "~0.3.0",
"expo-firebase-analytics": "~7.2.0",
"react-native-maps": "0.31.1",
```

RNMaps is on `0.31.1` because it's the last one supported by the current Expo version.

### Steps to reproduce

Just build `eas build -p ios`.

### Expected result

Build go green

### Actual result

Build red.

```
› Compiling react-native-maps Pods/react-native-google-maps » AIRGoogleMapWMSTileManager.m
› Compiling react-native-maps Pods/react-native-google-maps » AIRGoogleMapWMSTile.m
› Compiling react-native-maps Pods/react-native-google-maps » AIRGoogleMapUrlTile.m
› Compiling react-native-maps Pods/react-native-google-maps » AIRGoogleMapURLTileManager.m
› Compiling react-native-maps Pods/react-native-google-maps » AIRGoogleMapPolylineManager.m

❌  (node_modules/react-native-maps/ios/AirGoogleMaps/AIRGoogleMapPolylineManager.m:17:9)

  15 | #import <React/RCTViewManager.h>
  16 | #import <React/UIView+React.h>
> 17 | #import "RCTConvert+AirMap.h"
     |         ^ 'RCTConvert+AirMap.h' file not found
  18 | #import "AIRGoogleMapPolyline.h"
  19 | 
  20 | @interface AIRGoogleMapPolylineManager()
› Compiling react-native-maps Pods/react-native-google-maps » AIRGoogleMapPolyline.m

[...]

Issue #3597, opened 2020-10-29
'RCTConvert+AirMap.h' file not found


WASHSWATDAVID:
# Bug report

## Environment info

`react-native info` output:

```bash
System:
    OS: macOS 10.15.6
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Memory: 50.64 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 8.8.1 - ~/.nvm/versions/node/v8.8.1/bin/node
    Yarn: 1.12.1 - /usr/local/bin/yarn
    npm: 5.4.2 - ~/.nvm/versions/node/v8.8.1/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.10.0 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.1, DriverKit 19.0, macOS 10.15, tvOS 14.0, watchOS 7.0
    Android SDK: Not Found
  IDEs:
    Android Studio: 3.1 AI-173.4907809
    Xcode: 12.1/12A7403 - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_181 - /usr/bin/javac
    Python: 2.7.16 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: ^4.13.0 => 4.13.0 
    react: 16.11.0 => 16.11.0 
    react-native: 0.62.2 => 0.62.2 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```
My Xcode version 12.1 (12A7403)
"react-native-maps": "0.27.1",

Podfile
```
  rn_maps_path = '../node_modules/react-native-maps'
  pod 'react-native-google-maps', :path => rn_maps_path
  pod 'GoogleMaps'
  pod 'Google-Maps-iOS-Utils'
```

## Steps to reproduce

When I build my ios project always shown` 'RCTConvert+AirMap.h' file not found` message

<img width="1225" alt="Screen Shot 2020-10-30 at 4 25 46 AM" src="https://user-images.githubusercontent.com/39440825/97623017-68141680-1a68-11eb-8241-88e4a1ace8ba.png">

JavyMB:
Hi @WASHSWATDAVID 
[...]

PR #4578, opened 2022-12-31
added additional information about react-native-google-maps

Bilal-Abdeen:

No 

(please link the issue here)
#4552, #4492, #3597 
### How did you test this PR?

After adding the suggested statements, the error was not thrown anymore.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -40,6 +40,8 @@ If you want to enable Google Maps on iOS, obtain the Google API key and edit you
 
 (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+// The plus sign '+' below is part of the statement. Do NOT remove it. 
+// Replace _YOUR_API_KEY_ below with your projcet's Google API Key
 +  [GMSServices provideAPIKey:@"_YOUR_API_KEY_"]; // add this line using the api key obtained from Google Console
 ...
 ```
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -40,6 +40,8 @@ If you want to enable Google Maps on iOS, obtain the Google API key and edit you
 
 (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+// The plus sign '+' below is part of the statement. Do NOT remove it. 
+// Replace _YOUR_API_KEY_ below with your projcet's Google API Key
 +  [GMSServices provideAPIKey:@"_YOUR_API_KEY_"]; // add this line using the api key obtained from Google Console
```
Simon-TechForm, line 45: This is a standard diff view. It shouldn't be changed :)

docs/installation.md
```diff
@@ -52,17 +54,33 @@ Also make sure that your Podfile deployment target is set to >= 13.0 at the top
 
 ```ruby
 platform :ios, '13.0'
+...
+# if you get erros because of missing react-native-google-maps files e.g. RCTConvert+AirMap.h, consider adding the following 2 statements 
+pod 'GoogleMaps'
+pod 'Google-Maps-iOS-Utils'
```
Simon-TechForm, line 60: This shouldn't be added, as it is part of the podspec included in react-native-maps.

docs/installation.md
```diff
@@ -52,17 +54,33 @@ Also make sure that your Podfile deployment target is set to >= 13.0 at the top
...
+post_install do |installer|
+  ...
+  # if you get erros because of missing react-native-google-maps files e.g. RCTConvert+AirMap.h, consider adding the following section
+  installer.pods_project.targets.each do |target|
+    if target.name == 'react-native-google-maps'
+        target.build_configurations.each do |config|
+          config.build_settings['CLANG_ENABLE_MODULES'] = 'No'
+        end
+    end
+  end
```
Simon-TechForm, line 80: This seems like a fix for an issue that wasn't discussed. Feel free to open an issue regarding what is fixed here :)

docs/installation.md
Bilal-Abdeen: What I am trying to say is that in this case, the "+" sign is NOT part of the standard diff view. It is actually part of the code, and it should be typed as it is!

docs/installation.md
Bilal-Abdeen: In my case, I think it helped fix the issue. If you don't agree, you can decline it.

docs/installation.md
Bilal-Abdeen: I think it fixes the following issues. If you don't agree, you can decline it. 
#4552, #4492, #3597

docs/installation.md
Simon-TechForm: We are not trying to get the consumer to type the `+` as it is. While it might build while leaving the `+` in, the code should look like this:
```objc
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
  [GMSServices provideAPIKey:@"some-api-key"];
  self.moduleName = @"MapsTest";
  // You can add your custom initial props in the dictionary below.
  // They will be passed down to the ViewController used by React Native.
  self.initialProps = @{};

  return [super application:application didFinishLaunchingWithOptions:launchOptions];
}
```

not like this:
```objc
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
  + [GMSServices provideAPIKey:@"some-api-key"];
  self.moduleName = @"MapsTest";
  // You can add your custom initial props in the dictionary below.
  // They will be passed down to the ViewController used by React Native.
  self.initialProps = @{};

  return [super application:application didFinishLaunchingWithOptions:launchOptions];
}
```

Any reason why you would want to leave the `+` in there?

Bilal-Abdeen: You are right. Sorry.

docs/installation.md
BrodaNoel: Any news here?

Simon-TechForm: I have no interest in "declining it", but we need to have a more thorough discussion regarding why this fixes the issue.

If this fixes any issue, please open a new one, describe how and why this fixes the issue. Keep them coming! 👍

PR #4678, opened 2023-04-18
fix(ios): support for use_frameworks! :linkage => :static

Simon-TechForm:
react-native-google-maps relies on headers from react-native-maps. When using frameworks, headers from react-native-maps are no longer found, as the framework search path for react-native-maps isn't added to the react-native-google-maps framework.

Likewise, removing the import prefix from Google-Maps-iOS-Utils resolves the imports both with and without use_frameworks.

Note: this only brings support for `use_frameworks!` with `:linkage => :static.


MadeinFrance (approved):
Good job @Simon-TechForm ! Everything builds and run 🚀

Small note: 

I've tested the Example and I noticed an Error with the Animated Marker Position example: `Error: This JS animated node type cannot be used as native animated node, js engine: hermes`

Simon-TechForm:
> 

I appreciate you testing things out 👍 That error seems unrelated to this pr, do you agree? Lets start a new issue/discussion if/when we need to address it.

PR #4725, opened 2023-05-01
Add "showLabels" prop to URLTile and WMSTile components

stdavis:

This PR is intended to supersede #4549.

#4458 

### How did you test this PR?

Tested via iOS simulator (v16.*) using `example/src/examples/CustomTiles.tsx` toggling the `showLabels` prop.


Simon-TechForm:
Hi @stdavis. Thank you for putting this together and paying attention to the context in the files you are contributing to 👍 It makes me happy when alphabetical order, jsdoc comments etc. are included from the beginning.

I've left a few comments which should also fix the build error in the CI. Could you include the prop in the tiles doc file as well? :)

stdavis:
I just rebased and pushed a new commit adding the new prop to the tiles doc.

I don't see any comments from you. Perhaps you are still putting them together?

Simon-TechForm (changes requested):

Simon-TechForm:
> I just rebased and pushed a new commit adding the new prop to the tiles doc.
> 
> I don't see any comments from you. Perhaps you are still putting them together?

It might help if I remember submitting my review comments 😴

cixio:
> So I'd like to understand under which circumstances the issue occur, and whether we need to supply the setting as a prop, since it's already applied in the update function.

Open app with Apple Maps -> switch to Tile Map -> Apple Labels were shown -> switch to Other Tile Map -> Apple Labels gone

Open app with Tile Map visible at startup -> No Labels -> switch to Other Tile Map -> No Labels -> switch to Apple Maps -> Labels (thats ok!) -> switch back to Tile Map -> Labels (not ok!) -> Other Tile map -> No Labels

This means that the update function (switch tile map) hides the apple labels if they were shown before ( = if apple map was active some time before).

If you add "level:MKOverlayLevelAboveLabels" to the "init" function the labels were not shown (as it should be i think).

But warning: the "level:MKOverlayLevelAboveLabels" puts the map also above polylines (i dont know why, markers are every time on top of everything). So my solution is to add "level:MKOverlayLevelAboveLabels" also to polylines (in both, init and update) - that fixes all the problems, but makes a new bug: after changing from one tile map to another tile map you have to give the polylines an update (remove and show again) to be shown on top of everything. As far as I have found out, there is no further level at Apple such as e.g. above markers.

Sorry, much complex and english is not my native language... tell me if you dont understand something.

stdavis:
@monholm Is there anything that I can do to help get this pushed through?

stdavis:
Just realized that this makes it so that Polylines (and likely others like Polygons) are hidden under the tile layers. I'm not sure that I like this solution...

cixio:
I'm sorry, but that's a problem. It's important that the lines are displayed only after the map is switched. Alternatively, there's a small workaround:

```
const [ignored, forceUpdate] = React.useReducer(x => x + 1, 0);

// my polyline in the <MapView>
{(coordinates.length > 1) &&
        <Polyline 
          coordinates={coordinates}
          strokeColor="#ff950088"
          strokeWidth={4}
          zIndex={10}
        />
      }

// script to update the polyline after the map tile changes
var tmp_coordinates = coordinates;
coordinates = [];
setTimeout(function() { coordinates = tmp_coordinates; forceUpdate(); }, 100);
```

salah-ghanim:
@stdavis thank you so much for being persistent here trying to resolve this issue.

I'm wondering though about this approach since showLabels seems very cryptic and add a hardcoded dependency between whatever layer you're adding and the specific MKOverlayLevelAboveLabels.


have you considered adding a new iOS specific optional property calling it (aboveLayer) and if the value of the property is set you pass it a long to             [self addOverlay:(id<MKOverlay>)subview level:(providedvalue)];


in such a generic case it can solve any future issue regarding layers order, and provide better API parity with native.


looking forward for your feedback :)

stdavis:
@salah-ghanim I think that your suggestion is a much better solution. The only feedback that I have is that maybe the new prop name should be something like `level` to match [the Apple API docs](https://developer.apple.com/documentation/mapkit/mkmapview/1452635-addoverlay#:~:text=the%20MKOverlay%20protocol.-,level,-The%20map%20level). Or maybe `layerLevel`? `aboveLayer` made me think that I needed to pass some sort of layer object at first.

We've recently switched to native base maps in our app so this is not a huge priority for our project. However, this library has been great for us and I'd be happy to help get this PR updated to reflect your suggestion if you feel good about it. I'd also love to get @cixio's thought about it.

cixio:
As we are using this only for URLTile and WMSTile a _providedvalue_ does not make any sense for me because there are only 2 valid levels: aboveRoads and aboveLabels - see apple docs: https://developer.apple.com/documentation/mapkit/mkoverlaylevel as I can remember only aboveLabels is above a tile.

What has not yet been taken into account in this pull request are the update functions, e.g. https://github.com/react-native-maps/react-native-maps/blob/edfac8ae89f6f0e80e5c93604df02124b3b844ae/ios/AirMaps/AIRMapLocalTile.m#L48C21-L48C21

Also the polyline-functions had to be adjusted, because of creating a new "bug" like written here: https://github.com/react-native-maps/react-native-maps/pull/4725#issuecomment-1605755193

I've put a lot of time into this bug, but like @stdavis this is also no longer relevant for my project. I am sure that I have described all the necessary information in detail here and in my pull request there: https://github.com/react-native-maps/react-native-maps/pull/4549

salah-ghanim:
@stdavis 
1. regarding your feedback, yes layerLevel or aboveLevel work well.. I didn't put much thought into the naming.
2. regarding int issue with iOS,  I'm not sure exactly why it's a problem, you can define a set of values (constants as type) in typescript, any other value sent by the user is going against the docs and it's fine by me if the behaviour is unexpected... in the future whenever apple adds similar layer to their definitions we can simply define it by it's constant value.

jrmgx:
Hello, I had the same bug https://github.com/react-native-maps/react-native-maps/issues/4458 and from what I find, this PR would fix it. Any reason it has been abandoned?

PR #4752, opened 2023-06-02
Fix table in geojson.md

andokai:
The markdown for the table is broken. This fixes it.

No

It's fixing broken documentation. There is no issue reporting it.

### How did you test this PR?

I previewed the document after the fix and the table renders correctly.


noxan:
Stumbled across the same problem, the fix looks good, would be great to have it merged 👍

jaydensim (approved):
This change looks good and does indeed fix the issue with the table.

PR #4766, opened 2023-06-14
feat(UrlTile): allow to use google map opacity on IOS

rcadelw33:

There is no other PR to add this feature

Currently it's not possible to pass opacity to  UrlTile component when using Google Map on IOS

### How did you test this PR?

I fork this repo and use npm local dependency to test that opacity is modified wheni change the props from React component


rcadelw33:
Still relevant

Issue #4776, opened 2023-06-25
MapVIew/Marker/Callout onPress delay


mikescott:
### Summary

Reopening this as issues #833 #3429 and #3897 have been closed but the problem still exists

### Reproducible sample code

```TSX
<Callout onPress={() => console.log('press')} />
```

### Steps to reproduce

Using Google as the provider

Press the callout

### Expected result

Instant response

### Actual result

Delay of around 500 ms

### React Native Maps Version

1.3.2

### What platforms are you seeing the problem on?

Android, iOS (Google Maps)

### React Native Version

0.71.8

### What version of Expo are you using?

SDK 48

### Device(s)

All devices

### Additional information

This does not affect Apple Maps on iOS. A previous proposed solution of wrapping the MapView in Pressable does not have any effect.

MikeKovarik:
Still relevant

salah-ghanim:
@mikescott @mikescott there is a pull request to solve this issue, but the solution is not optimal, I'm waiting on a better pull request to solve this.

mikescott:
Still relevant

jordiup:
Still relevant

ABertaud:
Still relevant

ABertaud:
Still relevant

kris-devjet:
Still relevant

WeidsonWikaro:
Still relevant

Alssdev:
Still relevant

cubevis:
Still exists

gustavohariel:
Still relevant

JoachimAmoah:
still relevant

PR #4782, opened 2023-07-03
Improving performance for onMarkerClick event (Android only)

MartinaMuscarella:
No

(please answer here)

Some issue that could be get improvements form this is #4776 

### How did you test this PR?

We have an app that needs to load several thousands of marker. On Android devices the onPress event is very slow to be triggered. We discovered that it was caused by the iteration on the markers. We used the setTag(Object)/getTag() to make it faster and avoid the search on markers.

### DIFF
--- a/android/src/main/java/com/rnmaps/maps/MapMarker.java
+++ b/android/src/main/java/com/rnmaps/maps/MapMarker.java
@@ -430,6 +430,7 @@ public class MapMarker extends MapFeature {
   public void addToMap(Object collection) {
     MarkerManager.Collection markerCollection = (MarkerManager.Collection) collection;
     marker = markerCollection.addMarker(getMarkerOptions());
+    marker.setTag(this);
     updateTracksViewChanges();
   }
 
--- a/android/src/main/java/com/rnmaps/maps/MapView.java
+++ b/android/src/main/java/com/rnmaps/maps/MapView.java
@@ -1375,7 +1375,13 @@ public class MapView extends com.google.android.gms.maps.MapView implements Goog
   }
 
   private MapMarker getMarkerMap(Marker marker) {
-    MapMarker airMarker = markerMap.get(marker);
+    MapMarker airMarker = (MapMarker) marker.getTag();
+
+    if (airMarker != null) {
+      return airMarker;
+    }
+
+    airMarker = markerMap.get(marker);
 
     if (airMarker != null) {
       return airMarker;


MartinaMuscarella:
Still relevant

android/src/main/java/com/rnmaps/maps/MapMarker.java
```diff
@@ -430,6 +430,7 @@ public Object getFeature() {
   public void addToMap(Object collection) {
     MarkerManager.Collection markerCollection = (MarkerManager.Collection) collection;
     marker = markerCollection.addMarker(getMarkerOptions());
+    marker.setTag(this);
```
salah-ghanim, line 433: @MartinaMuscarella this code is way too heavy compared to the functionality we're trying to achieve here 

specifically, you're storing a full view collection as an "attachment" to each Marker, and according to google docs:
https://developers.google.com/android/reference/com/google/android/gms/maps/model/Marker#setTag(java.lang.Object)
there is no automated mechanism by google to free memory here

if you look at setOnMarkerClickListener inside MapView.java you can see that only the markerIdentifier is needed.

and if you look at almost all the calls for getMarkerMap in MapView.java you find out only the viewID is needed to dispatch the event through MapManager.receiveEvent

android/src/main/java/com/rnmaps/maps/MapView.java
```diff
@@ -1375,7 +1375,13 @@ public void setIndoorActiveLevelIndex(int activeLevelIndex) {
   }
 
   private MapMarker getMarkerMap(Marker marker) {
-    MapMarker airMarker = markerMap.get(marker);
+    MapMarker airMarker = (MapMarker) marker.getTag();
```
salah-ghanim, line 1378: @MartinaMuscarella I would suggest creating a lightweight class in Android and call it MarkerTag
in it you can simply add two properties: viewId, markerId

then you can attach it to the marker as a tag

and where ever in the MapView getMarkerMap is called, replace it with a call to get the tagFirst as you did, and have a new manager.pushEvent function that takes the viewID directly, not just the viewObject

salah-ghanim (changes requested):
Using Marker.setTag is extremely memory efficient.

let's avoid using views to keep track of IDs, let's keep track of IDs directly instead (markerId, viewId)

cctv1005s:
@MartinaMuscarella  Hi, could you please take a look at the comments? This PR is very important for Android users.

WeidsonWikaro:
Still relevant

Issue #4786, opened 2023-07-07
`subThoroughfare` is missing on Address type


huextrat:
### Summary

The following type does not offer subThoroughfare, even though this is returned on Android and iOS.

```ts
export type Address = {
  administrativeArea: string;
  country: string;
  countryCode: string;
  locality: string;
  name: string;
  postalCode: string;
  subAdministrativeArea: string;
  subLocality: string;
  thoroughfare: string;
};
```

### Reproducible sample code

```TSX
Not needed
```

### Steps to reproduce

Not needed

### Expected result

We should have the following type:

```ts
export type Address = {
  administrativeArea: string;
  country: string;
  countryCode: string;
  locality: string;
  name: string;
  postalCode: string;
  subAdministrativeArea: string;
  subLocality: string;
  thoroughfare: string;
  subThoroughfare: string;
};
```

### Actual result

Currently `subThoroughfare` is missing:

```ts
export type Address = {
  administrativeArea: string;
  country: string;
  countryCode: string;
  locality: string;
  name: string;
  postalCode: string;
  subAdministrativeArea: string;
  subLocality: string;
  thoroughfare: string;
};
```

### React Native Maps Version

All

### What platforms are you seeing the problem on?

Android, iOS (Apple Maps), iOS (Google Maps)

### React Native Version

0.72.1

### What version of Expo are you using?

Not using Expo

### Device(s)

All

salah-ghanim:
@huextrat see: https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_address
and https://developer.apple.com/documentation/corelocation/clplacemark

for me it's clear that this is:
1. available only for iOS MapKit and not google maps and not android.
2. optional for iOS to begin with.

I think such optional / platform specific properties should:
1. be optional (?) type.
2. ideally verbously specified as platform specific, for example mkSubThoroughfare

huextrat:
Hi @salah-ghanim !

Android is using [Geocoder class](https://developer.android.com/reference/android/location/Geocoder) which is based on `android.location` and it contains a `subThoroughfare`. 

https://github.com/react-native-maps/react-native-maps/blob/673e9d21212442af9795cc21a17bf477623c458b/android/src/main/java/com/rnmaps/maps/MapModule.java#L190-L242

I've just tested it and can confirm that we do indeed have `subThoroughfare` on Android.

[...]

PR #4787, opened 2023-07-07
fix: add missing `subThoroughfare` to Address type

huextrat:

NO

https://github.com/react-native-maps/react-native-maps/issues/4786

### How did you test this PR?

`subThoroughfare` Should be available on Android & iOS for Google Maps and Apple Maps

### DIFF
--- a/src/MapView.types.ts
+++ b/src/MapView.types.ts
@@ -181,6 +181,7 @@ export type Address = {
   subAdministrativeArea: string;
   subLocality: string;
   thoroughfare: string;
+  subThoroughfare: string;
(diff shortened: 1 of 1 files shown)


salah-ghanim:
@huextrat this doesn't seem like an accurate fix, specifically the subThoroughfare seems to be only relevent to apple maps, I can't find it in google maps docs.. I will reply in the relevent issue to have a discussion about this.

huextrat:
Still relevant

salah-ghanim (approved):
LGTM

Issue #4536, opened 2022-11-25
Flurry of setNativeProps deprecation warnings when using markers


sharvilak11:
### Summary

We are updating plotted markers for drivers frequently and it gives a flurry of warnings stating - setNativeProps is deprecated and will be removed in next major release. This ultimately bloats the console outputs.

`console.warn('setNativeProps is deprecated and will be removed in next major release');`

Please advise how to remove this warning.

### Reproducible sample code

```TSX
import {Animated, Easing, Image, Platform} from 'react-native';
import {useEffect, useRef, useState} from 'react';
import {AnimatedRegion, Marker} from 'react-native-maps';
import driverMarkerImg from '../../../../../assets/driver-marker.png';

const AnimatedImage = Animated.createAnimatedComponent(Image);

function DriverMarker({location, driver, ...rest}) {
  const [marker, setMarker] = useState(null);
  const [coordinate] = useState(
    new AnimatedRegion({
      latitude: location.Latitude,
      longitude: location.Longitude,
      latitudeDelta: 0.025,
      longitudeDelta: 0.025,
    })
  );
  const rotation = useRef(new Animated.Value(0.01)).current;
  const bearingDegree = rotation.interpolate({
    inputRange: [0, 360],
    outputRange: ['0deg', '360deg'],
  });

  function animateMarker(newLocation) {
    const newCoordinate = {
      latitude: newLocation.Latitude,
      longitude: newLocation.Longitude,
      latitudeDelta: 0.025,
      longitudeDelta: 0.025,
    };
    if (newLocation.Bearing !== null) {
      Animated.timing(rotation, {
        toValue: newLocation.Bearing,
        useNativeDriver: Platform.OS !== 'android',
        easing: Easing.linear,
        duration: 250,
      }).start();
      setTimeout(() => {
        smoothNavigation();
      }, 250);
    } else {
      smoothNavigation();
    }

    function smoothNavigation() {
      coordinate.timing(newCoordinate).start();
    }
  }

  useEffect(() => {
    animateMarker(location);
  }, [location]);

  return location.Bearing ? (
    <Marker.Animated
      identifier={driver.Id}
      coordinate={coordinate}
      ref={marker => {
        setMarker(marker);
      }}
      {...rest}
    >
      <AnimatedImage source={driverMarkerImg} className="w-10 h-14" style={{resizeMode: 'contain', transform: [{rotate: bearingDegree}]}} />
    </Marker.Animated>
  ) : null;
}

[...]

PR #4875, opened 2023-11-03
fix(map): remove spamming warning from Map 

watislaf:

Yes, this is very similar to https://github.com/react-native-maps/react-native-maps/pull/4644

(https://github.com/react-native-maps/react-native-maps/issues/4536)

### How did you test this PR?

locally 

Before fixing on IOS this looked like this:
drawing

### DIFF
--- a/src/MapView.tsx
+++ b/src/MapView.tsx
@@ -731,9 +731,9 @@ class MapView extends React.Component<MapViewProps, State> {
    * See https://reactnative.dev/docs/new-architecture-library-intro#migrating-off-setnativeprops
    */
   setNativeProps(props: Partial<NativeProps>) {
-    console.warn(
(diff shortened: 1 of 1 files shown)


salah-ghanim (approved):

salah-ghanim:
@watislaf I'm rejecting this pull request and submitting a different one because it simply commits out the code instead of deleting it ...

salah-ghanim:
@watislaf can you please re-create another version of the pull request with both the console.log is deleted and disabling of ts-warning @ts-ignore     this.marker.current?.setNativeProps(props);

I will work on updating RN to higher than 0.72 which should hopefully introduce again setNativeProps as type

salah-ghanim:
@watislaf I'm closing this, since react-native reintroduced setNativeProps. the log is now removed from the source code as part of the update to react-native 0.73

Issue #4862, opened 2023-10-11
EXC_BAD_ACCESS in iOS simulator when onMapReady receives null


kueda:
### Summary

When <MapView>'s `onMapReady` prop receives a value or `null` or `undefined`, the app crashes with a `EXC_BAD_ACCESS` error.

### Reproducible sample code

```TSX
import React from 'react';
import MapView from 'react-native-maps';

export default function App() {
  return (
    <MapView
      style={{width: "100%", height: "100%"}}
      onMapReady={null}
    />
  );
}
```

### Steps to reproduce

Set `onMapReady={null}` on any `<MapView />` and run it in an iOS simulator.

### Expected result

A component that receives an optional prop with a null value should behave the same as if it didn't receive the prop at all, e.g. `<MapView />` should be the same as <MapView onMapReady={null} />.

### Actual result

<img width="741" alt="Screenshot 2023-10-10 at 5 18 58 PM" src="https://github.com/react-native-maps/react-native-maps/assets/22691/bc1f8634-2576-4085-aab6-3651f7868327">

The app crashes and the simulator shows a popup saying "APP NAME quit unexpectedly".

### React Native Maps Version

1.8.0

### What platforms are you seeing the problem on?

iOS (Apple Maps)

### React Native Version

0.70.9

### What version of Expo are you using?

Not using Expo

### Device(s)

iPhone 14 (Simulator; iOS 16.4)

salah-ghanim:
@kueda I can't reproduce the issue under the following environment 

```
System:
    OS: macOS 14.1.1
    CPU: (10) arm64 Apple M1 Pro
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.17.0 - /var/folders/9x/9xzxgmhn75q66kr2x3n691hw0000gn/T/yarn--1700253494780-0.7101434196191923/node
    Yarn: 1.22.19 - /var/folders/9x/9xzxgmhn75q66kr2x3n691hw0000gn/T/yarn--1700253494780-0.7101434196191923/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v18.17.0/bin/npm
    Watchman: 2023.05.22.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.12.1 
  SDKs:
    iOS SDK:
      Platforms: DriverKit 23.0, iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0
    Android SDK: Not Found
[...]

PR #4887, opened 2023-11-10
fix: crash on iOS when onMapReady is null

mateki0:

No

https://github.com/react-native-maps/react-native-maps/issues/4862

### How did you test this PR?

I passed onMapReady={null} to MapView in one of library examples and ran it on iOS simulator. The changes will affect only iOS. I did not checked it on physical device, but it won't matter in this case

### DIFF
--- a/ios/Maps/RNMMapManager.m
+++ b/ios/Maps/RNMMapManager.m
@@ -543,7 +543,10 @@ static int kDragCenterContext;
 - (void)mapViewWillStartRenderingMap:(RNMMap *)mapView
 {
     if (!mapView.hasStartedRendering) {
-      mapView.onMapReady(@{});
+      if(mapView.onMapReady) {
+        mapView.onMapReady(@{});
+      }
+
       mapView.hasStartedRendering = YES;
     }
 }


salah-ghanim:
@mateki0 I'm not excited to merge this pull request ... it seems to be specific for the version of RN where the bug was reported ... on a more recent react-native version I tested the reported bug and I couldn't reproduce

mateki0:
> @mateki0 I'm not excited to merge this pull request ... it seems to be specific for the version of RN where the bug was reported ... on a more recent react-native version I tested the reported bug and I couldn't reproduce

@salah-ghanim Thanks for review. I was able to reproduce it in react-native-maps example so it's not too hard to find a version that will crash. Most of the applications that I've worked on are not running on the newest react-native so that may be a real problem.

salah-ghanim:
@mateki0 I just tested on the example project and you can see in the codebase there is no onMapReady defined and the app works fine ... so not passing any value is working as expected.

in case you explicitly pass null typescript complains that null value is not allowed (if you have type detection correctly setup).

this is the warning I received when I did so:
`TS2322: Type 'null' is not assignable to type '((event?: NativeSyntheticEvent<{}> | undefined) => void) | undefined'.`


from my point of view, this is a redundant fix since the behaviour is not supported based on the typing of the prop.

I will close this pull request for now :)

Issue #4835, opened 2023-08-29
Crash on android


haiderTkxel:
### Summary

Android app crashes. If I remove map view code then crash don't happen:

FATAL EXCEPTION: main
    Process: com.signaturepharmacy, PID: 16721
    com.facebook.react.bridge.NoSuchKeyException: pitch
        at com.facebook.react.bridge.ReadableNativeMap.getValue(ReadableNativeMap.java:113)
        at com.facebook.react.bridge.ReadableNativeMap.getValue(ReadableNativeMap.java:117)
        at com.facebook.react.bridge.ReadableNativeMap.getDouble(ReadableNativeMap.java:154)
        at com.rnmaps.maps.MapView.moveToCamera(MapView.java:570)
        at com.rnmaps.maps.MapView.applyBridgedProps(MapView.java:508)
        at com.rnmaps.maps.MapView.onMapReady(MapView.java:248)
        at com.google.android.gms.maps.zzaf.zzb(com.google.android.gms:play-services-maps@@18.0.0:1)
        at com.google.android.gms.maps.internal.zzaq.zza(com.google.android.gms:play-services-maps@@18.0.0:5)
        at com.google.android.gms.internal.maps.zzb.onTransact(com.google.android.gms:play-services-maps@@18.0.0:3)
        at android.os.Binder.transact(Binder.java:1084)
        at m.fh.c(:com.google.android.gms.dynamite_mapsdynamite@233217044@23.32.17 (190400-0):8)
        at com.google.maps.api.android.lib6.impl.bg.run(:com.google.android.gms.dynamite_mapsdynamite@233217044@23.32.17 (190400-0):19)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loopOnce(Looper.java:233)
        at android.os.Looper.loop(Looper.java:344)
        at android.app.ActivityThread.main(ActivityThread.java:8249)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:589)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1071)

### Reproducible sample code

```TSX
<MapView
          style={styles.mapView}
          provider={PROVIDER_GOOGLE}
          initialRegion={currentLocation}
          showsUserLocation={true}
          initialCamera={initialCamera}
[...]

Issue #4807, opened 2023-07-31
When the prop "altitude" is not set the MapView does not work on Android


felipehidra:
### Summary

If you skip the prop "pitch" on the MapView, the following error occurs on opening the app on Android (The app opens and close. I got this error on LogCat of the Android studio):
`FATAL EXCEPTION: main
           Process: com.<package>, PID: 29312
           com.facebook.react.bridge.NoSuchKeyException: pitch
`

The same problem occurs if the prop "heading" is skipped. This should raise an Exception or an Error on the build.

### Reproducible sample code

```TSX
import MapView, { PROVIDER_GOOGLE } from 'react-native-maps'

<MapView
    showsCompass={false}
    toolbarEnabled={false}
    showsUserLocation
    provider={PROVIDER_GOOGLE}
    maxZoomLevel={16}
    minZoomLevel={12}
    style={Styles.map}
    ref={mapRef}
    initialCamera={{
        center: {
            latitude: userCoordinate[0],
            longitude: userCoordinate[1],
        },
        pitch: 85, // <- if I remove this prop, the error occurs on the Android
        heading: 30,
        zoom: 14,
    }}
    onRegionChange={handleOnStartChangeArea}
    onRegionChangeComplete={handleChangeArea}
>
    {markers}
</MapView>
```

### Steps to reproduce

1 - Import the MapView component
2 - Set the prop "initialCamera" and remove/skip de camera prop "pitch"
3 - Run the command "yarn android"
4 - The app will open and close in sequence without any error

### Expected result

The expected result should be the map appears on the coordinates specified on the prop "initialCamera.center".

### Actual result

The app opens and close as soon as the MapView Component is called!

### React Native Maps Version

1.7.1

### What platforms are you seeing the problem on?

Android

### React Native Version

0.72.3

### What version of Expo are you using?

Not using Expo

### Device(s)

Motorola moto g(30)

haiderTkxel:
@felipehidra Did you find any solution?

felipehidra:
@haiderTkxel No. If the prop "pitch" is required, this should throw an error, no?

salah-ghanim:
@felipehidra  not an issue ... please checkout typing (specifically MapView.types.ts / Camera type where pitch is explicitly required

PR #4889, opened 2023-11-16
Fix/initial camera crash android

mateki0:

No

https://github.com/react-native-maps/react-native-maps/issues/4835
https://github.com/react-native-maps/react-native-maps/issues/4807

### How did you test this PR?

I've tested the code on Android physical device and iOS simulator. Both works fine. The change will only affect Android system.
To test the changes you have to:
1. Go to the CameraControl.tsx in examples.
2. Remove any of these from initialCamera passed to MapView - ['center', 'heading', 'pitch', 'zoom'];
3. You should see Uncaught Error saying whats wrong with your code.
4. Previously there was no error and app was just killed 

Those linked issues could be fixed without changing native code but I've added it to prevent similar issues with other props in the future

### DIFF
--- a/android/src/main/java/com/rnmaps/maps/MapView.java
+++ b/android/src/main/java/com/rnmaps/maps/MapView.java
@@ -523,25 +523,30 @@ public class MapView extends com.google.android.gms.maps.MapView implements Goog
 
   // called as soon as the map is ready. Assuming nothing was set on the map at this point.
   private void applyBridgedProps() {
-    if(boundary != null) {
-      map.setLatLngBoundsForCameraTarget(boundary);
+    try {
+      if(boundary != null) {
+        map.setLatLngBoundsForCameraTarget(boundary);
+      }
+      if(initialRegion != null) {
+        moveToRegion(initialRegion);
+        initialRegionSet = true;
+      } else if(initialCamera != null) {
+        moveToCamera(initialCamera);
+        initialCameraSet = true;
+      } else if(region != null) {
+        moveToRegion(region);
+      } else {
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/rnmaps/maps/MapView.java
```diff
@@ -523,25 +523,30 @@ public void setBoundary(ReadableMap boundary) {
...
+      } else if(region != null) {
+        moveToRegion(region);
+      } else {
+        moveToCamera(camera);
+      }
+      if(customMapStyleString != null) {
+        map.setMapStyle(new MapStyleOptions(customMapStyleString));
+      }
+    } catch (Exception e) {
+        System.err.println("Fail during applying MapView props");
```
salah-ghanim, line 545: @mateki0 I would say never silently catch an exception in this library, especially when it's a configuration issue by the user of the library.

and please avoid moving functionality from native to react-native because the same native code might end up being called from a different place from react-native and the behaviour becomes unpredictable

salah-ghanim:
@mateki0 this is a "non-issue" the Camera type in MapView.types.ts specifically show pitch as non-optional ... if users go against library documentation it's normal that the library crashes.

mateki0:
But the problem here is with no error thrown if user go against library documentation. I think that no one have problem with a crash, but right now it is crashing without any information what's wrong.

salah-ghanim:
@mateki0 from my point of view zoom should be correctly documented.

going against documentation (passing value that should not be passed) can result in an unexpected behaviour including crashing, I don't see any problem in that.

Issue #4880, opened 2023-11-06
MapKit Globe view camera resets when mov


subtleGradient:
### Summary

When viewing a map in globe view, occasionally the camera/region get reset back to 0.

Here I'm trying to drag the north pole into view, but when I get too far north it resets back out
![Simulator Screen Recording - 13 Mini - 2023-11-06 at 15 17 35](https://github.com/react-native-maps/react-native-maps/assets/4117/dd88b37a-7919-47c6-bae7-20e15c60c23f)

looks like it's related to this code:
https://github.com/react-native-maps/react-native-maps/blob/4514bd13a34e20f7e82e7de0b44224b6160195d5/ios/AirMaps/AIRMapManager.m#L1167-L1174

### Reproducible sample code

```TSX
import { Platform } from "react-native";
import MapView from "react-native-maps";

export function DevMap_camResetBugRepro() {
  const mapType =
    Platform.OS === "ios"
      ? ("hybridFlyover" as any) // NOTE: casting to any because this is not in the type definition, but it works to set the map type to a 3D globe
      : "standard";

  return (
    <MapView
      style={{ flex: 1 }}
      mapType={mapType}
      initialRegion={{
        latitude: 14,
        longitude: 30,
        latitudeDelta: 156,
        longitudeDelta: 173,
      }}
    />
  );
}
```

### Steps to reproduce

Drag the north pole into view

### Expected result

see the north pole

### Actual result

region/camera resets back out to latitude 0

### React Native Maps Version

1.7.1

### What platforms are you seeing the problem on?

iOS (Apple Maps)

### React Native Version

0.72.3

### What version of Expo are you using?

SDK 48

### Device(s)

iPhone 13 mini

subtleGradient:
also related: https://github.com/react-native-maps/react-native-maps/blob/4514bd13a34e20f7e82e7de0b44224b6160195d5/ios/AirMaps/AIRMapManager.m#L1090-L1093

subtleGradient:
looks like this is the code that triggers the zoom out:
https://github.com/react-native-maps/react-native-maps/blob/4514bd13a34e20f7e82e7de0b44224b6160195d5/ios/AirMaps/AIRMapManager.m#L951-L956

subtleGradient:
Workaround:
```ts
```

jan-kozinski:
Hi, that's because the calculation of zoom level doesn't take into account the map's rotation. I'll try to find a better algorithm

[...]

PR #4892, opened 2023-11-20
fix(RNMMap): include map rotation in zoom computing & fix flyovers maps zoom constrains

jan-kozinski:

No

https://github.com/react-native-maps/react-native-maps/issues/4880

The zoom limit algorithm does not take map rotation into account and also it does not work for flyover maps. Flyover maps are not typed and not documented.

### How did you test this PR?

iPhone 14 Pro simulator, I logged the zoom values with NSLog. The new algorithm produced same zoom values as the old one did when map was not rotated. When the map was rotated, the old algorithm tended to be unpredictable.


KrzysztofMoch (approved):

jan-kozinski:
@salah-ghanim can you take a look please?

salah-ghanim:
@jan-kozinski please re-submit the pull request against master so that it can be released.

PR #4897, opened 2023-11-24
fix(markdown): table format

wael-fadlallah:

No

Props table format

### How did you test this PR?

its a simple markdown change, looked good in review tab tho


salah-ghanim:
@wael-fadlallah this pull request can't be merged as is because prettier is complaining about the .md file

you can test it yourself:
`yarn format-check`

I recommend you apply a prettier ignore rule to all *.md files

see https://prettier.io/docs/en/ignore.html

Issue #4860, opened 2023-10-06
MarkerAnimated not working


MilosDanilovic98:
### Summary

 I'm trying to make an app using react-native-maps. I got everything setup with Expo and i have a map loading correctly. The issue is that I'm trying to animate the MapMarker component but nothing is happening. Even if i just copy paste the example 
https://github.com/react-native-maps/react-native-maps/blob/master/example/src/examples/AnimatedMarkers.tsx
still nothing is happening. I don't get any errors or any messages but the marker is not moving at all. Does anyone know what could be the issue?

### Reproducible sample code

```TSX
import React from 'react';
import {
  StyleSheet,
  View,
  Text,
  Dimensions,
  TouchableOpacity,
  Platform,
} from 'react-native';

import MapView, {Marker, AnimatedRegion} from 'react-native-maps';

const screen = Dimensions.get('window');

const ASPECT_RATIO = screen.width / screen.height;
const LATITUDE = 37.78825;
const LONGITUDE = -122.4324;
const LATITUDE_DELTA = 0.0922;
const LONGITUDE_DELTA = LATITUDE_DELTA * ASPECT_RATIO;

class AnimatedMarkers extends React.Component<any, any> {
  marker: any;
  constructor(props: any) {
    super(props);

    this.state = {
      coordinate: new AnimatedRegion({
        latitude: LATITUDE,
        longitude: LONGITUDE,
      }),
    };
  }

  animate() {
    const {coordinate} = this.state;
    const newCoordinate = {
      latitude: LATITUDE + (Math.random() - 0.5) * (LATITUDE_DELTA / 2),
      longitude: LONGITUDE + (Math.random() - 0.5) * (LONGITUDE_DELTA / 2),
    };

    if (Platform.OS === 'android') {
      if (this.marker) {
        this.marker._component.animateMarkerToCoordinate(newCoordinate, 500);
      }
    } else {
      // `useNativeDriver` defaults to false if not passed explicitly
      coordinate.timing({...newCoordinate, useNativeDriver: true}).start();
    }
  }

  render() {
    return (
      <View style={styles.container}>
        <MapView
          provider={this.props.provider}
          style={styles.map}
          initialRegion={{
            latitude: LATITUDE,
            longitude: LONGITUDE,
            latitudeDelta: LATITUDE_DELTA,
            longitudeDelta: LONGITUDE_DELTA,
          }}>
          <Marker.Animated
            ref={(marker: any) => {
              this.marker = marker;
[...]

Issue #4843, opened 2023-09-08
Can't set map's region prop to AnimatedRegion


moaazassali:
### Summary

I am following the Animated Region section on the main region and created an AnimatedRegion initial state as needed, but I cannot set an AnimatedRegion object to the region prop of the map. This is the error I get from that:
```
Type 'AnimatedMapRegion' is missing the following properties from type 'WithAnimatedObject<Region>':
latitude, longitude, latitudeDelta, longitudeDelta ts(2739)

MapView.d.ts(358, 5): The expected type comes from property 'region' which is declared here on type
'IntrinsicAttributes & AnimatedProps<ViewProps & { cacheEnabled?: boolean | undefined; camera?:
Camera | undefined; compassOffset?: Point | undefined; ... 63 more ...; zoomTapEnabled?: boolean |
undefined; } & RefAttributes<...>>'
```

### Reproducible sample code

```TSX
const initialRegion = new AnimatedRegion({
  latitude: 51.5072,
  longitude: 0.1276,
  latitudeDelta: 0.0922,
  longitudeDelta: 0.0421,
});

const MapScreen = () => {
  const [region, setRegion] = useState<AnimatedRegion>(initialRegion);

  return (
    <AnimatedMapView
      style={{height: '100%', width: '100%'}}
      region={region} />
      ...
    </AnimatedMapView>
  )
};
```

### Steps to reproduce

Create a react native, download react-native-maps and try to assign AnimatedRegion to region prop. The error will pop up on the IDE.

### Expected result

I should be able to assign the region prop to an AnimatedRegion object as is mentioned on the main page under the Animated Region section.

### Actual result

Received an error in the IDE instead.

### React Native Maps Version

1.7.1

### What platforms are you seeing the problem on?

Android, iOS (Apple Maps), iOS (Google Maps)

### React Native Version

0.72.4

### What version of Expo are you using?

Not using Expo

### Device(s)

Doesn't matter

Issue #4778, opened 2023-06-26
MarkerAnimated prop coordinate has wrong TypeScript type


karatekid430:
### Summary

The README.md for v1.7.1 says that MarkerAnimated will also take an AnimatedRegion in the coordinate prop. And it does. But its type only shows LatLng, meaning that passing AnimatedRegion is flagged as an error when it should not be.

### Reproducible sample code

```TSX
const coord = new AnimatedRegion();

<MarkerAnimated
  coordinate={coord} // ts should not show an error here
/>
```

### Steps to reproduce

Install, pass AnimatedRegion to coordinate prop of MarkerAnimated with TypeScript

### Expected result

It should say that passing an AnimatedRegion to coordinate prop of MarkerAnimated is not an error

### Actual result

It says that passing an AnimatedRegion to coordinate prop of MarkerAnimated is an error

### React Native Maps Version

1.7.1

### What platforms are you seeing the problem on?

iOS (Apple Maps)

### React Native Version

0.70.6

### What version of Expo are you using?

Not using Expo

### Device(s)

Simulator

mcoppola:
➕

ackermanGitHub:
anyone solved how to animate a marker smoothly on the latests versions of expo and react-native?

PR #4909, opened 2023-12-13
fix: fix AnimatedRegion types

mateki0:

No

https://github.com/react-native-maps/react-native-maps/issues/4843
https://github.com/react-native-maps/react-native-maps/issues/4778

### How did you test this PR?

- It should affect only types, not the functionality.
- I've tested it on: iOS simulator (iPhone 12), Android real device (Samsung M23)
- I wasn't able to test it on iPhone real device

It should work as it is now on both platforms. However I have problem on Android connected with this code because _component is undefined. If i remove **_component** and leave **this.marker.animateMarkerToCoordinate(newCoordinate, 500)** it works fine.

```tsx
// example/src/examples/AnimatedMarkers.tsx

animate() {
   ...
    if (Platform.OS === 'android') {
      if (this.marker) {
        this.marker._component.animateMarkerToCoordinate(newCoordinate, 500);
      }
    } 
  }
```

On iOS nothing happens after pressing Animate button, but I didn't want to fix it here because it was working the same before my changes and it is already reported in another [issue](https://github.com/react-native-maps/react-native-maps/issues/4860). And it should be fixed in separate pull request


salah-ghanim (changes requested):
@mateki0  why not extend Region type instead ? we can avoid duplication and in the future if Region object changes AnimatedRegion will be equally updated which is what we want.

mateki0:
@salah-ghanim I'm not sure if it's what you meant but I've updated region type in MapView.tsx instead of making changes in AnimatedRegion itself

FawadAhmad001 (approved):

salah-ghanim:
@mateki0 please rebase the branch on master so that I merge this type fix.

mateki0:
> @mateki0 please rebase the branch on master so that I merge this type fix.

Hi, I just did it, thanks for merging

salah-ghanim:
@mateki0 please rebase again so that I finally merge this as well.

mateki0:
@salah-ghanim done :)

salah-ghanim:
@mateki0 your branch is out of date again

brianbowden:
Always love it when a fix gets released just in time for me to need it. :) Thanks for the effort here!

PR #4931, opened 2024-01-10
docs: fix formatting of table with escaped pipe characters

noquinn:

#4897 attempts to fix the formatting but is incompatible with the Prettier config. 
This PR uses escaped pipe symbols to fix the formatting in a way that is compatible with Prettier.

docs/geojson.md props table formatting

### How did you test this PR?

Markdown preview looks good and `yarn lint` / `yarn format-check` is happy :)


salah-ghanim:
thanks @noquinn 🔥

Issue #4948, opened 2024-01-31
customMapStyle has no effect when using Apple maps as provider


gkpo:
### Summary

I'm running an expo based app using the react-native-maps plugin. For iOS it's using Apple Maps. When I try to add custom styling to the map using customMapStyle, I have no effect at all.

### Reproducible sample code

```TSX
import React from 'react';
import { View, StyleSheet } from 'react-native';
import MapView from 'react-native-maps';

const MapScreen = () => {
  const customMapStyles = [
    {
      elementType: 'geometry',
      stylers: [{ color: '#ff0000' }], // Red color for map geometry
    },
    {
      featureType: 'road',
      elementType: 'geometry',
      stylers: [{ color: '#ff0000' }], // Red color for roads
    },
  ];

  return (
    <View style={styles.container}>
      <MapView
        style={styles.map}
        initialRegion={{
          latitude: 37.7749, // Initial latitude for the map
          longitude: -122.4194, // Initial longitude for the map
          latitudeDelta: 0.0922,
          longitudeDelta: 0.0421,
        }}
        customMapStyle={customMapStyles} // Set custom map styles
      />
    </View>
  );
};

const styles = StyleSheet.create({
  container: {
    flex: 1,
  },
  map: {
    flex: 1,
  },
});

export default MapScreen;
```

### Steps to reproduce

Run the above code

### Expected result

A map with red roads

### Actual result

A map with no particular styling applied

### React Native Maps Version

1.3.2

### What platforms are you seeing the problem on?

iOS (Apple Maps)

### React Native Version

0.71.8

### What version of Expo are you using?

SDK 48

### Device(s)

iPhone 14 Simulator

thomasviaud:
This prop is only working with Google provider. https://mapstyle.withgoogle.com/

PR #4950, opened 2024-02-01
Clarify documentation about customMapStyles being only effective with Google provider

gkpo:

No

[customMapStyle has no effect when using Apple maps as provider #4948](https://github.com/react-native-maps/react-native-maps/issues/4948)

### How did you test this PR?

This is a change in the Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ render() {
 }
 ```
 
-### Customizing the map style
+### Customizing the map style (Google Maps provider only)
 
 Create the json object, or download a generated one from the [google style generator](https://mapstyle.withgoogle.com/).
 
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ render() {
 }
 ```
 
-For iOS, in addition to providing the `mapStyle` you will need to do the following
+This option only works with the Google provier. So for iOS, in addition to providing the `mapStyle` you will need to do specify the provier with the following:
 
 ```jsx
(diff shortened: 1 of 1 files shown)


salah-ghanim:
@gkpo would be nice if you update the documentations instead in two ways:
1. the section title: Customizing the map style could have (Provider: GoogleMaps) as part of the title
2. mention that this is a depricated way of customyzing style instead using google maps cloud and using mapId is the way to go (this feature was introduced recently but not well documented and doesn't work yet for iOS).

README.md
```diff
@@ -246,7 +246,7 @@ render() {
 }
 ```
 
-### Customizing the map style
+### Customizing the map style (Google Maps provider only)
```
jan-kozinski, line 249:
```suggestion
### Customizing the map style (Google Maps only)
```

README.md
```diff
@@ -265,7 +265,7 @@ render() {
 }
 ```
 
-For iOS, in addition to providing the `mapStyle` you will need to do the following
+This option only works with the Google provier. So for iOS, in addition to providing the `mapStyle` you will need to do specify the provier with the following:
```
jan-kozinski, line 268:
```suggestion
For iOS, in addition to providing the `mapStyle` you will need to set the provider to Google Provider as in:
```

charneykaye, line 268: typo: should be "Google provider"

PR #4971, opened 2024-02-15
docs: elaborate on the performance impact of the tracksViewChanges prop

jared-ca:

No

This PR elaborates a little on the tracksViewChanges prop for the Marker component.

When I read the docs I first assumed it was talking about tracking changes to the MapView so that the marker would still stay in the same position when the map was panned. Next I assumed it would need to be true if any details of the marker changed such as the position.

 I only understood what it actually did once I read the source so I figured the docs could use a bit of elaboration.

### How did you test this PR?

NA


jan-kozinski (approved):
Thanks for adding this

salah-ghanim (approved):

Issue #4963, opened 2024-02-11
Polyline doesn't update when the coordinates array is changed.


panpan31415:
### Summary

I want to use the Polyline component from React-Native-Maps. It requires an array of coordinates to draw a polyline. However, when I update the coordinate array, the polyline doesn't update accordingly.

### Reproducible sample code

```TSX
import React, { useState } from "react";
import { Pressable, StyleSheet, Text, View } from "react-native";
import MapView, { PROVIDER_GOOGLE, Polyline } from "react-native-maps";

const coordinates = [
  { latitude: 55.861324310302734, longitude: 12.387187004089355 },
  { latitude: 55.8612174987793, longitude: 12.387340545654297 },
  { latitude: 55.86132049560547, longitude: 12.387341499328613 },
  { latitude: 55.86111831665039, longitude: 12.38720417022705 },
  { latitude: 55.861358642578125, longitude: 12.387207984924316 },
  { latitude: 55.86152648925781, longitude: 12.387125015258789 },
  { latitude: 55.866825103759766, longitude: 12.383121490478516 },
  { latitude: 55.87004089355469, longitude: 12.379796028137207 },
  { latitude: 55.87102508544922, longitude: 12.381945610046387 },
  { latitude: 55.870330810546875, longitude: 12.384469985961914 },
  { latitude: 55.870487213134766, longitude: 12.387078285217285 },
  { latitude: 55.86918258666992, longitude: 12.387304306030273 },
  { latitude: 55.86808395385742, longitude: 12.387526512145996 },
  { latitude: 55.8694953918457, longitude: 12.38739013671875 },
  { latitude: 55.86945343017578, longitude: 12.38755989074707 },
  { latitude: 55.86931610107422, longitude: 12.387386322021484 },
  { latitude: 55.868534088134766, longitude: 12.387174606323242 },
  { latitude: 55.86812973022461, longitude: 12.38566780090332 },
  { latitude: 55.8681526184082, longitude: 12.3834867477417 },
  { latitude: 55.86741256713867, longitude: 12.383044242858887 },
  { latitude: 55.867210388183594, longitude: 12.38311767578125 },
  { latitude: 55.86640548706055, longitude: 12.383851051330566 },
  { latitude: 55.865966796875, longitude: 12.382963180541992 },
  { latitude: 55.86558532714844, longitude: 12.38154411315918 },
  { latitude: 55.86520767211914, longitude: 12.379416465759277 },
[...]

PR #4973, opened 2024-02-17
fix(iOS): removing polylines on iOS with googleProvider

mateki0:

No

https://github.com/react-native-maps/react-native-maps/issues/4963

### How did you test this PR?

- It affects only iOS with GoogleProvider
- I have tested it on simulator, there shouldn't be difference on real device

From what I can see the same problem will be with Polygon but I will fix it in another PR if this one will be merged.


salah-ghanim:
@mateki0 thank you for trying to fix this, I just reviewed google's APIs and there is a simple one for GMSMutablePath removeAllCoordinates 

[docs](https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_mutable_path#aedcebc615273a6fd648527e6f546a7f2)

Please use that instead of introducing a reference to the map and actively removing the polyline itself from the map each time the coordinates are reset.

ios/AirGoogleMaps/AIRGoogleMapPolyline.m
```diff
@@ -29,11 +29,25 @@ -(void)setCoordinates:(NSArray<AIRMapCoordinate *> *)coordinates
   _coordinates = coordinates;
 
   GMSMutablePath *path = [GMSMutablePath path];
-  for(int i = 0; i < coordinates.count; i++)
+
+  if (!coordinates || coordinates.count == 0) 
   {
+    _polyline.map = nil; // Remove polyline from the map
```
salah-ghanim, line 35: simply use removeAllCoordinates API instead

salah-ghanim (changes requested):
please

ios/AirGoogleMaps/AIRGoogleMapPolyline.m
mateki0: Unfortunately it doesn't work. Also I've tried to assign nil to the path instead of map but without success. This is my code:

```
if (!coordinates || coordinates.count == 0) 
  {
    [path removeAllCoordinates]; // Doesn't work
    // path = nil // Doesn't work
    // _polyline.map = nil; // Remove polyline from the map
    return;
  }
 ```
 
 @salah-ghanim

salah-ghanim:
@mateki0 please rebase so that I merge this :)

salah-ghanim (approved):

mateki0:
@salah-ghanim done :)

Issue #4970, opened 2024-02-15
provider={PROVIDER_GOOGLE} doesn't allow onSelect and onDeselect on iOS


rarenatoe:
### Summary

When using provider `google`, the Marker props onSelect and onDeselect do not work on iOS (which supposedly does work). This should either be fixed or specified in the docs that it doesn't for this case.

### Reproducible sample code

```TSX
import React from 'react'
import { Text } from 'react-native'
import MapView, { Marker, Callout, PROVIDER_GOOGLE } from 'react-native-maps'

<MapView
  provider={PROVIDER_GOOGLE}
  customMapStyle={MapStyle}
>
  <Marker coordinate={someCoordinate}
    onSelect={() => console.log('onSelect called')}
    onDeselect={() => console.log('onDeselect called')}
  >
    <Callout>
      <Text>Hola</Text>
    </Callout>
  </Marker>
</MapView>
```

### Steps to reproduce

1. Find the marker on the map.
2. Press the marker.
3. Check that there is **no** log `onSelect called`.
4. Press anywhere else in the map.
5. Check that there is **no** log `onDeselect called`.

### Expected result

onSelect and onDeselect should work.

### Actual result

onSelect and onDeselect do not work.

### React Native Maps Version

1.10.1

### What platforms are you seeing the problem on?

iOS (Google Maps)

### React Native Version

0.70.9

### What version of Expo are you using?

Not using Expo

### Device(s)

iPhone 14 Pro simulator

jan-kozinski:
Hi, the onSelect and onDeselect props are not supported on Google Maps. I'll update the readme accordingly. Thanks for reporting!

salah-ghanim:
@jan-kozinski I will merge your pull request to update the docs, but do you want to create a pull request for implementing marker select / deselect functionality on google maps ? 
see this from chatgpt:

> 
> Detecting marker selection and deselection in the Google Maps SDK for both iOS and Android involves different approaches tailored to each platform. Below, I'll outline how to implement these functionalities for both environments.
> 
> ### For iOS (Google Maps SDK for iOS)
> 
> In Google Maps SDK for iOS, you can detect marker selection (tap) and deselection by implementing the `GMSMapViewDelegate` methods. Specifically, you'll use `mapView:didTapMarker:` to detect when a marker is tapped (selected) and potentially `mapView:didTapAtCoordinate:` to manage deselection, since there isn't a direct method for detecting marker deselection.
> 
[...]

Issue #114, opened 2016-03-02
MapView.Marker onSelect support for Android


stinju:
Is MapView.Marker onSelect support for Android on your short-term roadmap?

alvelig:
https://stackoverflow.com/questions/29345928/changing-the-icon-of-google-maps-marker-while-its-selected
https://stackoverflow.com/questions/29740839/how-to-select-and-deselect-a-marker-in-google-maps-in-android

tobiasf:
Any updates on this? Is MapView.Marker going to get onSelect and onDeselect support for Android too?

alvelig:
@tobiasf Not likely. It's out of the focus as far as I know. But I left those links if someone wants to contribute.

christopherdro:
I’m closing this issue because it has been inactive for a while. This probably means that it is not reproducible or it has been fixed in a newer version. If it’s an enhancement and hasn’t been taken on for so long, then it seems no one has the time to implement this.

Please reopen if you still encounter this issue with the [latest stable version](https://github.com/react-native-community/react-native-maps/releases).

Thank you!

christopherdro:
I’m closing this issue because it has been inactive for a while. This probably means that it is not reproducible or it has been fixed in a newer version. If it’s an enhancement and hasn’t been taken on for so long, then it seems no one has the time to implement this.

Please reopen if you still encounter this issue with the [latest stable version](https://github.com/react-native-community/react-native-maps/releases).

Thank you!

chrise86:
Just ran into this issue. When you say those events are not supported on Android, does that actually mean Google Maps? As it does not seem to work for me on iOS using Google Maps.

PR #4990, opened 2024-03-03
feat(AIRGoogleMapsMarker): add onSelect/onDeselect support fo google maps

jan-kozinski:

no

https://github.com/react-native-maps/react-native-maps/issues/4970

https://github.com/react-native-maps/react-native-maps/issues/114
### How did you test this PR?

andoid emulator & ios simulator


EJohnF:
Wow! i was looking for this functionality, and you just pushed it 🎉

mateki0 (approved):
Looks and works fine on iOS simulator and Android device. I'm just thinking if second press on the same marker shouldn't be considered as deselect on iOS. It works like that on Android though.

jan-kozinski:
Thanks @mateki0 good catch! I've tweaked the google maps implementation to match the mapkit one's. So no it should be impossible to deselect a marker by clicking on it multiple times

salah-ghanim:
@jan-kozinski the code looks good, but I have a small feedback

I wouldn't change the AIRGoogleMapMarker.m eventFromMarker AIRGMSMarker.m for some reason that class was left empty all the history of the repo, I would simply keep the same tradition.

instead just modify the function in AIRGoogleMapMarker to accept action as a parameter, for example:

```
eventFromMarker: marker action:(_Nullable NSString *) action {
}
```

finally, In general I would avoid adding anything to the .h files, because that implies new public interfaces / functionality ... I don't see the need for such APIs most of the time, rather the body of a class (.m file) can manage the state internally and propagate the events correctly as you can see.

jan-kozinski:
Hi @salah-ghanim thanks for the feedback. I've added the makeEventData to the AIRGMSMarker, because selectedMarker in AIRGoogleMap must ba an instance of GSMMarker (it can't be a AIRGoogleMapMarker because the  `[super setSelectedMarker:selectedMarker];` expects a AIRGMSMarker. 
Still, I wanted both classes to use the same method for event generating as to avoid incosistent marker events in future.
I think the reason for AIRGMSMarker.m to be left empty is that it simply didn't require any functionality in the past. I don't consider it an intentinal action as there is no comment that would document such decision

jan-kozinski:
Also thanks for the advice with .h files, I've removed as many public interfaces as possible. Removing them further causes the code to not compile.

salah-ghanim (approved):

PR #5005, opened 2024-03-17
feat(ios): Add possibility to use both MKMarkerAnnotationView and MKPinAnnotationView

mateki0:

No

There's no issue for it but it was reported in discussions

https://github.com/react-native-maps/react-native-maps/discussions/4975#discussioncomment-8795595

### How did you test this PR?

It affects only iOS and Apple Maps. I've tested it on simulator. To test all functionalities you can use examples:
- Draggable Markers
- Custom Callout
- Default Markers
- Animated Marker Position

I decide to stay with pinColor name even though it's marker now. But we have the same name on Android and Google Maps for iOS so markerColor could be confusing for users.


jan-kozinski:
Hi, I've tested it a bit and it seems like the marker color feature is no longer working
image
image
Also I think we should set the default value for "titleVisibility" and "subTitleVisibility" to  0, at least when a custom callout is set so to not show the text twice We can expose those two properties as new properties for the marker component to allow customization.
image
image

salah-ghanim:
@mateki0  @jan-kozinski 
 1. there is https://developer.apple.com/documentation/mapkit/mkmarkerannotationview/2873822-markertintcolor MKMarkerAnnotationView so it should be replaceable.
 2. we shouldn't merge until the functionality is as similar as possible to MKPinAnnotationView, I see the feedback regarding the text (title, subtitle) it needs to be addressed.
 3. we shouldn't simply replace because this will be a breaking change that might introduce unexpected behaviour, I would recommend we introduce a property call it legacy, or usePinView that can be set to true for an annotation so that users can keep the old behaviour, new users or anyone who wants to migrate then can simply not set such property and use the new mkmarkerannotationview

mateki0:
Hi @salah-ghanim. First two points are already addressed. I didn't mention it here because I was waiting for Jan's re-review. 

If we want to not introduce breaking change shouldn't we make old pin used by default? And then add a flag for turning on a new one?

salah-ghanim:
@mateki0 good point, but my assumption is that most users don't depend on the pin behaviour rather add their custom markers, in this case it won't cause a breaking change.

hence using legacy flag will allow users who need this behaviour to enable it, everyone else will upgrade without changing anything about their code.

mateki0:
It's ready to re-review.
1. Title and subtitle are hidden by default like in legacy pin, and both can be enabled by props.
2. I've added useLegacyPinView prop which allow user to use MKPinAnnotationView. It's even possible to use legacy and new one on the same screen.
3. Changing colour works as it was before.

Thanks for your suggestions @salah-ghanim and @jan-kozinski

salah-ghanim (approved):

owattenmaker:
It would be nice to at least list this as a breaking change in the release notes. Not a huge deal to add the flag to my code when upgrading, but took a bit to 1, understand that there was a breaking change by upgrading, and 2, understand what was happening.

It's weird to assume that "most users" would be using custom pins when the default pin works well for simple cases. The most breaking change about this is that it groups pins together, which can lead to some weird looking maps depending on your usecase.

PR #5026, opened 2024-04-05
fix(ios-google): zoomTapEnabled works on map load

lucaslt89:
See the issue for a test example and the issue details

Closes issue #5024

No

#5024 

### How did you test this PR?
The issue #5024 contains a simple example to reproduce it.

This is a simple app to test it:

```
import MapView from 'react-native-maps';

const App = () => {
  const [zoomEnabled, setZoomEnabled] = useState(false);

  useEffect(() => {
    setTimeout(() => {
      setZoomEnabled(true);
    }, 1000);
  }, []);

  return (
    <MapView provider="google" style={{flex: 1}} zoomTapEnabled={zoomEnabled} />
  );
};

export default App;
```

### DIFF
--- a/ios/GoogleMaps/RNMGoogleMap.m
+++ b/ios/GoogleMaps/RNMGoogleMap.m
@@ -323,6 +323,9 @@ id regionAsJSON(MKCoordinateRegion region) {
   [self overrideGestureRecognizersForView:mapView];
 
   if (!_didCallOnMapReady && self.onMapReady) {
+    if(self.zoomTapEnabled == NO){
+        [self toggleGesture:@"handleZoomTapGesture:" toggle:self.zoomTapEnabled];
+    }
     self.onMapReady(@{});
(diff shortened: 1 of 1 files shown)


salah-ghanim (changes requested):
@lucaslt89 thank you very much for your contribution, after reviewing the code I realised that it depends on undocumented behaviour, instead of depending on google's official APIs, I would avoid such customisations as much as I can

please look into: https://developers.google.com/maps/documentation/ios-sdk/controls#maps_ios_controls_and_gestures_map-objective-c

and recently I introduced the ability to load initialProps on the native side for both android and iOS to solve this "the proper way"

see: AIRGoogleMapManager view method where initialProps are loaded.

you can load zoomGesturesEnabled from there and pass it on to google maps settings to enable / disable correctly on launch

salah-ghanim:
@lucaslt89 should be fixed in v1.15.2

PR #4922, opened 2023-12-24
New Feature: Support for Google Map Config (mapId, liteMode)

salah-ghanim:
### Does any other open PR do the same thing?
No

### What issue is this PR fixing?

Google Maps introduced a new styling capability via google cloud, this library didn't support any kind of initialisation on view creation before, and depended even on creating a custom module just to pass an initial property for Android: liteMode

This is now fixed and initial properties can be passed

supported properties: googleMapId (iOS, Android) , liteMode (Android only)

### How did you test this PR?
Physically on android and iOS devices, the values are being setup correctly.

Unfortunately for iOS without updating to google-maps SDK v 8.2 the style won't be visible .. waiting on google

PR #5027, opened 2024-04-08
Fix `liteMode` being set on new views if one used it previously

renchap:
In our app, some of the maps use `liteMode`, while some others do not.

Since version 1.9.0, if you rendered a map with `liteMode`, then another map without the `liteMode` prop, then the second map is displayed as if `liteMode={true}`. If you render it with `liteMode={false}` it behaves correctly.

This behaviour comes from this change: https://github.com/react-native-maps/react-native-maps/pull/4922/files#diff-7e4f66bee1e03862b509bbbc95c0961454226b558f6073a6b699adb5a29f5045R81

I am unsure of why it behaves this way (some props caching? component re-use?) but this PR restores the correct behaviour. I am unsure if something similar needs to be done for the other props around the changed lines.

No

### How did you test this PR?

I applied those changes locally and tested them with my own app, which fixed the issue.

### DIFF
--- a/android/src/main/java/com/rnmaps/maps/MapManager.java
+++ b/android/src/main/java/com/rnmaps/maps/MapManager.java
@@ -86,6 +86,8 @@ public class MapManager extends ViewGroupManager<MapView> {
             }
             if (initialProps.hasKey("liteMode")) {
                 googleMapOptions.liteMode(initialProps.getBoolean("liteMode", false));
+            } else {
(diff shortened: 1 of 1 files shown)


salah-ghanim:
@renchap thank you so much for your contribution and yes the bug is clear, but I had to re-do the fix via https://github.com/react-native-maps/react-native-maps/pull/5034 

because that way the bug will be fixed for all initialProp values not just liteMode.

PR #5040, opened 2024-04-24
Chore/bump example rn to 0.74

mateki0:

No

It's just a bump of example app react-native version. I want to tackle new architecture support after merging this one. 

### How did you test this PR?

It affects both platforms but only our example application.  
I have tested it on real device Android and iOS simulator.


example/ios/rnmshowcase/AppDelegate.mm
```diff
@@ -35,19 +31,4 @@ - (NSURL *)getBundleURL
 #endif
 }
 
-- (BOOL)turboModuleEnabled {
```
mateki0, line 38: I had to delete these booleans because it was causing crash on app init

mateki0:
Unfortunately I'm not able to fix pipelines for now because highest available xCode version in "macos latest" is 14.2, and there's an error that it requires 14.3. Other than that everything else should work fine.

example/ios/rnmshowcase/AppDelegate.mm
mateki0: Screenshot 2024-04-29 at 08 39 10

We should resolve this before merging since it's blocking us from using new arch on 0.74

salah-ghanim:
@mateki0 thanks for your contribution here, two things:
 1. Yes simply removing (fabricEnabled, newArchEnabled ...) seems like the right way to go, check out a new RN project using `npx react-native@latest init AwesomeProject` you will see that no such methods in the AppDelegate.
 
 2. please remove the changes to the ci files / github actions, you can see clearly that macos latest image does include xcode 15.3 and even 15.4 
 https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md
 we might need an xcode-select step, see: https://stackoverflow.com/questions/65821440/how-to-select-xcode-version-in-fastlane-when-using-github-actions
 
 but that's it.
 
 tag me please once done so that we merge this :)

salah-ghanim (changes requested):
see my previous comment

salah-ghanim:
@mateki0 branch is outdated, please rebase

mateki0:
@salah-ghanim done

salah-ghanim (approved):

PR #5061, opened 2024-05-20
fix(android): UIManagerModule fix for Bridgeless 0.74

arushikesarwani94:

No

Following APIs will be broken in Bridgeless mode after https://github.com/react-native-maps/react-native-maps/pull/4985 adds support for Bridgeless in React Native 0.74 on Android :
- takeSnapshot()
- getCamera()
- getAddressFromCoordinates()
- pointForCoordinate()
- coordinateForPoint()
- getMapBoundaries()

Since these [depend](https://github.com/react-native-maps/react-native-maps/blob/master/android/src/main/java/com/rnmaps/maps/MapModule.java#L93-L94) on `context.getNativeModule(UIManagerModule.class)`:

which returns null in Bridgeless

https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/runtime/ReactHostImpl.java#L600-L606

### How did you test this PR?

Test `takeSnapshot()` functionality for Android in Bridgeless

### DIFF
--- a/android/src/main/java/com/rnmaps/maps/MapModule.java
+++ b/android/src/main/java/com/rnmaps/maps/MapModule.java
@@ -19,9 +19,6 @@ import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.bridge.WritableNativeMap;
 import com.facebook.react.module.annotations.ReactModule;
-import com.facebook.react.uimanager.NativeViewHierarchyManager;
-import com.facebook.react.uimanager.UIBlock;
-import com.facebook.react.uimanager.UIManagerModule;
 import com.google.android.gms.maps.GoogleMap;
 import com.google.android.gms.maps.model.CameraPosition;
 import com.google.android.gms.maps.model.LatLng;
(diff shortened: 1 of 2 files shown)


arushikesarwani94:
cc: @lukmccall @salah-ghanim for review

salah-ghanim:
@Arushi thank you so much for your contribution, and good that you caught this crash.

but before merging can you please look into those two topics:

1. use of anonymous classes in java can cause overhead and ideally it gets avoided (mainly anonymous classes hold strong ref to parent class which can cause GC issues).


2. you can add the addUIBlock encapsulation inside such a class as well so that you don't need to repeat the code and anyone in the future can reuse it.

the code will look something like this 

```

public class MapUIBlock implements UIBlockInterface {
    private int tag;
    private Promise promise;
    private ReactApplicationContext context;
    private Function<MapView, Void> mapOperation;

    public MapUIBlock(int tag, Promise promise, ReactApplicationContext context, Function<MapView, Void> mapOperation) {
        this.tag = tag;
        this.promise = promise;
        this.context = context;
        this.mapOperation = mapOperation;
    }

    @Override
    public void execute(NativeViewHierarchyManager nvhm) {
        executeImpl(nvhm, null);
    }

    @Override
    public void execute(UIBlockViewResolver uiBlockViewResolver) {
        executeImpl(null, uiBlockViewResolver);
    }

    private void executeImpl(NativeViewHierarchyManager nvhm, UIBlockViewResolver uiBlockViewResolver) {
        MapView view = uiBlockViewResolver != null ? (MapView) uiBlockViewResolver.resolveView(tag) : (MapView) nvhm.resolveView(tag);
        if (view == null) {
            promise.reject("AirMapView not found");
            return;
        }
        if (view.map == null) {
            promise.reject("AirMapView.map is not valid");
            return;
        }

        mapOperation.apply(view);
    }

    public void addToUIManager() {
        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
            UIManager uiManager = UIManagerHelper.getUIManager(context, UIManagerType.FABRIC);
            ((FabricUIManager) uiManager).addUIBlock(this);
        } else {
            UIManagerModule uiManager = context.getNativeModule(UIManagerModule.class);
            uiManager.addUIBlock(this);
        }
    }
}

```

salah-ghanim (changes requested):
please review my last comment

iM-GeeKy:
@salah-ghanim Will this fix be included in the next release? I suppose I could use `patch-package` for the time being.

salah-ghanim (approved):

salah-ghanim:
@iM-GeeKy should be live in 10 minutes

Issue #4996, opened 2024-03-10
Marker's `onPress` doesn't return `nativeEvent.position` on iOS


melyux:
### Summary

The event from `onPress` for a Marker doesn't return the `position` field that's supposed to be inside `e.nativeEvent` on iOS. This is returned only on Android.

This bug is further evinced on [Stack Overflow](https://stackoverflow.com/questions/72015734/is-there-a-way-to-get-markers-on-screen-position-in-react-native-maps).

### Reproducible sample code

```TSX
<Marker
  ...
  onPress={(event) => {
   console.log(event.nativeEvent)
  }}
/>
```

### Steps to reproduce

1. Use this project on iOS.
2. Specify an `onPress` for a `Marker`
3. See that the event's `nativeEvent` does not return a position, only coordinates. (it should return both, per [the docs](https://github.com/react-native-maps/react-native-maps/blob/master/docs/marker.md))

### Expected result

`nativeEvent` should return a `position` value.

### Actual result

`nativeEvent` does not return a `position` value on iOS.

### React Native Maps Version

1.10.3

### What platforms are you seeing the problem on?

iOS (Google Maps)

### React Native Version

0.72.4

### What version of Expo are you using?

SDK 48

### Device(s)

iPhone 15 Po

mateki0:
Hi @melyux, thanks for reporting this. I think this is a mistake in the documentation because this marker uses GMSMarker underneath and his position property type is CLLocationCoordinate2D. This type has only latitude, longitude and animatableData.

Sources:
https://developer.apple.com/documentation/corelocation/cllocationcoordinate2d
https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_marker#property-documentation

abdullahbayram:
I have reproduced and encountered the same issue on an iPhone 15 simulator. Here are the details of my environment:

React Native Version: 0.71.14
Device: iPhone 15 simulator

The other fields in the bug report are the same as those in my environment.

abdullahbayram:
Still relevant

abdullahbayram:
`Still relevant`

abdullahbayram:
I added this lines to [AIRMapMarker.m](https://github.com/react-native-maps/react-native-maps/blob/da11dd647733846dd297a5a4df78bf0edfa0c7b8/ios/AirMaps/AIRMapMarker.m#L10) file and then `nativeEvent` returns `position` as well: 

```diff
[...]

PR #5092, opened 2024-06-16
fix(iOS): ensure onPress(Marker) returns nativeEvent.position #4996

abdullahbayram:

No.

#4996 

### How did you test this PR?

I have tested it on iPhone 15 Pro simulator(iOS 17.5) and iPhone 14 Pro(iOS 17.5.1).
(The issue occurs only iOS)

### Steps:
1. Specify an  onPress for a Marker on a iOS device.
2. Tap a marker and see the event.nativeEvent returns position as well.

In addition, this is a repository AirMapMarker.m file patched with patch-packager: 
You can also check it. [repro-on-marker-press](https://github.com/abdullahbayram/repro-on-marker-press)

### DIFF
--- a/ios/AirMaps/AIRMapMarker.m
+++ b/ios/AirMaps/AIRMapMarker.m
@@ -264,12 +264,17 @@ NSInteger const AIR_CALLOUT_OPEN_ZINDEX_BASELINE = 999;
     }
 
     // the actual marker got clicked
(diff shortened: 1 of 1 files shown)


perryfaro (approved):

salah-ghanim (approved):

melyux:
Thanks for solving this paşam. But there's a problem: on iOS, the returned field is now called `point` while it's called `position` in Android.

PR #5115, opened 2024-07-10
Implement dash options for geojson polygon

perryfaro:

No

Add missing props for polygon in the Geojson component. So I provided these props (lineDashPhase, lineDashPattern, lineCap, lineJoin and miterLimit) so an Geojson component with a polygon can also render a dashed line.

### How did you test this PR?

I have tested the change on iOS using an iPhone and in the simulator. For Android, this will not work because this is not implemented yet in Android.


vjanssens (approved):

salah-ghanim (approved):

salah-ghanim:
@perryfaro thanks for your contribution, merged :)

PR #5124, opened 2024-07-21
Update README.md

salah-ghanim:
Clarify that v1.4.0 and above require react-native 0.74 and above

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,10 @@ See [Installation Instructions](docs/installation.md).
 See [Setup Instructions for the Included Example Project](docs/examples-setup.md).
 
 ## Compatibility
+**v1.4.0** and above requires `react-native >= 0.74`.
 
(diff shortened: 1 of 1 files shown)


GETAWAY-Dev (approved):

alexkev:
@salah-ghanim Are you sure this right? It looks like you just approved your own PR too lol. I have been follow the some the issues tied to this and seems related fabric and that wasn't supported for that introduced much later than 1.4.0, I am guessing 1.14.0 because that is when the new architecture was added.

salah-ghanim:
@alexkev thanks for pointing the mistake out I guess, it would have been 10x better if you simply created a pull request yourself so that I can review it and merge.

this is an open source project and relies heavily on the public's contribution, by complaining instead of fixing it I have to again approve my own pull request ;)

PR #5196, opened 2024-10-19
fix(iOS): Refactor onPress(Marker) to return nativeEvent.position for consistency with Android

abdullahbayram:

No.

This PR is a follow-up to issue [#4996](https://github.com/react-native-maps/react-native-maps/issues/4996) 
In the initial fix, I mistakenly used `nativeEvent.point` instead of `nativeEvent.position`, which should have been used to ensure consistency with both the documentation and Android behavior.

### How did you test this PR?
I have tested it on:

- iPhone 13 Pro simulator (iOS 17.5)
- iPhone 14 Pro (iOS 17.5.1)

(The issue occurs only iOS)

### Steps
1. Specify an `onPress` for a Marker on a iOS device.
2. Tap a Marker and verify that `event.nativeEvent `returns the correct `position` instead of `point`

### DIFF
--- a/ios/AirMaps/AIRMapMarker.m
+++ b/ios/AirMaps/AIRMapMarker.m
@@ -272,7 +272,7 @@ NSInteger const AIR_CALLOUT_OPEN_ZINDEX_BASELINE = 999;
                          @"latitude": @(marker.coordinate.latitude),
                          @"longitude": @(marker.coordinate.longitude)
                          },
-                 @"point": @{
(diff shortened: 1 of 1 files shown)


Ayoubdesigner101:
can any one show this question https://stackoverflow.com/questions/79133414/invariant-violationnew-nativeeventemitter-requires-a-non-null-argument-js-e

salah-ghanim (approved):

PR #5227, opened 2024-11-22
fix(ios): color update in AIRGoogleMap

pinpong:
This PR fix the ios google maps marker color update:

https://snack.expo.dev/4kryDTGdDqr1zkOgnYq-g

PR #5360, opened 2025-02-18
fix(ios): removed unsupported fill color props

pinpong:
fixed polyline stroke color introduced with https://github.com/react-native-maps/react-native-maps/pull/5227


salah-ghanim:
thanks so much @pinpong for contributing

PR #5369, opened 2025-02-25
release fix

pinpong:
added src to shipped npm files

### DIFF
--- a/ios/AirGoogleMaps/RNMapsGoogleMapView.mm
+++ b/ios/AirGoogleMaps/RNMapsGoogleMapView.mm
@@ -9,9 +9,9 @@
 
 #import "RNMapsGoogleMapView.h"
 #import "RNMapsGooglePolygonView.h"
-#import "AirGoogleMap.h"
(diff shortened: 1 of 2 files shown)


salah-ghanim:
@pinpong srcs shouldn't be added to the npm, lib is the output of the ts compilation 

what are you trying to achieve here ?

pinpong:
check this comment, i think the specs need to be published as well

salah-ghanim:
@pinpong thanks for looking into this, yes I think it will need to be, but I believe there must be a way "to package it" / copy it to the lib folder, can be done via the push commit hook to do the copy or via the NPM publish github action

pinpong:
If you create a new package with `npx create-react-native-library@latest <Name of Your Library>`  this is also set within the package.json file.

salah-ghanim:
@pinpong sure but I'm sure it wouldn't lead to the same code being shipped twice, we can't simply ship the same code as ts and js in the same npm package, we need to make sure that specs folder is copied to the lib, that should be it

PR #5383, opened 2025-03-21
fix(codegen): Make LatLng codegen type as Float

yungblud:

I don't think so.

In Fabric (New Arch Mode), iOS's defaultProps coordinate lat, lng value is always 0.
So I changed codegen type for LatLng struct, it worked fine.

### How did you test this PR?

- checked working in my app


Stecken:
It makes sense, but will it affect android apps by some way? I said that because that I know LatLng model from Google uses double type, does codegen can be affected? It is just a question, I don't know to be honest.

[LatLng Google reference](https://developers.google.com/android/reference/com/google/android/gms/maps/model/LatLng)

yungblud:
> It makes sense, but will it affect android apps by some way? I said that because that I know LatLng model from Google uses double type, does codegen can be affected? It is just a question, I don't know to be honest.
> 
> [LatLng Google reference](https://developers.google.com/android/reference/com/google/android/gms/maps/model/LatLng)

Aha, I see. But I can say two things.

- Float type worked in android too.
- Maybe react-native's codegen works differently (from pure Java or Kotlin)

yungblud:
Cool. I will have a test and leave the feedback!

yungblud:
[!] Unable to find a specification for `react-native-maps-generated` depended upon by `react-native-maps`

You have either:
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.
 
 I am continuosly getting this error. Am I doing something wrong? (Pod install phase)
 Tested also with 1.21.0-alpha.126

yungblud:
Aha, I think I found the solution.

Just needed to add 

```ruby
pod 'react-native-maps-generated', :path => '../../../node_modules/react-native-maps'
```

in my Podfile.

I will have a test!

yungblud:
And also, I found this issue too.

As-is

`#import <react-native-maps-generated/ComponentDescriptors.h>`

To-be
`#import <ReactNativeMapsGenerated/ComponentDescriptors.h>`

because `react-native-maps-generated.podspec`'s module name is ReactNativeMapsGenerated

yungblud:
But after all, I ended up with this issue when build with xcode.

`'react/utils/fnv1a.h' file not found`

PR #5394, opened 2025-04-05
chore: update semantic-release

pinpong:
update semantic-release


salah-ghanim:
@pinpong can you please recreate this with master instead of alpha

PR #5414, opened 2025-04-16
chore: update semantic-release

pinpong:
update semantic-release


salah-ghanim:
@pinpong are those changes needed as well ? or the other pull request is sufficient ?

PR #5441, opened 2025-04-21
Feat/mono repo

salah-ghanim:

No

Autolinking issues with the example project due to the way react-native autolinking logic handles symlink 

### How did you test this PR?

locally so far, iOS worked


GETAWAY-Dev (approved):
LGTM

PR #5453, opened 2025-04-28
fix: re-add support for verbatimModuleSyntax

wojtekmaj:

No

v1.21.0 introduced a regression where projects using verbatimModuleSyntax option in TypeScript were throwing 200+ errors when typechecking (even with skipLibCheck enabled!). This fixes _all_ of these issues.

### How did you test this PR?

Enabled the option, ran typecheck.


iM-GeeKy:
When can we expect this change to land? Thanks @wojtekmaj 👍🏻

salah-ghanim (approved):

PR #5454, opened 2025-04-28
fix: re-add support for strictNullChecks

wojtekmaj:

No

v1.21.0 introduced a regression where projects using strictNullChecks option in TypeScript were throwing 12 errors when typechecking (even with skipLibCheck enabled!). This fixes _all_ of these issues.

### How did you test this PR?

Enabled the option, ran typecheck.


src/createFabricMap.tsx
```diff
@@ -41,213 +41,217 @@ export interface FabricMapHandle {
 }
 
 const createFabricMap = (ViewComponent: React.ComponentType, Commands: any) => {
-  return forwardRef<FabricMapHandle, FabricMapViewProps>((props, ref) => {
-    const fabricRef = useRef<React.ElementRef<typeof ViewComponent>>(null);
+  return forwardRef<FabricMapHandle | null, FabricMapViewProps>(
```
wojtekmaj, line 44: Scary diff head! But this `| null` bit is actually the only change, the rest is Prettier's work.

salah-ghanim (approved):

salah-ghanim:
thanks @wojtekmaj

Issue #5423, opened 2025-04-16
Typecheck fails in a repo pulling react-native-maps


VaniaPopovic:
### Summary

**Title:** TypeScript Errors in `react-native-maps/src` Despite `tsconfig.json` Exclusions and `skipLibCheck: true` 

Might be a duplicate of https://github.com/react-native-maps/react-native-maps/issues/5397 , but tsc is reporting different errors.
**Environment:**

*   `react-native-maps`: 0.77.2
*   `typescript`: 5.3.3
*   `expo`: 52.0.46
*   `node`: v22.14.0
*   OS: macOS (darwin 24.4.0)

**Problem Description:**

When running `tsc --noEmit` (via `npm run typecheck`) in our Expo project (monorepo), we encounter several TypeScript errors originating from files within the `react-native-maps/src` directory inside `node_modules`. This occurs even though we have configured `tsconfig.json` to exclude these files and have `skipLibCheck: true` enabled.

<img width="1420" alt="Image" src="https://github.com/user-attachments/assets/fc890c2e-72b7-45ea-b331-45d663784823" />

Example errors include:

*   `../../node_modules/react-native-maps/src/decorateMapComponent.ts:26:27 - error TS5097: An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.`
*   `../../node_modules/react-native-maps/src/decorateMapComponent.ts:129:38 - error TS2538: Type 'undefined' cannot be used as an index type.`
*   `../../node_modules/react-native-maps/src/Geojson.tsx:388:5 - error TS2322: Type 'LatLng[] | undefined' is not assignable to type 'LatLng[]'.`
*   `../../node_modules/react-native-maps/src/MapOverlay.tsx:82:27 - error TS2532: Object is possibly 'undefined'.`

**Attempts to Resolve:**

1.  **`skipLibCheck: true`:** Ensured `compilerOptions.skipLibCheck` is set to `true` in `tsconfig.json`.
2.  **Targeted Excludes:** Added specific paths to the problematic files in the `exclude` array:
    ```json
    "exclude": [
      "node_modules",
      "../../node_modules/react-native-maps/src/decorateMapComponent.ts",
      "../../node_modules/react-native-maps/src/Geojson.tsx",
      "../../node_modules/react-native-maps/src/MapOverlay.tsx"
    ]
    ```
    *Result:* Errors persisted.
[...]

PR #5465, opened 2025-05-02
fix: change import to not include a .ts file

wwdrew:

No, there's another PR that aims to fix other type issues but doesn't address this one.

Fixes the TypeScript errors:

```
../../node_modules/react-native-maps/src/decorateMapComponent.ts:26:27 - error TS5097: An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.
```

https://github.com/react-native-maps/react-native-maps/issues/5423

### How did you test this PR?

Updated the file and ran the TS compiler to confirm the error was gone.

This only affects the TypeScript compilation, it doesn't affect functionality.


salah-ghanim:
@wwdrew thank you so much for the contribution, I moved away from monorepo so had to create a similar pull request

Issue #5469, opened 2025-05-02
onRegionChange callback no longer executes on Android


trooperandz:
### Summary

The `onRegionChange` callback used to execute when the user would drag the map.  It is no longer firing for me at all on Android when dragging the map (have not tested iOS).

The other two handlers (`onRegionChangeStart()` and `onRegionChangeComplete()`) are still working for me, however.

### Reproducible sample code

```TSX
import React from 'react';
import {
  Dimensions,
  PermissionsAndroid,
  Platform,
  StatusBar,
  StyleSheet,
  View,
} from 'react-native';
import MapView, {PROVIDER_GOOGLE} from 'react-native-maps';

const screenDimensions = Dimensions.get('screen');

function App(): React.JSX.Element {
  React.useEffect(() => {
    async function requestLocationPermission(): Promise<boolean> {
      if (Platform.OS !== 'android') {
        return true; // iOS handled differently
      }

      try {
        const granted = await PermissionsAndroid.request(
          PermissionsAndroid.PERMISSIONS.ACCESS_FINE_LOCATION,
          {
            title: 'Location Access Required',
            message: 'This app needs access to your location.',
            buttonNeutral: 'Ask Me Later',
            buttonNegative: 'Cancel',
            buttonPositive: 'OK',
          },
        );

        return granted === PermissionsAndroid.RESULTS.GRANTED;
      } catch (err) {
        console.warn('Permission error:', err);
        return false;
      }
    }

    requestLocationPermission();
  }, []);

  return (
    <View style={styles.container}>
      <StatusBar />
      <View style={styles.mapContainer}>
        <MapView
          onRegionChangeStart={() => console.log('region change started...')}
          onRegionChange={() => console.log('region changed...')}
          onRegionChangeComplete={() =>
            console.log('region change completed...')
          }
          showsUserLocation={true}
          showsMyLocationButton={false}
          provider={PROVIDER_GOOGLE}
          style={styles.map}
          region={{
            latitude: 37.78825,
            longitude: -122.4324,
            latitudeDelta: 0.015,
            longitudeDelta: 0.0121,
          }}
        />
      </View>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    backgroundColor: 'white',
    flex: 1,
  },
  mapContainer: {
    ...StyleSheet.absoluteFillObject,
    height: screenDimensions.height,
    width: screenDimensions.width,
    justifyContent: 'flex-end',
    alignItems: 'center',
  },
[...]

Issue #5429, opened 2025-04-18
onRegionChange is never being called


DarlanSchwartz:
### Summary

This onRegionChange doesn't fires the event anymore, it used to work, after the fabric update it stopped, i only tried on android.

### Reproducible sample code

```TSX
import React from 'react';
import MapView, { Marker } from 'react-native-maps';
 export default function App(){
  const [coordinates, setCoordinates] = React.useState({
    latitude: 37.78825,
    longitude: -122.4324,
  });
  return (
    <MapView
        provider="google"
        style={{flex: 1}}
        onRegionChange={(region) => setCoordinates(region)}
        initialRegion={{
          latitude: 37.78825,
          longitude:  -122.4324,
          latitudeDelta: 0.0922,
          longitudeDelta: 0.0421,
        }}
      >
          <Marker 
            coordinate={coordinates}
            title="Marker Title"
            description="Marker Description"
          />
      </MapView>
  );
}
```

### Steps to reproduce

Create a blank project using command:

npx @react-native-community/cli@latest init AwesomeProject

cd AwesomeProject

npm install react-native-maps

place the example code on the app function

### Expected result

The event should be firing every single frame that the camera is moving.

### Actual result

The onRegionChange event is not firing at all, never. Only the onRegionChangeComplete is firing when it should be.

### React Native Maps Version

1.22.1

### What platforms are you seeing the problem on?

Android

### React Native Version

0.79.1

### What version of Expo are you using?

Not using Expo

### Device(s)

Android 15

DarlanSchwartz:
### Update 

**onRegionChange** works on iOS , seems like it's only a android problem.

ynejati:
Looks like a copy/paste mistake.  Firing region start event when its supposed to be firing a region change event. @salah-ghanim 

```diff
diff --git a/android/src/main/java/com/rnmaps/maps/MapView.java b/android/src/main/java/com/rnmaps/maps/MapView.java
index 0e69e8d..f913bf0 100644
--- a/android/src/main/java/com/rnmaps/maps/MapView.java
+++ b/android/src/main/java/com/rnmaps/maps/MapView.java
@@ -574,7 +574,7 @@ public class MapView extends com.google.android.gms.maps.MapView implements Goog
             cameraLastIdleBounds = null;
             boolean isGesture = GoogleMap.OnCameraMoveStartedListener.REASON_GESTURE == cameraMoveReason;
[...]

PR #5474, opened 2025-05-03
fix(android): onRegionChange not being dispatched

salah-ghanim:

No

https://github.com/react-native-maps/react-native-maps/issues/5469
https://github.com/react-native-maps/react-native-maps/issues/5429

### How did you test this PR?

Yes, android


AlixH:
@salah-ghanim we need the fix for v1.21.0 to support old arch please

Issue #5513, opened 2025-05-22
Marker presses on Android do not fire the correct event on the map view itself


fabiog27:
### Summary

When a marker press is detected on a marker (via the `onPress` prop on the `<Marker>` component), a press is also detected on the map component via `onPress`. On iOS, the corresponding event on the map component has the property `event.nativeEvent.action` set to `marker-press`. On Android, this is not the case anymore (having updated from 1.20.1 to 1.23.8), instead it's just a regular `press`.

Since my app relies on the marker press to select the marker and the map press to deselect them, this causes problems for me.

### Reproducible sample code

```TSX
import React from 'react';
import {View} from 'react-native';
import MapView, {MapPressEvent, Marker} from 'react-native-maps';

export default function App() {
  const onMapPress = (event: MapPressEvent) => {
    console.log('map pressed');
    // should be 'marker-press' when the marker was pressed, but is just 'press'
    console.log('event.nativeEvent.action:', event.nativeEvent.action);
  };

  const onMarkerPress = () => {
    console.log('Marker pressed');
  };

  return (
    <MapView
      style={{flex: 1}}
      initialRegion={{
        latitude: 42,
        longitude: 30,
        latitudeDelta: 1,
        longitudeDelta: 1,
      }}
      onPress={onMapPress}
    >
      <Marker
        coordinate={{latitude: 42, longitude: 30}}
        onPress={onMarkerPress}
      >
        <View style={{width: 10, height: 10, backgroundColor: 'red'}} />
      </Marker>
    </MapView>
  );
}
```

### Steps to reproduce

- Run the code above in a fresh React Native project
- Click the marker
- Observe the console

### Expected result

On map press outside marker: I expect it to log "map pressed" and "event.nativeEvent.action: press"

On marker press: I expect it to log "Marker pressed" and "map pressed" and "event.nativeEvent.action: marker-press"

### Actual result

On map press outside marker: correctly logs "map pressed" and "event.nativeEvent.action: press"

On marker press: logs "Marker pressed" and "map pressed" and "event.nativeEvent.action: press" instead of "marker-press"

### React Native Maps Version

1.23.8

### What platforms are you seeing the problem on?

Android

### React Native Version

0.78.0

### What version of Expo are you using?

Not using Expo

### Device(s)

Google Pixel 9 (Android 16)

salah-ghanim:
[...]

PR #5514, opened 2025-05-22
fix: correctly emit marker-press event for Android

fabiog27:

There is no other open PR doing the same thing.

https://github.com/react-native-maps/react-native-maps/issues/5513

### How did you test this PR?

I have created a patch for the package with this change in my own project and it fixed the issue I was encountering. The issue also links to a commit that was made for the rewrite for Fabric where these lines were apparently incorrectly changed.


salah-ghanim:
thanks @fabiog27

PR #5515, opened 2025-05-23
Update README.md

donni106:
Remove the part for `unstable_reactLegacyComponentNames` as it was removed with 1.19.1 see https://github.com/react-native-maps/react-native-maps/pull/5209#issuecomment-2498390067 and https://github.com/react-native-maps/react-native-maps/compare/v1.19.0...v1.19.1


salah-ghanim:
thanks @donni106 for the contribution

Issue #5492, opened 2025-05-07
TypeScript errors now being reported from command line check due to raw .tsx files type errors


trooperandz:
### Summary

I started getting TypeScript errors when running my checks via command line, regardless of exluding `node_modules` (see output below).  According to ChatGPT this is due to `react-native-maps` publishing raw `.tsx` and `.ts` files instead of publishing precompiled .js and .d.ts files (with type errors in them) like other libraries. I am not super familiar with publishing libraries so I am just quoting what I have read. According to this, when you run a TypeScript check from the command line, TypeScript now treats all `react-native-maps` imports as internal files instead of external modules; therefore when any of these contain type errors, there is no way to ignore them.

My errors list:

```
Errors  Files
     1  node_modules/react-native-maps/src/createFabricMap.tsx:46
     7  node_modules/react-native-maps/src/specs/NativeComponentGoogleMapView.ts:864
     7  node_modules/react-native-maps/src/specs/NativeComponentMapView.ts:1043
     6  node_modules/react-native-maps/src/specs/NativeComponentMarker.ts:296
```

Full error output:
```
node_modules/react-native-maps/src/createFabricMap.tsx:46:49 - error TS2344: Type 'ComponentType<{}>' does not satisfy the constraint 'ForwardRefExoticComponent<any> | (new (props: any) => Component<any, {}, any>) | ((props: any) => ReactNode) | keyof IntrinsicElements'.

46       const fabricRef = useRef<React.ElementRef<typeof ViewComponent>>(null);
                                                   ~~~~~~~~~~~~~~~~~~~~

node_modules/react-native-maps/src/specs/NativeComponentGoogleMapView.ts:864:31 - error TS2344: Type 'ComponentType<{}>' does not satisfy the constraint 'ForwardRefExoticComponent<any> | (new (props: any) => Component<any, {}, any>) | ((props: any) => ReactNode) | keyof IntrinsicElements'.

864     viewRef: React.ElementRef<React.ComponentType>,
                                  ~~~~~~~~~~~~~~~~~~~

node_modules/react-native-maps/src/specs/NativeComponentGoogleMapView.ts:870:31 - error TS2344: Type 'ComponentType<{}>' does not satisfy the constraint 'ForwardRefExoticComponent<any> | (new (props: any) => Component<any, {}, any>) | ((props: any) => ReactNode) | keyof IntrinsicElements'.

870     viewRef: React.ElementRef<React.ComponentType>,
[...]

PR #5524, opened 2025-05-27
fix: react ref element type deprecation fix

Blendfoul:

https://github.com/react-native-maps/react-native-maps/issues/5492

### How did you test this PR?

Screenshot 2025-05-27 at 10 28 07

The changes were only affecting typescript types so no functionality was altered.

### DIFF
--- a/src/createFabricMap.tsx
+++ b/src/createFabricMap.tsx
@@ -45,7 +45,7 @@ export interface FabricMapHandle {
 const createFabricMap = (ViewComponent: React.ComponentType, Commands: any) => {
   return forwardRef<FabricMapHandle | null, FabricMapViewProps>(
     (props, ref) => {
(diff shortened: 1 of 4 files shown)


Blendfoul:
@salah-ghanim I'm having issues with the codegen step on ios. I believe it has to do with the `includesGeneratedCode` flag on the package json but I'm not familiar with the steps to regenerate the native modules with the new specs can you please let me know how to do it?

Blendfoul:
Closing this pr as there's a strict check on codegen:

https://github.com/facebook/react-native/blob/main/packages/react-native-codegen/src/parsers/typescript/components/commands.js

that checks for the presence of the `React.ElementRef` this makes any solution for React 19 support which deprecates this interface not really possible...

Issue #5402, opened 2025-04-12
Performance drop on new architecture with custom view markers


metkm:
### Summary

In the older versions of `react-native-maps` (before 1.21), there were no performance drops with custom markers with `trackViewChanges` set to false in 1.21, there is a performance drop with custom markers even with `trackViewChanges` set to false.

### Reproducible sample code

```TSX
import { View } from 'react-native';
import MapView, { Marker } from 'react-native-maps';

export default function HomeScreen() {
  return (
    <MapView style={{ flex: 1 }}>
      {[...Array(50).keys()].map((item) => {
        return (
          <Marker
            key={item}
            tracksViewChanges={false}
            coordinate={{
              latitude: 38 + item / 4,
              longitude: 16
            }}
          >
            <View
              style={{
                width: 10,
                height: 10,
                backgroundColor: 'red'
              }}
            />
          </Marker>
        )
      })}
    </MapView>
  );
}
```

### Steps to reproduce

Render a lot of markers

### Expected result

For performance to not drop

### Actual result

Dragging on the map is low fps

### React Native Maps Version

1.21.0

### What platforms are you seeing the problem on?

Android

### React Native Version

0.76.9

### What version of Expo are you using?

SDK 52

### Device(s)

Oppo Reno 4 Lite

mr-tm:
Same here

mattproch:
Same here

salah-ghanim:
@metkm you need to use memorize on the JS side to avoid re-renders and should achieve better performance than before.

in your code you're recalculating the components tree each time render is called, and this leads to native changes.

please make sure your implementation keep the view tree unchanged for the map child components until you want something rendered / changed. 

chatgpt can help you here:
https://chatgpt.com/share/67fd36a4-6d50-800b-b789-e72b3adf824d

metkm:
@salah-ghanim 
I still have the same issue. Here's a recording of it with the code that chatgpt gave you

https://github.com/user-attachments/assets/9715f2d6-6967-4d8b-981e-6370a844d260

salah-ghanim:
@metkm please share a github / code repo with the example, happy to fix it in the next release

metkm:
@salah-ghanim Here you go
https://github.com/metkm/markers

karem-details:
Same .. Also the app crash on android

salah-ghanim:
@karem-details you will need to open a new ticket with the stack trace regarding the app crash.

[...]

PR #5545, opened 2025-06-07
fix(android): marker was constantly re-rendering

salah-ghanim:

No

https://github.com/react-native-maps/react-native-maps/issues/5402

### How did you test this PR?
Android emulator


GETAWAY-Dev (approved):

pinpong:
@salah-ghanim please revert this PR. In my case i have few markers that changed the size after selecting and the markers not updating anymore.

salah-ghanim:
@pinpong sorry didn't see your comment here, is it still an issue, if so please submit a bug report, the performance issue was needed to be solved.

PR #5575, opened 2025-06-15
CAShapeLayer Implementation for High-Performance Overlays

4ver:
## Summary
Replaces MKPolyline/MKPolygonRenderer/MKCircleRenderer with CAShapeLayer for polylines, polygons, and circles to achieve excellent FPS performance during real-time coordinate updates.

## Key Benefits
- **Excellent FPS**: No overlay recreation - only path updates
- **Smooth Movement**: Real-time updates during pan/zoom gestures  
- **Simple Implementation**: Direct Core Animation rendering
- **Full Compatibility**: All existing properties and features supported

## Technical Changes
- Uses single persistent CAShapeLayer per overlay instead of recreating MKOverlay objects
- Updates only the `path` property when coordinates change
- Automatic coordinate-to-screen conversion during map region changes

## Backwards Compatibility
✅ All existing polyline/polygon/circle properties work unchanged  
✅ Gradient colors, stroke width, patterns all supported  
✅ No API changes required

If accepted this replaces PR #5573 (which only prevents tearing) .


johnbindel:
This improves performance quite a bit for Polygons!

CostasCF:
can we reopen this?

PR #5576, opened 2025-06-17
Add typescript type info for showsTraffic property in MapView.tsx

StefanWallin:

Not that I could find.

Typescript type information for `showsTraffic` property on `MapView`.

### How did you test this PR?
I've not tested this, but I noticed a discrepancy between the documentation for the `<MapView/>`-api (https://github.com/react-native-maps/react-native-maps/blob/5873c0cd8aaac04e08cf29bf30762cadc54379d6/docs/mapview.md?plain=1#L31) and the actual `MapViewProps` in typescript.

I'm unsure if the properties get passed somehow to lower level components or api's, but the documentation states that there is a property however it does not exist in the typescript type definition. This PR adds that property info, perhaps it should instead be removed from documentation?

### DIFF
--- a/src/MapView.tsx
+++ b/src/MapView.tsx
@@ -597,6 +597,15 @@ export type MapViewProps = ViewProps & {
   showsScale?: boolean;
 
   /**
+   * A Boolean value indicating whether the map displays traffic information. 
+   *
+   * @default false
+   * @platform iOS: Supported
+   * @platform Android: Supported
+   */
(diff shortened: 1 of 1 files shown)


pinpong:
@salah-ghanim please merge this.

salah-ghanim (approved):

avi312singh:
This doesnt work anymore in latest version

Issue #5574, opened 2025-06-14
Request: Add support for controlling Points of Interest visibility and consistent map styling across platforms


shridharkaushik29:
### Summary

Request to add support for controlling Points of Interest visibility and consistent map styling across platforms, as these features are currently missing or inconsistent between Google Maps and Apple Maps.

### Reproducible sample code

```TSX
import MapView from 'react-native-maps';

// Attempting to hide POIs (not supported)
<MapView
  showsPointsOfInterest={false}  // This prop doesn't exist
  style={{ flex: 1 }}
/>

// Attempting to style Apple Maps (only works on Google Maps)
<MapView
  customMapStyle={mapStyle}  // Only affects Google Maps
  style={{ flex: 1 }}
/>
```

### Steps to reproduce

1. Install react-native-maps version 1.24.3
2. Try to use `showsPointsOfInterest` prop to hide POIs
3. Try to use `customMapStyle` on Apple Maps
4. Observe that neither feature works as expected

### Expected result

- `showsPointsOfInterest` prop should be available to control POI visibility on both platforms
- `customMapStyle` should work consistently across both Google Maps and Apple Maps, or there should be an equivalent prop for Apple Maps

### Actual result

- No way to control POI visibility on either platform
- `customMapStyle` only affects Google Maps, with no equivalent functionality for Apple Maps

### React Native Maps Version

1.24.3

### What platforms are you seeing the problem on?

iOS (Apple Maps)

### React Native Version

0.79.3

### What version of Expo are you using?

SDK 53

### Device(s)

any iPhone

### Additional information

This feature is available in native map implementations:
- iOS: `MKMapView` has a `showsPointsOfInterest` property
- Android: Google Maps has similar functionality through `setMapToolbarEnabled`

For map styling:
- Google Maps: Currently supports `customMapStyle`
- Apple Maps: No equivalent functionality available

Current workarounds:
1. Accept that POIs cannot be hidden
2. Use different styling approaches for different platforms
3. Potentially use platform-specific code to handle these differences

kstolk:
I was looking for the option to hide POI on my google maps.. not possible yet.
showsPointsOfInterest is apple maps only

vpcano:
> I was looking for the option to hide POI on my google maps.. not possible yet. showsPointsOfInterest is apple maps only

On Apple Maps doesn't work either, at least on the latest version. The prop `showsPointsOfInterest` just doesn't exist.

27leaves:
[...]

PR #5597, opened 2025-06-27
fix(ios): restore showsPointsOfInterest prop

vpcano:

No

This PR restores the functionality of the `showsPointsOfInterest` prop on iOS, which was deleted a few versions before because of it using a [deprecated property](https://developer.apple.com/documentation/mapkit/mkmapview/showspointsofinterest?language=objc) of `MKMapView`. This now uses the new [`MKPointOfInterestFilter`](https://developer.apple.com/documentation/mapkit/mkpointofinterestfilter?language=objc) pattern for disabling points of interest.

I believe this solves #5574.

### How did you test this PR?

This change only affects iOS platform and Apple Maps. The default behaviour remains unchanged.
To test the prop, I created a **new example** called **`PointsOfInterestToggle`**. Tested on iOS simulator.

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -2,7 +2,7 @@ PODS:
   - boost (1.84.0)
   - DoubleConversion (1.1.6)
   - fast_float (6.1.4)
-  - FBLazyVector (0.79.1)
+  - FBLazyVector (0.79.3)
   - fmt (11.0.2)
   - glog (0.3.5)
   - Google-Maps-iOS-Utils (6.1.0):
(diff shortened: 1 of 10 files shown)


salah-ghanim:
@vpcano thank you so much for the pull request, quick feedback:
 1. please remove the example ... it's not necessary.
 2. remove the support for iOS 3 or less, as you mentioned in the code we don't support lower iOS v than iOS 15.
 3. it seems like PoI filter can show / hide a lot of categories, not only show / hide, I wonder if it's worth it to actually give control to the user / add API to specify which category to include / exclude.
 4. I see that you say showPoI as true by default yet I see you proactively setting it to true, I dislike that this behaviour doesn't match the native iOS behaviour, default means what you should expect from native, so I would recommend you make it false .

vpcano:
> 3. it seems like PoI filter can show / hide a lot of categories, not only show / hide, I wonder if it's worth it to actually give control to the user / add API to specify which category to include / exclude.

Could be a nice feature. I'm thinking of defining a type with [all the possible POI categories](https://developer.apple.com/documentation/mapkit/mkpointofinterestcategory?language=objc) and add a new prop `poiCategories` that can be assigned a list of them. Nevertheless, I believe that the possibility to either include or exclude all the categories directly should remain, by only setting the existing`showsPointsOfInterest` prop.

> 4. I see that you say showPoI as true by default yet I see you proactively setting it to true, I dislike that this behaviour doesn't match the native iOS behaviour, default means what you should expect from native, so I would recommend you make it false .

I'm setting it to true on line 79 of `AIRMap.mm`. I haven't found what the default value of `MKStandardMapConfiguration.pointOfInterestFilter` is, but I could initialise this and only change it when the props are defined.

Let me know what you think about both things.

PR #5627, opened 2025-07-19
chore(android): remove deprecated buildDrawingCache call

pinpong:
See: https://developer.android.com/reference/android/view/View.html#buildDrawingCache()


pinpong:
closed in favour of: https://www.npmjs.com/package/react-native-google-maps-plus

PR #5635, opened 2025-07-21
fix(android): location updates

pinpong:
No

Changing `userLocationUpdateInterval` | `userLocationFastestInterval` | `userLocationPriority` has no effect until `activate()` is called. This happens only if map is calling, for example the app goes into background. Therefore , `deactivate()` and `activate()` need to be called again whenever `locationRequest` changes.
        See: `fusedLocationClientProviderClient.requestLocationUpdates(locationRequest, locationCallback, Looper.myLooper());`

Additionally, i implemented the new `LocationRequest.Builder` SDK version for  >= 31

### How did you test this PR?
Tested on Android emulator with location route

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,40 @@
 # Changelog
 
+## [1.24.13](https://github.com/react-native-maps/react-native-maps/compare/v1.24.12...v1.24.13) (2025-07-26)
+
+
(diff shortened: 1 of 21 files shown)


pinpong:
closed in favour of: https://www.npmjs.com/package/react-native-google-maps-plus

PR #5636, opened 2025-07-21
fix(android): heatmap radius not within bounds

pinpong:
No

prevent crash because heatmap radius must be in range of 10 -50

### How did you test this PR?

Tested with the example app on android emulator


salah-ghanim:
@pinpong thanks ! but I will delay merging this, I don't mind severe unexpected behaviour if wrong values are passed by the user (developer) 
since the docs clearly say 10-50 is the range, so passing values outside the range can cause a crash and that's ok
(google didn't block it from crashing, why should we ?)
https://github.com/react-native-maps/react-native-maps/blob/eebf1b411c241d334cb43ac3a269c4e349db236c/docs/heatmap.md

pinpong:
you are right

Issue #5645, opened 2025-07-24
onRegionChangeComplete() not executing on Android on initial map load


trooperandz:
### Summary

`onRegionChangeComplete()` is not executing on Android after an `onRegionChange()` event fires on the initial map load.  If a drag event occurs, it does execute at that point. On iOS, however, `onRegionChangeComplete()` executes after the initial map load (without any user drag event).  

So this is an inconsistency between the two platforms.  We are relying on the `onRegionChangeComplete()` to execute on the initial load so that a map pin will animate down to its starting position (which typically happens after `onRegionChange()` executes).  Subsequent user map drag events properly move the pin down because `onRegionChangeComplete()` is firing after the user drags the map.

One thing to note in the test example below:  on Android, `onRegionChange()` fires, and then `onRegionChangeComplete()` never fires. On iOS, however, `onRegionChange()` is never logged initially, but `onRegionChangeComplete()` is.  This is different from our production app, where both platforms initially fire `onRegionChange()`, but _only_ iOS then fires a subsequent `onRegionChangeComplete()`. 

### Reproducible sample code

```TSX
import React from 'react';
import {Dimensions, StatusBar, StyleSheet, View} from 'react-native';
import MapView, {
  MapMarker,
  Marker,
  Polygon,
  PROVIDER_GOOGLE,
} from 'react-native-maps';

const screenDimensions = Dimensions.get('screen');

function App(): React.JSX.Element {
  const markerRef = React.useRef<MapMarker>(null);

  return (
    <View style={styles.container}>
      <StatusBar />
      <View style={styles.mapContainer}>
        <MapView
          initialRegion={{
            latitude: 37.78825,
            longitude: -122.4324,
            latitudeDelta: 0.015,
            longitudeDelta: 0.0121,
          }}
          onRegionChange={() => {
            console.log('onRegionChange...');
          }}
          onRegionChangeComplete={() => {
            console.log('onRegionChangeComplete...');
          }}
          showsUserLocation={true}
          showsMyLocationButton={false}
          userInterfaceStyle="light"
          provider={PROVIDER_GOOGLE}
          style={styles.map}>
          <Marker
            coordinate={{latitude: 37.78825, longitude: -122.4324}}
            title="My Marker"
            description="This is a marker example"
            ref={markerRef}
          />
          <Polygon
            coordinates={[
              {latitude: 37.78825, longitude: -122.4324},
              {latitude: 37.78925, longitude: -122.4334},
[...]

PR #5648, opened 2025-07-26
fix(android): onRegionChangeComplete on initial map loaded

pinpong:
No

#5645

### How did you test this PR?
tested on android emulator


salah-ghanim:
thanks @pinpong

Issue #5497, opened 2025-05-11
Android Crashes


mattproch:
### Summary

Hey, on the initial mount, android crashes. It starts in the 1.21.0 version and it is still present in the 1.23.7 version.

I get this error:

![Image](https://github.com/user-attachments/assets/70a12de1-5705-4c35-89b5-b7236d3b5613)

### Reproducible sample code

```TSX
<MapView
        ref={map}
        superClusterRef={superCluster}
        initialRegion={
          location.latitude && location.longitude
            ? {
                latitude: location.latitude,
                longitude: location.longitude,
                latitudeDelta: DEFAULT_LAT_DELTA,
                longitudeDelta: DEFAULT_LNG_DELTA,
              }
            : appConfig.MAP.INITIAL_REGION
        }
        toolbarEnabled={false}
        animationEnabled={false}
        showsUserLocation={!!permissions.locationGranted}
        showsMyLocationButton={false}
        style={styles.map}
        radius={40}
        onPress={onMapPress}
        onPanDrag={onMapDrag}
        renderCluster={renderCluster}
        googleRenderer="LATEST"
        {...otherProps}
      >
        {markers}
      </MapView>


interface Props {
  coordinate: { latitude: number; longitude: number };
  onMarkerPress?: (markerRef: MapMarker | null) => void;
}

const Marker = ({ coordinate, onMarkerPress }: Props) => {
  const ref = useRef<MapMarker>(null);

  return (
    <DefaultMarker
      ref={ref}
      centerOffset={{ x: 0.0, y: -26 }}
      coordinate={coordinate}
      tracksViewChanges={false}
      image={{ uri: appConfig.MAP.MARKER_IMAGE }}
      style={styles.marker}
      pinColor={COLORS['solid/blue']}
      zIndex={1}
      onPress={() => {
        if (onMarkerPress) {
          onMarkerPress(ref.current);
        }
      }}
    ></DefaultMarker>
  );
};

export default memo(Marker);

const styles = StyleSheet.create({
  marker: {
    width: 43,
    height: 52,
  },
});
```

### Steps to reproduce

I also use "react-native-map-clustering" version 3.4.2.

### Expected result

App not crashing

### Actual result

App crashes

### React Native Maps Version

1.23.7

### What platforms are you seeing the problem on?

Android

### React Native Version

0.76.9

### What version of Expo are you using?

SDK 52

### Device(s)

Honor, Android 9

salah-ghanim:
[...]

PR #5662, opened 2025-08-02
fix: crash android react 19 parallel features

salah-ghanim:

No
What issue is this PR fixing?

https://github.com/react-native-maps/react-native-maps/issues/5497
How did you test this PR?

Android emulator, the bug was reproducible here:
https://github.com/ammar-madni/repro-index-out-of-bounds/


flmalte:
Unfortunately this version creates a bug where under certain ocasions the custom markers also render a normal marker:
Screenshot_1754313479

Issue #5578, opened 2025-06-17
NPE: Android crashes when app goes to background mode (regression after 1.24.3 similar problem)


isinuyk:
### Summary

Hey, I have a react native CLI app that is using the latest react-native-maps version. Previously, I was experiencing a crash when moving the app to the background after going back to the previous stack screen. [Link](https://github.com/react-native-maps/react-native-maps/issues/5449). That issue has been resolved, but a new one has appeared. 
If I open a screen with a map and then immideately move app to the background I will get a crash.

### Reproducible sample code

```TSX
<MapView
          onPress={onPress}
          onMapLoaded={() => {
            if (!renderRef.current) {
              fitToCoordinates(!isMiddleTab);
              renderRef.current = true;
            }
          }}
          ref={mapRef}
          provider={PROVIDER_GOOGLE}
          region={initialLocation}
          showsScale
          style={[styles.map, style]}/>
```

### Steps to reproduce

1. You need to have 2 screens in your stack
2. Go from screen A to screen B ( where the map is located)
3. Immediately move your app to the background

### Expected result

Do not have a crash

### Actual result

App crashes with

NullPointerException
Attempt to invoke virtual method 'boolean java.util.AbstractCollection.isEmpty()' on a null object reference
com.google.android.gms.dynamic.DeferredLifecycleHelper in zae
com.google.android.gms.dynamic.DeferredLifecycleHelper in onDestroy
com.google.android.gms.dynamic.DeferredLifecycleHelper in onDestroy at line 2
com.google.android.gms.maps.MapView in onDestroy
com.rnmaps.fabric.MapViewManager in onDropViewInstance at line 621
com.rnmaps.fabric.MapViewManager in onDropViewInstance at line 43
com.facebook.react.fabric.mounting.SurfaceMountingManager in onViewStateDeleted
com.facebook.react.fabric.mounting.SurfaceMountingManager in onViewStateDeleted at line 1022
com.facebook.react.fabric.mounting.SurfaceMountingManager in lambda$stopSurface$1
com.facebook.react.fabric.mounting.SurfaceMountingManager in lambda$stopSurface$1 at line 313
com.facebook.react.fabric.mounting.SurfaceMountingManager in $r8$lambda$PLPI0kX_cXj3vbfkGy4qh_tXRXA
com.facebook.react.fabric.mounting.SurfaceMountingManager$$ExternalSyntheticLambda1 in run
android.os.Handler in handleCallback at line 959
[...]

Issue #5658, opened 2025-08-01
Crash on Android with Fabric if map is on hidden tab


andreialecu:
### Summary

We have a tabview with a map that is displayed in `Tab #2`, so initially it is not rendered. When going back from this screen we're seeing a native crash:

```
Attempt to invoke virtual method 'boolean java.util.LinkedList.isEmpty()' on a null object reference
com.google.android.gms.dynamic.DeferredLifecycleHelper in zae at line 1
com.google.android.gms.dynamic.DeferredLifecycleHelper in onDestroy at line 2
com.google.android.gms.maps.MapView in onDestroy at line 1
com.rnmaps.fabric.MapViewManager in onDropViewInstance at line 621
com.rnmaps.fabric.MapViewManager in onDropViewInstance at line 43
```

Setting the tabview to render screens with `lazy: true` works around it.

Also, if the user manually goes to the second tab and then back, the crash is prevented.

It appears that not rendering the map in the first place causes this crash.

### Reproducible sample code

```TSX
<Tab.Navigator>
  <Tab.Screen>{(tabProps) => <View />}<Tab.Screen>
  <Tab.Screen>{(tabProps) => <MapView />}<Tab.Screen>
</Tab.Navigator>
```

### Steps to reproduce

Using react navigation, navigate to a screen that has the mapview on the second tab, and go back without activating the second tab.

### Expected result

No crash.

### Actual result

Native crash.

### React Native Maps Version

1.24.13

### What platforms are you seeing the problem on?

Android

### React Native Version

0.79.4

### What version of Expo are you using?

Not using Expo

### Device(s)

Any device, including simulators

jamesholcomb:
seeing this as well, 1.24.12 react native 0.79.x android new architecture

<img width="571" height="1206" alt="Image" src="https://github.com/user-attachments/assets/cf6734b0-1afe-4875-a78d-79721d10fba7" />

chernandezDA:
I have the same configuration and the same issue, someone can help us?

salah-ghanim:
please submit a sample app to reproduce the bug

salah-ghanim:
I just created this pull request https://github.com/react-native-maps/react-native-maps/pull/5664
and will merge it assuming it will solve the problem, from what I see, we already have doDestroy native method on MpaView that manually detaches lifecycleObservers and pause MapView proactively before calling onDestroy.

will merge it and include it in the next release, would be nice if anyone can test because I can't reproduce the bug.

[...]

PR #5664, opened 2025-08-03
fix(android): fix NPE crash on destroy

salah-ghanim:

No

https://github.com/react-native-maps/react-native-maps/issues/5578

### How did you test this PR?
couldn't reproduce the crash but the logic is consistent, naked use of onDestroy causes NPE while use of the safe method doDestroy (checks for observers) solved other NPE caused by lifecycle before.


HugoGresse:
Hey, I can confirm this PR does not solve the NPE see https://github.com/react-native-maps/react-native-maps/issues/5658#issuecomment-3168559648

HugoGresse:
but PR https://github.com/react-native-maps/react-native-maps/pull/5666 does

Issue #5595, opened 2025-06-25
Map region lost on navigation (Android only)


VadymVashchuk:
### Summary

I am using @react-navigation/native-stack, react-native 0.79 (new Architecture) and noticed an issue with new versions of this package. 
When I navigate from screen with map to another screen and go back, map loses it's position. And it is not even rerendered. Only android.

### Reproducible sample code

```TSX
const MainScreen = () => {
  return (
    <View style={{flex: 1}}>
      <MapView style={{flex: 1}} provider={'google'} />
      <Pressable onPress={() => navigationRef.navigate('Profile'} style={{position: 'absolute', top: 50, left: 50, backgroundColor: 'red', width: 50, height: 50}}></Pressable>
    </View>
  );
};

export default MainScreen;
```

### Steps to reproduce

1. MainScreen has <MapView />. It is as simple as it could be.
2. We navigate to another screen with navigation.navigate('screenB') - common pattern.
3. We go back go MainScreen (no matter how, swiper or navigation.goBack())
4. Map loses it's position and shows default position if Africa with max zoom level I suppose.

### Expected result

Like in IOS map should not change it's region.

### Actual result

Map changes position although it's component is not rerendered.

### React Native Maps Version

1.24.3, maybe earlier

### What platforms are you seeing the problem on?

Android

### React Native Version

0.79.1 (new Architecture)

### What version of Expo are you using?

Not using Expo

### Device(s)

Samsung A05

markwilcox:
We also have this issue. Additionally, trying to `animateToRegion` via the ref fails after navigating away and back because the ref no longer points to the native view. Seems like the teardown and restore of the native view doesn't work properly in the new architecture on Android.

I'd love to try to help fix it, but I'm not sure I understand how it's supposed to work with the new architecture.

noe4400:
What worked for me was downgrading to version 1.24.2.

CongCaoDEV:
I also had the app closed for no reason.

markwilcox:
> What worked for me was downgrading to version 1.24.2.

Downgrading to 1.24.2 does indeed resolve this, but the change from 1.24.2 to 1.24.3 is a fix for an Android crash bug, so you might not want that if you get the crash.

kstolk:
Same issue here on android only. Navigating to a location does not work anymore since this version.

LukasB-DEV:
have the same issue @salah-ghanim can you take a look into this?

Bilalget:
[...]

PR #5665, opened 2025-08-03
feat(android): support saving view state

salah-ghanim:

No

https://github.com/react-native-maps/react-native-maps/issues/5595
and other similar issues reported

### How did you test this PR?

Android emulator

### DIFF
--- a/android/src/main/java/com/rnmaps/maps/MapView.java
+++ b/android/src/main/java/com/rnmaps/maps/MapView.java
@@ -13,6 +13,7 @@ import android.location.Location;
 import androidx.lifecycle.DefaultLifecycleObserver;
 import androidx.lifecycle.LifecycleOwner;
 
+import android.os.Bundle;
 import android.view.GestureDetector;
 import android.view.MotionEvent;
 import android.view.View;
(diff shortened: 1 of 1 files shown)


GETAWAY-Dev (approved):

pennersr:
@salah-ghanim I just tested 1.25.0 -- it indeed remembers the region now, thanks. But, the map type (`mapType=...`) is still lost.

GETAWAY-Dev:
@pennersr I just tested the mapType prop and it's working as expected... you might have issues with cached code, from what I see it works as expected in a clean example app

TheSloanRanger:
On 1.25.0 when navigating to a different screen in the same stack, and then navigating back to the screen with the MapView, it removes the blue user dot and reverts to initially given region (regardless of user dragging the map). Can manually update the region using onRegionChange but never had to do this in the past.

bigcupcoffee:
This isn't really a good fix, Map still acts horribly with navigation, you can visibly see Map having a reset as if you supplied `key={Math.random()` to it. Especially visible with drawer navigation, if you open or close the drawer Map "unmounts" and then restores all before user's eyes

PR #5725, opened 2025-09-05
fix: update versions for new/old arch support in README

tranjog:

No

The README file lists 1.21.1 as the last supported version for the old architecture.
1.21.0 is the first version with Fabric support (and where the old architecture is no longer supported).
1.20.1 is the last version where the old architecture is supported

### How did you test this PR?

N/A


salah-ghanim:
@tranjog thanks for the support !

PR #5727, opened 2025-09-07
fix(android): Prevent duplicate initialization during MapView state r…

zhuscat:
…estore

No

This PR fixes two critical state restoration issues in Android MapView that occur during view detach/attach cycles:

1. **Duplicate initialization of `initialRegion`/`initialCamera`**: When a MapView with `initialRegion` or `initialCamera` props gets detached and reattached (e.g., during navigation or screen changes), the map would incorrectly reset to the initial position, ignoring any programmatic changes made via `animateToRegion()` or similar methods.

2. **Lost map padding state**: When `mapPadding` props are set and the MapView undergoes detach/attach, the padding configuration would be lost and not properly restored.

### How did you test this PR?

**Platform**: Android only  
**Testing method**: Real device testing using a production application

**Test setup**:
1. Modified the react-native-maps library locally with the proposed fixes
2. Updated the production app's dependency to point to the modified version
3. Performed comprehensive testing on physical Android devices

**Test scenarios**:
1. **Initial position persistence test**:
   - Set `initialRegion` or `initialCamera` on MapView
   - Use `animateToRegion()` to change the map position programmatically
   - Trigger view detach/attach by navigating away and back to the screen
   - ✅ **Expected**: Map maintains the programmatically set position
   - ✅ **Result**: Map no longer resets to initial position after the fix

2. **Map padding restoration test**:
   - Configure MapView with `mapPadding` props
   - Trigger view detach/attach cycle
   - ✅ **Expected**: Map padding is preserved after reattachment
   - ✅ **Result**: Padding state is now correctly maintained

**Technical changes**:
- Added proper state tracking with `initialRegionSet` and `initialCameraSet` flags to prevent duplicate initialization
- Introduced `shouldRestorePadding` flag to manage padding restoration during state recovery
- Modified `applyBridgedProps()` to respect the initialization state and handle padding restoration correctly


salah-ghanim (approved):

salah-ghanim:
@zhuscat thank you so much the code worked perfectly !

Issue #5730, opened 2025-09-12
[iOS] Crash when region is set with span out of specific range.


LukaszPenczak:
### Summary

When calling `setRegion` of `MKMapView` the data provided has to be valid. The region is a combination of center position and span. The current codebase is validating only the center but not the span. The span is expected to be in a specific range:
latitude [0, 180]
longitude [0, 360]
If it's outside of this range the app will crash.

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Invalid Region <center:+0.00000000, +0.00000000 span:+181.00000000, +361.00000000>'
```

### Reproducible sample code

```TSX
import React from 'react';
import MapView from 'react-native-maps';

export default function App() {
  return (
    <MapView
      style={{flex: 1}}
      initialRegion={{
        latitude: 0,
        longitude: 0,
        latitudeDelta: 181,
        longitudeDelta: 361,
      }}
    />
  );
}
```

### Steps to reproduce

Just run the app.

### Expected result

Just don't crash and show the map correcting the span to the valid value. For negative values set 0 and for the values exceeding the expectation just trim them to the maximum ones.

### Actual result

Crashing

### React Native Maps Version

1.24.3

### What platforms are you seeing the problem on?

iOS (Apple Maps)

### React Native Version

0.78.3

### What version of Expo are you using?

Not using Expo

### Device(s)

all apple devices

salah-ghanim:
@LukaszPenczak thanks for reporting this and creating the pull request.

but unfortunately I have to reject this. 

1. please share the stack trace next time so that it's easier to analyze the bug figure out exactly what happened.
2. I don't think this library should introduce behaviour not matching native as much as possible, it should mirror as much as possible the native components, without being opinionated. 

in this case MKMapView the native MapView by apple is crashing if you pass such region (you can test it also in swift) ... so I don't see why we should change that behaviour, the developer should be careful with what they pass, if something crashes due to the lib (conversion, nullability ... we have to fix it) but if it crashes because that's what the native components will do for that input then we should keep it transparent and deliver the same result to the developer.

LukaszPenczak:
@salah-ghanim I thought that's the idea of react native is to have the same code on all platforms. What's the benefit of crashing the app instead of correcting invalid values. You wrote developer should be careful whit what they pass, but the same values are valid on android and when google maps are used.  I guess they internally correct the value because span greater than 360 doesn't make any sense. Same for the negative values.

[...]

PR #5731, opened 2025-09-12
fix: Fixes crash when region is out of range.

LukaszPenczak:

Nope.

https://github.com/react-native-maps/react-native-maps/issues/5730

### How did you test this PR?

Simply run the example app provided in the bug report.


salah-ghanim:
@LukaszPenczak please see my comment on the issue #5730

Issue #5736, opened 2025-09-15
Android: Markers not disappearing when removed from state


brunobock-del:
### Summary

Markers are not disappearing when removed from the JSX tree via state changes.
Even after toggling the state to hide specific groups of markers, they remain rendered on the map.

On Android, logcat shows errors like:
```bash
removeViewAt: [4260] -> [1530] @1288: view already removed from parent! Children in parent: 1
```

### Reproducible sample code

```TSX
import React, {useState} from 'react';
import {View, TouchableOpacity, Image, Keyboard} from 'react-native';
import MapView, {Marker} from 'react-native-maps';

export default function MapScreen() {
  const [hideGroupA, setHideGroupA] = useState(false);
  const [hideGroupB, setHideGroupB] = useState(false);

  const [filteredGroupA, setFilteredGroupA] = useState([]);
  const [filteredGroupB, setFilteredGroupB] = useState([]);

  const allStationsGroupA = [
    {id: 'a1', lat: -23.5, lng: -46.6, label: 'Station A1', owner: 'Owner A'},
    {id: 'a2', lat: -23.52, lng: -46.62, label: 'Station A2', owner: 'Owner A'},
    {id: 'a3', lat: -23.48, lng: -46.58, label: 'Station A3', owner: 'Owner A'},
    {id: 'a4', lat: -23.55, lng: -46.65, label: 'Station A4', owner: 'Owner A'},
    {id: 'a5', lat: -23.6, lng: -46.63, label: 'Station A5', owner: 'Owner A'},
    {id: 'a6', lat: -23.45, lng: -46.55, label: 'Station A6', owner: 'Owner A'},
    {id: 'a7', lat: -23.58, lng: -46.7, label: 'Station A7', owner: 'Owner A'},
    {id: 'a8', lat: -23.62, lng: -46.72, label: 'Station A8', owner: 'Owner A'},
    {id: 'a9', lat: -23.49, lng: -46.68, label: 'Station A9', owner: 'Owner A'},
    {id: 'a10', lat: -23.53, lng: -46.74, label: 'Station A10', owner: 'Owner A'},
  ];

  const allStationsGroupB = [
    {id: 'b1', lat: -23.6, lng: -46.7, label: 'Station B1', owner: 'Owner B'},
    {id: 'b2', lat: -23.62, lng: -46.72, label: 'Station B2', owner: 'Owner B'},
[...]

PR #5752, opened 2025-09-29
fix(android): fix remove marker from map issues

salah-ghanim:

No

https://github.com/react-native-maps/react-native-maps/issues/5736

### How did you test this PR?
Didn't test it, but looks clear

### DIFF
--- a/android/src/main/java/com/rnmaps/fabric/MarkerManager.java
+++ b/android/src/main/java/com/rnmaps/fabric/MarkerManager.java
@@ -153,6 +153,12 @@ public class MarkerManager extends ViewGroupManager<MapMarker> implements RNMaps
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


Jhamnerx:
Should this fix the issue when using cluster?
Why do the markers still appear when zooming in and out after interacting with the zoom?

image

lorenc-tomasz:
@Jhamnerx Could you manually apply the changes and test it in your case, please? It's the easiest way to check if it works or not :)

GETAWAY-Dev (approved):

salah-ghanim:
merging because I got testing feedback it shows clear improvements in our apps.

NabeelZanabeet:
Here is proper issue reporting for the red markers with repo to reproduce the issue easily 
https://github.com/react-native-maps/react-native-maps/issues/5756

Issue #5750, opened 2025-09-29
Markers/Polyline disappear after navigating back to MapView screen


Ikthegreat:
### Summary

When navigating away from a screen containing a MapView (using React Navigation stack push) and then returning with pop, the Marker and Polyline overlays disappear. The map itself still renders, but the overlays are gone.

### Reproducible sample code

```TSX
import { useNavigation } from '@react-navigation/native';
import { Button, View } from 'react-native';
import MapView, { Marker, Polyline, PROVIDER_GOOGLE } from 'react-native-maps';

const MapScreen = () => {
  const navigation = useNavigation();

  return (
    <View style={{ flex: 1 }}>
      <MapView
        provider={PROVIDER_GOOGLE}
        initialRegion={{
          latitude: 37.78825,
          longitude: -122.4324,
          latitudeDelta: 0.005,
          longitudeDelta: 0.005,
        }}
        style={{ flex: 1 }}
      >
        <Marker
          image={require('../../assets/marker.png')}
          coordinate={{ latitude: 37.78825, longitude: -122.4324 }}
        />

        <Polyline
          coordinates={[
            { latitude: 37.788, longitude: -122.432 },
            { latitude: 37.786, longitude: -122.431 },
            { latitude: 37.787, longitude: -122.433 },
            { latitude: 37.785, longitude: -122.434 },
          ]}
          strokeWidth={5}
        />
      </MapView>

      <Button title="Go to Next" onPress={() => navigation.navigate('Next')} />
    </View>
  );
};

export default MapScreen;
```

### Steps to reproduce

1. Create a screen (MapViewScreen) with a MapView containing at least one Marker and a Polyline.
2. From that screen, push another screen onto the navigation stack.
3. Navigate back to the MapViewScreen using pop (or back navigation).
4. Observe that the MapView is still present, but the overlays (Marker/Polyline) are missing.

### Expected result

When navigating back to the MapViewScreen, all previously rendered state should still be visible and unchanged.

### Actual result

<img src="https://github.com/user-attachments/assets/27d78cbe-7e81-4b64-9366-82da122bc23e" width="200" />

<img src="https://github.com/user-attachments/assets/a9c345e6-e5de-4ec3-be86-3ac15c880f10" width="200" />

<img src="https://github.com/user-attachments/assets/d44b3d5d-c486-4d14-b79b-8aa09bceecf9" width="200" />

### React Native Maps Version

1.26.9

### What platforms are you seeing the problem on?

Android

[...]

PR #5765, opened 2025-10-05
fix(android): correctly add and remove markers

salah-ghanim:
### Does any other open PR do the same thing?

No

### What issue is this PR fixing?

https://github.com/react-native-maps/react-native-maps/issues/5750

https://github.com/react-native-maps/react-native-maps/issues/5669

### How did you test this PR?

Android emulator with very complex cases of adding / removing / navigating

PR #5770, opened 2025-10-07
Fix for marker/polyline disappearing after the first load when naviga…

saiyam-zepto-dev:

No.
This PR fixes a specific issue where map features such as MapMarker and Polyline disappear after navigating back to a screen containing a MapView. I couldn’t find any other PR addressing this behavior.

What issue is this PR fixing?

#5750 

The fix changes the feature key assignment logic to use the current map size instead of a fixed index, ensuring each feature has a unique key and preventing overwriting.

How did you test this PR?

Tested on:

Android (Google Maps)

iOS Simulator (Apple Maps)

Steps to reproduce and verify fix:

Open a screen containing a MapView with multiple markers and/or polylines.

Navigate to another screen.

Navigate back to the map screen.

Before fix: some or all markers/polylines disappear.

After fix: all features remain visible and render correctly.

✅ Fix summary: Prevent map features from disappearing when navigating back by using map size instead of fixed index for hash map keys.

### DIFF
--- a/android/src/main/java/com/rnmaps/fabric/MapViewManager.java
+++ b/android/src/main/java/com/rnmaps/fabric/MapViewManager.java
@@ -323,7 +323,7 @@ public class MapViewManager extends ViewGroupManager<MapView> implements RNMapsM
                 parent.addFeature(child, parent.getFeatureCount());
             });
         } else {
-            parent.addFeature(child, index);
+            parent.addFeature(child, parent.getFeatureCount());
         }
     }
 


ChielBruin:
Related to #5765

salah-ghanim:
@saiyam-zepto-dev can you please rebase based on latest master.

I don't understand specifically why would this fix the issue, but will wait on the community to report that it's solved

salah-ghanim:
the comment here: https://github.com/react-native-maps/react-native-maps/issues/5750#issuecomment-3388630099 says it doesn't solve the problem

cortigeronimo:
Any update for this?

Issue #5722, opened 2025-09-04
Attempt to invoke virtual method 'boolean java.util.LinkedList.isEmpty()' on a null object reference(Android)


SKHRAPP:
### Summary

“I’m not sure why this is happening. I’ve already updated to the latest react-native-maps version (1.26.1), but the issue still occurs. Can anyone help me figure out how to fix it?”

only In Android ,my IOS is working fine

<img width="358" height="779" alt="Image" src="https://github.com/user-attachments/assets/a719dd78-9f40-457f-bd57-0e4c391387fe" />

React Native Version - 0.81.1
react-native-map - 1.26.1

### Reproducible sample code

```TSX
import React,{useState, useRef, useEffect} from 'react';
import MapView, { PROVIDER_GOOGLE, Marker } from 'react-native-maps';


const TeamMap = () => {
  const mapRef = useRef(null);
  const [region, setRegion] = useState({
    latitude: 11.605846,
    longitude: 75.79384,
    latitudeDelta: 30.0,
    longitudeDelta: 0.02,
  });

 useEffect(() => {
    return () => {
      if (mapRef.current) {
        try {
          mapRef.current.setMapBoundaries(null, null); // force detach
          mapRef.current = null;
        } catch (e) {
          console.log('Error cleaning map:', e);
        }
      }
    };
  }, []);

return (
 <MapView
        clusterColor={Colors.white}
        clusterTextColor={Colors.primary}
        spiderLineColor={Colors.primary}
        clusterBorderWidth={2}
        provider={PROVIDER_GOOGLE}
        ref={mapRef}
        mapType={'standard'}
        initialRegion={region}
        onMapReady={() => {
          setTimeout(() => {
            if (mapRef.current) {
              mapRef.current.animateToRegion({
                latitude: parseFloat(latitude),
                longitude: parseFloat(longitude),
                latitudeDelta: 30.0, // Adjust delta values as needed
                longitudeDelta: 0.02,
              });
            }
          }, 1000);
        }}
        onRegionChangeComplete={setRegion}
        style={{ width: '100%', height: '100%' }}
      >
</MapView>
)
}

export default TeamMap
```

### Steps to reproduce

Jus open Map initial its opens then goback screen again map screen select error getting in android

### Expected result

need to proper load map and view

### Actual result

.

### React Native Maps Version

1.26.1

### What platforms are you seeing the problem on?

Android

### React Native Version

0.81.1

### What version of Expo are you using?

Not using Expo

[...]

PR #5788, opened 2025-10-22
fix(android): fix crash on some devices with doDestroy

salah-ghanim:

No

[5722](https://github.com/react-native-maps/react-native-maps/issues/5722)

### How did you test this PR?
Analyzing stacktrace


yittoo:
this is merged and released - thank you for that. but isn't this going to cause memory leak if destroy fails?

PR #5806, opened 2025-11-10
fix(ios): Prevent auto bottom inset adjustment during map initialization

lodev09:

No, there are no other open PRs addressing this specific issue. This PR fixes a bug where the Google Maps SDK automatically adjusts the bottom inset during map initialization on iOS, which conflicts with React Native Maps' intended default behavior.

This PR fixes the issue where iOS Google Maps automatically adjusts bottom inset during map initialization, causing the map region to be incorrectly positioned on devices with safe area insets (e.g., iPhones with notches).

**Root Cause:**
When `GMSMapView` is initialized via `[super initWithOptions:options]`, the Google Maps SDK sets `paddingAdjustmentBehavior` to its default value of `kGMSMapViewPaddingAdjustmentBehaviorAlways`. This causes the SDK to automatically add safe area insets (including bottom inset) to the map padding before React Native Maps can apply its intended default value of `Never`.

### How did you test this PR?

**Platform:** iOS (Google Maps)

**Testing Steps:**
1. Initialize a MapView component with an `initialRegion` prop on an iPhone with safe area insets (device with notch/Dynamic Island)
2. Observe that the map region is now correctly positioned without unwanted bottom padding adjustment during initialization
3. Verify that explicitly setting `paddingAdjustmentBehavior="always"` or `paddingAdjustmentBehavior="automatic"` still works as expected
4. Test on both old architecture and new architecture (Fabric) to ensure consistent behavior

**Changes Made:**
- Set `paddingAdjustmentBehavior` to `Never` immediately after `GMSMapView` initialization in `AIRGoogleMap.mm` to prevent Google Maps SDK default behavior from automatically adjusting padding based on safe area insets
- Add `paddingAdjustmentBehavior` handling in Fabric component's `updateProps` method (`RNMapsGoogleMapView.mm`) to properly respect the prop value from the React Native spec
- Ensures consistent behavior across both old and new architecture
- Default value (`Never`) is defined in the generated spec files (Props.h/Props.cpp)

**Tested on:**
- iOS Simulator (iPhone 15 Pro)
- [Add your real device testing here if applicable]

**Not tested on:**
- Android (this fix is iOS Google Maps specific)
- Apple Maps (not affected by this issue)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.26.19](https://github.com/react-native-maps/react-native-maps/compare/v1.26.18...v1.26.19) (2025-11-30)
+
+
+### Bug Fixes
+
+* re-order prop / feature color / stroke width selection ([#5800](https://github.com/react-native-maps/react-native-maps/issues/5800)) ([6d5c9a8](https://github.com/react-native-maps/react-native-maps/commit/6d5c9a842e43d221bc56026f6c280033d965d26f)), closes [#3385](https://github.com/react-native-maps/react-native-maps/issues/3385)
+
 ## [1.26.18](https://github.com/react-native-maps/react-native-maps/compare/v1.26.17...v1.26.18) (2025-10-22)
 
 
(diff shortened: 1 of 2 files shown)


salah-ghanim (approved):

salah-ghanim:
thanks @lodev09 merged

PR #5832, opened 2026-01-12
fix(example): get example working

wolfgang42:

Not so far as I can see at a glance

The example was not starting properly, which made it hard to test changes in-situ.

### How did you test this PR?

```sh
cd example
yarn install
yarn run ios
yarn run android
```
...and observe that the app now boots.

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -1,3 +1,4 @@
+import 'react-native-gesture-handler';
 import {AppRegistry} from 'react-native';
 import App from './src/App';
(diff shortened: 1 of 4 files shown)


wolfgang42:
Still relevant

wolfgang42:
...though I’m not sure why I even bothered to say that, since a library with a bot that auto-closes pull requests doesn’t seem like one that’s particularly interested in accepting contributions.

PR #5833, opened 2026-01-12
fix(android): implement userInterfaceStyle prop

wolfgang42:
Doesn’t look like it.

`<MapView userInterfaceStyle="...">` doesn’t do anything on Android.

I only fixed it on Fabric, so this is still broken on Paper; I get the impression that `android/src/main/java/com/rnmaps/maps/MapManager.java` needs to change, but I set `newArchEnabled=false` and `android/src/main/java/com/rnmaps/fabric/MapViewManager.java` was still being constructed so I'm not sure what’s going on there.

### How did you test this PR?

Combined with #5832 and tested using the included changes to the ThemeMap screen in the demo app.

### DIFF
--- a/android/src/main/java/com/rnmaps/fabric/MapViewManager.java
+++ b/android/src/main/java/com/rnmaps/fabric/MapViewManager.java
@@ -476,7 +476,7 @@ public class MapViewManager extends ViewGroupManager<MapView> implements RNMapsM
 
     @Override
     public void setUserInterfaceStyle(MapView view, @Nullable String value) {
-        // do nothing (initialProp)
+        view.setUserInterfaceStyle(value);
     }
 
     @Override
--- a/android/src/main/java/com/rnmaps/maps/MapView.java
+++ b/android/src/main/java/com/rnmaps/maps/MapView.java
@@ -55,6 +55,7 @@ import com.google.android.gms.maps.model.IndoorBuilding;
 import com.google.android.gms.maps.model.IndoorLevel;
 import com.google.android.gms.maps.model.LatLng;
 import com.google.android.gms.maps.model.LatLngBounds;
+import com.google.android.gms.maps.model.MapColorScheme;
 import com.google.android.gms.maps.model.MapStyleOptions;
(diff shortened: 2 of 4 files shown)


wolfgang42:
> Does this work when the user switches to dark mode without restarting the app?

I don’t think it explicitly fixes that, since I only wired up the prop. (But you can follow the testing instructions in the PR description to try for yourself.)

However it *should* at least let you work around the problem by passing <code>&lt;MapView userInterfaceStyle={<a href="https://reactnative.dev/docs/appearance#getcolorscheme">Appearance.getColorScheme</a>() ?? 'light'} .../&gt;</code> to explicitly pass the scheme through.

wolfgang42:
@github-actions Well, as far as I know the problem still exists and the PR merges cleanly.

But sure, since no maintainer has bothered to take a look at it I guess you may as well close it 🤷‍♂️🙄

fabriziocucci:
Confirming this does fix the no-restart case, which answers @dacoto's question above.

I hit the same bug and wrote the same patch before finding this PR, so treat this as independent corroboration rather than a competing change.

The reason it works: `GoogleMap.setMapColorScheme` applies at runtime and the Android setter being an empty method was the whole bug. The prop only ever reached `GoogleMapOptions` at construction, so the value was right on first paint and every later change was dropped on the floor.

Verified by logging inside the patched setter. Pixel 9 emulator, API 36 `google_apis_playstore`, react-native-maps 1.28.2, play-services-maps 19.1.0, new architecture:

```
setUserInterfaceStyle called value=dark  mapNull=false
applying scheme 1
setUserInterfaceStyle called value=light mapNull=false
applying scheme 0
setUserInterfaceStyle called value=dark  mapNull=false
applying scheme 1
```

Three theme toggles in a running app, map restyled each time, no restart and no remount. Scheme 1 is `MapColorScheme.DARK`, 0 is `LIGHT`.

One thing worth keeping as-is: the `initialProps` path in `createViewInstance` should stay, since that is what avoids a flash of the wrong scheme on first paint. This PR adds the runtime path without removing it, which is the right shape.

PR #5834, opened 2026-01-12
fix(ios): use relative path for Podfile pod reference

dawidvdh:

I dont think so.

Changes the iOS Expo config plugin to generate relative paths instead of absolute paths in the Podfile for the react-native-maps pod reference.

The current implementation uses require.resolve with File.dirname which resolves to an absolute path:

```
rn_maps_path = File.dirname(`node --print "require.resolve('react-native-maps/package.json')"`)
# Results in: /Users/username/project/node_modules/react-native-maps
```

This causes issues in CI environments where the build machine path differs from the development machine, resulting in Podfile.lock containing machine-specific absolute paths that fail on other systems.

### How did you test this PR?

Tested it on iOS via Expo.

### DIFF
--- a/plugin/src/ios.ts
+++ b/plugin/src/ios.ts
@@ -101,7 +101,11 @@ export function addMapsCocoapods(
   useGoogleMaps: boolean,
 ): MergeResults {
   let newSrc =
-    '  rn_maps_path = File.dirname(`node --print "require.resolve(\'react-native-maps/package.json\')"`) \n';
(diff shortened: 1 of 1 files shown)


oleksiihuspy:
It would great to get it merged for everyone who is using expo prebuild!

Tehnix:
Still relevant

dawidvdh:
Still relevant

Issue #5840, opened 2026-01-16
Previously rendered polylines stay rendered until the map is remounted (Android only)


ktroitsky:
### Summary

The Polyline components that were previously rendered on the map are not cleared when a completely different Polyline is configured for rendering. This creates the behaviour when multiple polylines are rendered on the map at the same time.

### Reproducible sample code

```TSX
import { skipToken, useQuery } from '@tanstack/react-query';
import React, { useCallback, useMemo, useState } from 'react';
import { StyleSheet } from 'react-native';
import MapView, { LatLng, Marker, Polyline, Region } from 'react-native-maps';

const DIRECTIONS_COLOR = '#3170f0';
const STROKE_WIDTH = 7;

function isCoordinateInRegion(region: Region, coord: LatLng): boolean {
  const latHalf = region.latitudeDelta / 2;
  const lngHalf = region.longitudeDelta / 2;

  return (
    coord.latitude >= region.latitude - latHalf &&
    coord.latitude <= region.latitude + latHalf &&
    coord.longitude >= region.longitude - lngHalf &&
    coord.longitude <= region.longitude + lngHalf
  );
}

const INITIAL_REGION: Region = {
  latitude: 37.78825,
  longitude: -122.4324,
  latitudeDelta: 0.05,
  longitudeDelta: 0.05,
};

type MarkerType = {
  latitude: number;
  longitude: number;
  path: LatLng[];
};

export default function MapMarkersWithPolyline() {
  const [selectedPoint, setSelectedPoint] = useState<MarkerType | null>(null);
  const [mapRegion, setMapRegion] = useState<Region>();

  const onRegionChangeComplete = useCallback((region: Region) => {
    setMapRegion(region);
  }, []);

  const markersInView = useMemo(
    () => markers.filter(marker => (mapRegion ? isCoordinateInRegion(mapRegion, marker) : false)),
    [mapRegion],
  );

  const { data } = useQuery({
    queryKey: [selectedPoint],
    queryFn: selectedPoint
      ? () => {
          return selectedPoint.path;
        }
      : skipToken,
  });

  return (
    <MapView
      style={styles.map}
      initialRegion={INITIAL_REGION}
      onRegionChangeComplete={onRegionChangeComplete}>
      {markersInView.map((marker, index) => (
        <Marker
          key={index}
          coordinate={marker}
          onPress={() => {
            setSelectedPoint(marker);
          }}
        />
      ))}

      {data && (
[...]

PR #5856, opened 2026-02-04
fix(android): features stuck

RayanMoarkech:

No

https://github.com/react-native-maps/react-native-maps/issues/5840

I am having this issue as well on v1.26.19+. It is not just for polylines. I have the issue for polygons as well.
Downgrading to v1.26.18 resolved the issue.

When features are being mounted/unmounted fast after each other, android previously rendered polylines or polygons remain mounted forever until map is fully remounted.

Checking the diff between versions: https://github.com/react-native-maps/react-native-maps/compare/v1.26.18...v1.26.19

Seems like a missed understanding of .add and .set for features with the index. 😄 

### How did you test this PR?

- Android Pixel 8 Pro Real Device
- Built the code into my project and checked if the issue is still there...


Smarker:
I haven't tried the latest react-native-maps version, but on `react-native-maps` version `1.23.12`, I had issues with my custom markers only rendering half of the markers.

When trying different versions, I also hit these issues:
* "Invariant Violation: TurboModuleRegistry.getEnforcing(...): 'RNMapsAirModule' could not be found.".
* Had to disable "New Architecture"
* Saw markers flicker

Taking your suggestion of using `1.26.18` for `react-native-maps` worked for me on my android pixel 10 (own phone, not emulator) and my custom markers now look great! 🙏

salah-ghanim (approved):

PR #5851, opened 2026-01-30
feat(ios): add appleLogoInsets prop for Apple Maps logo positioning

sbs44:
## Summary

Adds a new `appleLogoInsets` prop to allow repositioning the Apple Maps logo on iOS. This is useful when app UI elements would otherwise cover the logo.

- Uses `EdgePadding` type for flexible edge-based positioning (same as existing `legalLabelInsets`)
- Supports both Legacy and Fabric architectures
- iOS Apple Maps only

## Usage

```tsx
// Position from top-left
<MapView appleLogoInsets={{ top: 50, left: 10 }} />

// Position from bottom-right
<MapView appleLogoInsets={{ bottom: 100, right: 20 }} />
```

## Test Plan

- [x] Tested on iOS 26.2 simulator
- [x] Verified logo repositions correctly with `top`/`left` values
- [x] Verified logo repositions correctly with `bottom`/`right` values
- [x] Lint, TypeScript, and tests pass
- [x] Example added to `LegalLabel.tsx`

## Notes

- The Apple logo view class name changed from `MKAppleLogoImageView` to `MKAppleLogoLabel` in recent iOS versions; the implementation handles both
- Context/rationale: logo must remain visible for App Store compliance

PR #5858, opened 2026-02-09
fix(ios): reposition attribution views synchronously in layoutSubviews

sbs44:
## Does any other open PR do the same thing?

No.

## What issue is this PR fixing?

Fixes a bug where `appleLogoInsets` (#5851) and `legalLabelInsets` have no visible effect on iOS.

Both props use `dispatch_async(dispatch_get_main_queue(), ...)` to reposition MKMapView's internal subviews, but Auto Layout constraints run on every layout pass and immediately override the async frame changes.

## Changes

- Move repositioning into `layoutSubviews` (runs synchronously after Auto Layout resolves)
- Set `translatesAutoresizingMaskIntoConstraints = YES` to prevent constraint overrides
- Cache `_appleLogo`/`_legalLabel` refs; re-scan only when the view leaves the hierarchy
- Remove `updateAppleLogoInsets`/`updateLegalLabelInsets` dispatch_async methods
- Simplify setters to call `setNeedsLayout`

`layoutSubviews` is called ~3x at startup, not during pan/zoom — zero performance overhead. Inset-zero guard skips entirely for users who don't use these props.

## How did you test this PR?

- Tested on iOS 18.2 simulator and device
- Verified Apple logo and legal label reposition with top/left and bottom/right values
- Verified no repositioning when props are not set
- Confirmed layoutSubviews is not called during map pan/zoom (instrumented with os_log counters)
- Confirmed views survive MKMapView internal subview recycling
- Existing `LegalLabel.tsx` example works correctly

### DIFF
--- a/ios/AirMaps/AIRMap.mm
+++ b/ios/AirMaps/AIRMap.mm
@@ -829,8 +829,6 @@ const NSInteger AIRMapMaxZoomLevel = 20;
 
         [self updateScrollEnabled];
         [self updateZoomEnabled];
-        [self updateLegalLabelInsets];
-        [self updateAppleLogoInsets];
     }
 }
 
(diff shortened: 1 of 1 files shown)


sbs44:
This fixes a bug introduced in #5851 where `dispatch_async` frame changes are overridden by MKMapView's Auto Layout constraints. The same issue also affects the pre-existing `legalLabelInsets` implementation.

sbs44:
@salah-ghanim Would you be able to review this when you get a chance? Thanks!

sbs44:
Still relevant

sbs44:
Rebased onto latest `master`... no conflicts, and all four checks are green (`lint-test`, `build-android`, `build-ios`, `build-ios-frameworks`).

Quick recap since this has been open a while: `appleLogoInsets` (which I added in #5851) and the pre-existing `legalLabelInsets` both write frames inside `dispatch_async`, so MKMapView's Auto Layout pass runs afterward and overwrites them... meaning both props are silently no-ops in practice. This moves the repositioning into `layoutSubviews` after `[super layoutSubviews]` and sets `translatesAutoresizingMaskIntoConstraints = YES` so the positions stick.

Scope is a single file (`ios/AirMaps/AIRMap.mm`, +49/−54) and it's a net deletion.

@salah-ghanim — you're in CODEOWNERS here and merged the recent iOS work; would you have a few minutes to take a look?

PR #5859, opened 2026-02-11
fix(android): fix ghost features on MapView

kiwidood:

#5856 incorrectly adds to savedFeatures during pause when savedFeatures do not get restored on resume, potentially causing missing features from the map. It's also missing the guard in addFeature to prevent using addToMap on a stale map. 

When the Android MapView is detached from the window (e.g. navigating to another screen with react-native-screens), onDetachedFromWindow saves features to savedFeatures and clears features. However, getFeatureCount()/getFeatureAt() still return from the empty features list. React sees 0 native children, never calls removeFeatureAt, and only issues addFeature calls. This causes savedFeatures to grow with duplicates that all get  restored as ghost features on reattach.

### How did you test this PR?

  - Platform: Android with Google Maps
  - Tested on a real device (Samsung Galaxy A54 5G)
  - Added different batches of polygons and markers that periodically re-render via API fetches
  - Navigated away from the map screen and back repeatedly while updates occurred
  - Verified no duplicate/ghost polygons appear on the map
  - iOS is not affected by this issue


pedrolopes10:
thank you! it worked perfectly

orv1s:
Tested on `1.27.2`, resolves disappearing and duplicate default markers! Thanks @kiwidood

y25zhao:
It works like a charm! Ship it!

y25zhao (approved):

Shinyshark1:
This fixed the issue for me as well, I've made a .patch in my project until this is merged.

android/src/main/java/com/rnmaps/maps/MapView.java
```diff
@@ -243,6 +242,11 @@ public void onPause(LifecycleOwner owner) {
         }
     }
 
+    @Override
+    public void onPause(LifecycleOwner owner) {
+        pauseSafely();
```
falk-stefan, line 247: Why create a new method and not implement it directly inside `onPause()`?

falk-stefan:
I was able to test it as well. 

Using the patch:

[react-native-maps@1.27.2.patch](https://github.com/user-attachments/files/28678288/react-native-maps%401.27.2.patch)

with

```yaml
# pnpm-workspace.yaml
patchedDependencies:
  react-native-maps@1.27.2: patches/react-native-maps@1.27.2.patch
```

Markers are not finally unmounted correctly. Thanks a lot :heart:

falk-stefan (approved):

salah-ghanim (approved):

salah-ghanim:
@kiwidood thanks for the contribution, the change is clean and make sense.
I'm still not sure regarding modifying views off screen (adding and removing) from the savedFeatures but it does make sense

PR #5226, opened 2024-11-21
fix (ios): Add check for nil values before adding to subviews in iOS Google Maps

CamBowes:
Adds `nil` checks before pushing new subviews in iOS - Fixes #5217 

### Does any other open PR do the same thing?

N/A - No prior PR raised

### What issue is this PR fixing?

#5217 - *** -[__NSArrayM insertObject:atIndex:]: object cannot be nil issue

### How did you test this PR?

Tested in iOS Real Device & Simulator - Expo 52/RN 0.76.2 (New Architecture Enabled) (Google Maps)

Issue #5217, opened 2024-11-15
[Expo SDK 52] App creashes with error "NSInvalidArgumentException  *** -[__NSArrayM insertObject:atIndex:]: object cannot be nil"


BrodaNoel:
### Summary

In some way, I think this issue may be related to https://github.com/react-native-maps/react-native-maps/issues/5206 and https://github.com/react-native-maps/react-native-maps/issues/5214

After upgrading to SDK 52, I'm getting an issue when I remove and add new markers.

In my app, I could remove all the code unrelated to this issue, and get this small version of the "Map" component, where I can still reproduce the error. I will add a video in the comments, so you see how to reproduce it (you have to play around with the zoom).

I migrated that code to a Expo Snack, but you won't be able to reproduce it from the snack, because iOS seems to be broken with SDK 52.

Here is the SNACK: https://snack.expo.dev/@brodanoel/react-native-maps-not-working

### Reproducible sample code

```TSX
import { useState } from 'react';
import MapView, { Marker } from 'react-native-maps';

const getZoom = region => Math.round(Math.log(360 / region.longitudeDelta) / Math.LN2);

const initialRegion = {
  latitude: -31.4205589,
  latitudeDelta: 0.5,
  longitude: -64.1734649,
  longitudeDelta: 0.5,
};

const clusters = [
  { coordinate: { latitude: -31.42, longitude: -64.185 }, i: 'aaa' },
  { coordinate: { latitude: -31.439, longitude: -64.128 }, i: 'bbb' },
];

const Map = () => {
  const [pointsToShow, setPointsToShow] = useState([]);

  const onRegionChange = currentRegion => {
    const zoom = getZoom(currentRegion);

    if (zoom >= 12) {
      setPointsToShow([]);
    } else {
      setPointsToShow(clusters);
    }
  };

  return (
    <MapView
      style={{ flex: 1 }}
      provider="google"
      onRegionChangeComplete={onRegionChange}
      initialRegion={initialRegion}
    >
      {pointsToShow.map(point => {
        return <Marker key={point.i} pinColor="red" coordinate={point.coordinate} />;
      })}
    </MapView>
  );
};

export default Map;

```

### Steps to reproduce

You should get the error while zooming in and out. Please wait for the video that I'm gonna add in the comments

### Expected result

App not crashing

### Actual result

App crashes and get closed

### React Native Maps Version

1.18.0 and 1.20.0

### What platforms are you seeing the problem on?

iOS (Google Maps)

[...]

PR #5871, opened 2026-02-26
fix(ios): add nil check for subview insertion in Apple Maps

### LINKED ISSUE #5345: React Native MapView crash application when dynamic sub view are added or removed rapidly
### Summary

My Application "Air Map Israel" started failing in production on iOS
"expo": "~52.0.27",
"react-native": "0.76.6",
"react-native-maps": "1.18.0",

Failing on AIRMap.m method 
`  (void)insertReactSubview:(id<RCTComponent>)subview`
on last method line:
`    [_reactSubviews insertObject:(UIView *)subview atIndex:(NSUInteger) atIndex];`
**In some cases due to null subview and some cases the at index is grater than subviews size**
To fix it i wrapped it with a check
```TSX
    if (subview != nil) {
        NSUInteger safeIndex = MIN(atIndex, _reactSubviews.count);
        [_reactSubviews insertObject:(UIView *)subview atIndex:safeIndex];
    }
```
Did the same in method
`  (void)removeReactSubview:(id<RCTComponent>)subview`
```TSX
    if (subview != nil) {
        [_reactSubviews removeObject:(UIView *)subview];
    }
```
Need a fix for it.
Working fine on android

### Reproducible sample code

```TSX
// noinspection JSValidateTypes

import {View, Text, StyleSheet} from "react-native";
import MapView, {Marker} from "react-native-maps";
import {useEffect, useState} from "react";
import {Checkbox} from "expo-checkbox";

let isDebug = true;
let appTimerId = null;
let markersList = [];

    const [addMarkers, setAddMarkers] = useState(true);
    const [markers, setMarkers] = useState([]);

    const [appTimerTime, setAppTimerTime] = useState(0)

    const startAppTimer = () => {
        if (appTimerId) clearInterval(appTimerId);
        appTimerId = setInterval(() => {
            setAppTimerTime((previousTime) => previousTime + 1);
        }, 1000);
    }

    const onMapReady = async () => {
        isDebug && console.log(new Date(), 'onMapReady');
        startAppTimer();
    };

    function addMarker() {
        markersList.push({
            title: 'Added ' + new Date().toLocaleTimeString(),
            coordinate: {
                latitude: 32.2 + appTimerTime / 10,
                longitude: 34.8 + appTimerTime / 10,
            }
        })
    }

    useEffect(() => {
        addMarker();
        setMarkers(markersList);
    }, [appTimerTime]);

    return (
        <View
            style={styles.header}
        >
            <MapView
                style={{
                    width: "100%",
                    height: "80%",
                }}
                showsUserLocation={true}
                onMapReady={onMapReady}
            >
                {addMarkers && markers?.length > 0 &&
                    markers.map((m, i) => {
                        return (
                            <Marker key={'marker_' + i}
                                    zIndex={99}
                                    coordinate={m.coordinate}
                                    anchor={{x: 0.5, y: 0.5}}
                                    title={m.title}
                            >
                            </Marker>
                        )
                    })
                }
            </MapView>
            <Text style={{
                padding: 10,
                fontWeight: 'bold',
                height: 50,
            }}>
                {'Show Markers'}
            </Text>
            <Checkbox
                value={addMarkers}
                onValueChange={(value) => setAddMarkers(value)}
                color={addMarkers ? 'blue' : undefined}
            />
        </View>
    );
}

    header: {
        flex: 1,
        justifyContent: "center",
        alignItems: "center",
        backgroundColor: 'oldlace'
    },
});
```

### Steps to reproduce

This demo Application adds markers to map view dynamically
There is a checkbox to enable and disable it

Play with the checkbox and see ios app crashes

### Expected result

Markers will be shown or removed from the map based on the checkbox

### Actual result

App crashes

### React Native Maps Version

1.18.0

iOS (Apple Maps)

### React Native Version

0.76.6

SDK 52

### Device(s)

iPhone 14 pro, iPhone 16 Pro simulator 

### Additional information

_No response_

rohhj622:

PR #5230 addressed the same issue for Apple Maps, but it was automatically closed as stale on March 10, 2025 without any review or merge. The author did not continue working on it. Issue #5345 remains open and the crash is still reproducible on the latest version.

Fixes #5345
Related: #5217, #5230

### How did you test this PR?

- **Platform**: iOS (Apple Maps)
- **Environment**: Expo SDK 54, React Native with New Architecture enabled, iOS 26.2
- **Device**: Real device — iPhone 16 Pro (not simulator)
- **Before fix**: Frequent `NSInvalidArgumentException` crash (`object cannot be nil`) when dynamically adding/removing map markers during zoom/pan
- **After fix**: Zero crashes over several months of production use
- **Side effects**: None — markers display correctly, no disappearing markers (which was a concern raised in the related Google Maps PR #5226)

### DIFF
--- a/ios/AirMaps/AIRMap.mm
+++ b/ios/AirMaps/AIRMap.mm
@@ -157,7 +157,10 @@ const NSInteger AIRMapMaxZoomLevel = 20;
             [self insertReactSubview:(UIView *)childSubviews[i] atIndex:atIndex];
         }
     }
-    [_reactSubviews insertObject:(UIView *)subview atIndex:(NSUInteger) atIndex];
+    if (subview != nil) {
+        NSUInteger safeIndex = MIN((NSUInteger)atIndex, _reactSubviews.count);
+        [_reactSubviews insertObject:(UIView *)subview atIndex:safeIndex];
+    }
 }
 #pragma clang diagnostic pop
 


bcawrse:
Thank you! 

I'm clustering pins in the front-end due to constraints and when the region changes I query a new data-set & re-cluster / load pins to a global grid & zoom level. I've taking great precautions to tip-toe around this crash, and it still occurs.

For anyone hitting this crash who can't wait for the PR to merge, you can apply this fix locally using patch-package:

  1. Install patch-package
  npm install --save-dev patch-package

  2. Add postinstall to package.json
  "scripts": {
    "postinstall": "patch-package"
  }

  3. Edit node_modules/react-native-maps/ios/AirMaps/AIRMap.mm

  Find insertReactSubview:atIndex: and apply two changes:

  Add a nil guard at the top of the method:
  - (void)insertReactSubview:(id<RCTComponent>)subview atIndex:(NSInteger)atIndex {
      if (subview == nil) return;  // ← add this
      ...

  Replace the final insertObject call with a bounds-safe version:
  // before
  [_reactSubviews insertObject:(UIView *)subview atIndex:(NSUInteger)atIndex];

  // after
  NSUInteger safeIndex = MIN((NSUInteger)atIndex, _reactSubviews.count);
  [_reactSubviews insertObject:(UIView *)subview atIndex:safeIndex];

  4. Generate and commit the patch
  npx patch-package react-native-maps
  git add patches/
  git commit -m "patch react-native-maps nil subview guard"

  The patch will be re-applied automatically on every npm install via the postinstall hook.

lejahmie:
Corroborating evidence that this fix is needed — with a symbolicated production crash.

We hit this exact crash on **Apple Maps** and traced it to the line this PR guards. Sharing the details in case it helps get this merged (it's currently marked stale).

**Environment**
- react-native-maps 1.27.2
- React Native 0.85.3, **New Architecture (Fabric)** enabled, Hermes
- Expo prebuild, iOS 26.5.1, physical device (iPhone 17 Pro Max)

**Crash (from Apple's `.ips` report — `bug_type 309`, `EXC_CRASH / SIGABRT`)**
```
*** -[__NSArrayM insertObject:atIndex:]: index 37 beyond bounds for empty array
```
Symbolicated crashed thread (main):
```
-[__NSArrayM insertObject:atIndex:]
-[AIRMap insertReactSubview:atIndex:]            AIRMap.mm
-[RNMapsMapView mountChildComponentView:index:]  RNMapsMapView.mm
RCTPerformMountInstructions(...)                 React (Fabric mount transaction)
-[RCTMountingManager performTransaction:]
```

**Root cause (why the index goes out of bounds)**
Under Fabric, component views are recycled. `RNMapsMapView`'s `prepareForRecycle` nils out the underlying map:
```objc
- (void) prepareForRecycle {
    [super prepareForRecycle];
    [_view removeFromSuperview];
    _view = nil;
    ...
}
```
On reuse a **fresh `AIRMap`** is created whose `_reactSubviews` is an **empty** `NSMutableArray`. Fabric then replays its mount instructions and calls `mountChildComponentView:index:` with a child's **original index** (37 in our case) → the unguarded `[_reactSubviews insertObject:subview atIndex:37]` throws `NSRangeException` → `abort()`.

Note the exception text is **"index N beyond bounds for empty array"** — i.e. this is specifically the **out-of-range index** case, so the `MIN((NSUInteger)atIndex, _reactSubviews.count)` clamp in this PR is exactly what's required (a nil-only guard would not have prevented our crash).

**How to trigger it**
Two `MapView`s that briefly coexist during a navigation transition: one screen with a long-lived map + many markers, and another screen that also renders a `MapView` and unmounts (→ Fabric recycle pool) during the same transaction. This matches @bcawrse's note above (front-end clustering + re-querying markers on region change).

**Why upgrading doesn't help**
The insert is still unguarded on `master` (and v1.29.0) — same `AIRMap.mm` line — so this needs the guard, not a version bump.

**Verification**
We applied this PR's change verbatim as a local `patch-package`/`pnpm patch` on 1.27.2 (both `AIRMap.mm` and the Google equivalent in #5940). [Confirmed on-device: the Apple Maps crash no longer reproduces across repeated runs of the path above. — UPDATE THIS LINE AFTER YOUR REBUILD/DEVICE TEST]

The removal path is index-independent (`removeReactSubview:` uses `removeObject:`) and marker z-order on a map isn't meaningful, so clamping the insert index has no behavioral downside.

Thanks @rohhj622 — hope this helps push it over the line.

lejahmie:
Still relevant

daredoes:
We hit this on iOS 26.2 with Apple Maps and the new architecture, and ended up shipping the same `MIN()` clamp as a patch against 1.27.2 (the line is unchanged as of 1.29.0), so getting this merged would let us drop that patch. One thing I found while tracing it: the index runs past the end because `RNMapsMapView.prepareForRecycle` nils `_view`, so any child mounted before the next `updateProps` rebuilds it is sent to nil and silently dropped, which leaves `_reactSubviews` empty while Fabric keeps handing up indices from the shadow tree (adding `[self prepareMapView]` at the top of `mountChildComponentView` closed both the crash and the missing overlays for us, and might also cover #5911). Would it make sense to fold that line into this PR, or would you rather it came as its own?

salah-ghanim (approved):

Issue #5863, opened 2026-02-13
Android: IndexOutOfBoundsException in removeFeatureAt with Fabric renderer


isaacrowntree:
### Summary

`MapView.removeFeatureAt()` throws `IndexOutOfBoundsException` when the Fabric renderer emits stale remove indices during rapid marker updates (map panning with clustering, or navigating away from a screen with many markers).

This is the same class of bug as #5497, which was fixed for `addFeature` in v1.24.15 via index clamping in `MapViewManager.addView()`. However, `removeFeatureAt` was never given the same bounds guard and still crashes. `getFeatureAt` also lacks bounds checking (as reported by @scottaglia in #5497).

Related PRs:
- #5511 (proposed bounds checks for all three methods — closed without merge)
- #5859 (open — adds `savedFeatures` redirect but no bounds check on the normal `features` path)

### Reproducible sample code

```TSX
import React, { useState, useEffect } from 'react';
import MapView, { Marker } from 'react-native-maps';

// Generate many markers that change rapidly (simulates clustering/panning)
function generateMarkers(count: number, offset: number) {
  return Array.from({ length: count }, (_, i) => ({
    id: `marker-${i}`,
    coordinate: {
      latitude: -41.28 + (i * 0.01) + (offset * 0.001),
      longitude: 174.77 + (i * 0.01),
    },
  }));
}

export default function App() {
  const [markers, setMarkers] = useState(generateMarkers(200, 0));

  // Simulate rapid marker updates (e.g. from clustering or data refresh)
  useEffect(() => {
    let offset = 0;
    const interval = setInterval(() => {
      offset++;
      // Vary the count to trigger add/remove race conditions
      const count = 150 + Math.floor(Math.random() * 100);
      setMarkers(generateMarkers(count, offset));
    }, 500);
    return () => clearInterval(interval);
  }, []);

  return (
    <MapView
      style={{ flex: 1 }}
      initialRegion={{
        latitude: -41.28,
        longitude: 174.77,
        latitudeDelta: 1,
        longitudeDelta: 1,
      }}
    >
      {markers.map((m) => (
        <Marker key={m.id} coordinate={m.coordinate} />
      ))}
    </MapView>
  );
}
```

### Steps to reproduce

1. Render a `MapView` with 100+ markers on Android with Fabric (New Architecture) enabled
2. Rapidly change the marker array (e.g. pan the map with clustering, or toggle marker visibility)
[...]

Issue #5807, opened 2025-11-12
Android: App crashes when dynamically adding/removing UrlTile components from MapView


isaacrowntree:
### Summary

On Android, dynamically toggling UrlTile components on and off within a MapView causes the application to crash. This occurs even when toggling slowly (not just rapid toggling). The crash happens when UrlTile components are conditionally rendered based on state changes, specifically when regions are enabled/disabled in an offline maps feature.

### Reproducible sample code

```TSX
import React, { useState } from 'react';
import { View, Button, StyleSheet } from 'react-native';
import MapView, { UrlTile } from 'react-native-maps';

const INITIAL_REGION = {
  latitude: -25.2744,
  longitude: 133.7751,
  latitudeDelta: 20,
  longitudeDelta: 20,
};

export default function App() {
  const [showTiles, setShowTiles] = useState(true);

  return (
    <View style={styles.container}>
      <MapView
        style={styles.map}
        initialRegion={INITIAL_REGION}
      >
        {showTiles && (
          <UrlTile
            key="offline-tiles"
            urlTemplate="file:///path/to/tiles/{z}/{x}/{y}.png"
            tileSize={256}
            zIndex={1000}
            shouldReplaceMapContent={true}
          />
        )}
      </MapView>
      
      <View style={styles.buttonContainer}>
        <Button
          title={showTiles ? "Disable Tiles" : "Enable Tiles"}
          onPress={() => setShowTiles(!showTiles)}
        />
      </View>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
  },
  map: {
    flex: 1,
  },
  buttonContainer: {
    position: 'absolute',
    bottom: 50,
    left: 20,
    right: 20,
  },
});
```

### Steps to reproduce

1. Create a React Native app with react-native-maps
2. Add a MapView component with conditional UrlTile rendering
3. Add a button/toggle to change the state that controls UrlTile visibility
4. Run the app on an Android device/emulator
5. **Important**: Pan the map so it's viewing the geographic region where the UrlTile covers (e.g., if using Australian tiles, view Australia)
6. Toggle the UrlTile on/off 2-3 times slowly (wait 2-3 seconds between toggles)
7. App crashes on Android

**Critical Finding**: The crash **only occurs when the MapView is actively displaying the region covered by the UrlTile**. If the map is panned to a different location (e.g., viewing Europe when toggling Australian tiles), toggling works without crashing. This suggests the crash is related to:
[...]

PR #5878, opened 2026-03-16
fix(android): bounds checks for Fabric stale indices in feature operations

isaacrowntree:

- #5511 proposed the same bounds checks but was closed without merge
- #5859 takes a different approach (savedFeatures redirect) but doesn't address the index clamping in `addFeature` or the tile overlay null safety

This PR is a minimal, focused fix for the index bounds crashes.

- #5863 — `IndexOutOfBoundsException` in `removeFeatureAt` / `getFeatureAt` with Fabric renderer
- #5807 — App crashes when dynamically adding/removing `UrlTile` components from `MapView`

Both issues share the same root cause: the Fabric renderer can issue mount/unmount instructions with stale indices that don't match the current `features` list size. This happens during rapid child updates (marker clustering, map panning, toggling tile overlays on/off).

### Root cause

Fabric's shadow tree and the native `features` ArrayList can become temporarily out of sync. When Fabric dispatches a batched mount item with an index based on the shadow tree state, the native list may have already changed size — causing `IndexOutOfBoundsException` on `ArrayList.add()` or `ArrayList.remove()`.

### Changes

**`MapView.java`:**
- **`addFeature`**: Clamp index to `[0, features.size()]` before processing. This is the same approach that was proposed in #5511 and confirmed working by @Bordei08 in #5863.
- **`removeFeatureAt`**: Add bounds guard (`index < 0 || index >= features.size()`) to silently skip stale remove indices instead of crashing.
- **`getFeatureAt`**: Add `index >= 0` check (previously only checked upper bound).

**`MapUrlTile.java` / `MapLocalTile.java`:**
- Add null check on `tileOverlay` in `removeFromMap()` to prevent `NullPointerException` when a tile overlay is removed before it was fully added (rapid toggle scenario from #5807). `MapWMSTile` inherits this fix via `MapUrlTile`.

### How did you test this PR?

- Platform: Android with Google Maps (Fabric / New Architecture enabled)
- Tested on a real device with react-native-maps 1.27.1
- Reproduced #5863 by rapidly panning a map with 200+ clustered markers, causing Fabric to send stale indices — crash no longer occurs
- Reproduced #5807 by toggling UrlTile components on/off while panning — crash no longer occurs
- Verified no behavioral change for normal (in-bounds) operations
- iOS is not affected by these issues

### Risk assessment

This is a defensive-only change — all four modifications only affect the error case (out-of-bounds index or null overlay) which currently crashes. Normal operation is completely unchanged.

### DIFF
--- a/android/src/main/java/com/rnmaps/maps/MapLocalTile.java
+++ b/android/src/main/java/com/rnmaps/maps/MapLocalTile.java
@@ -146,6 +146,8 @@ public class MapLocalTile extends MapFeature {
 
     @Override
     public void removeFromMap(Object map) {
-        tileOverlay.remove();
+        if (tileOverlay != null) {
+            tileOverlay.remove();
+        }
     }
 }
--- a/android/src/main/java/com/rnmaps/maps/MapUrlTile.java
+++ b/android/src/main/java/com/rnmaps/maps/MapUrlTile.java
@@ -202,6 +202,8 @@ public class MapUrlTile extends MapFeature {
 
   @Override
   public void removeFromMap(Object map) {
-    tileOverlay.remove();
+    if (tileOverlay != null) {
+      tileOverlay.remove();
+    }
   }
 }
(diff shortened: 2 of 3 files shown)


tb-nishant:
Still relevant

Issue #5901, opened 2026-04-14
[Android] Map polyline resets to initial state when map is re-attached


CptFabulouso:
### Summary

When props of MapPolyline change, like `coordinates`, `strokeColor` etc., changes are reflected on screen, but if user moves to another screen (MapView is detached) and returns back, then the displayed polyline returns to initial state, changes are no longer reflected.

### Reproducible sample code

```TSX
https://github.com/CptFabulouso/react-native-maps-issues/tree/Issue-1
```

### Steps to reproduce

- Add points with "Add polyline point" button
- Change color with "Change polyline color" button
- Open another screen with "Open another screen" button
- Go back with "Go back" button
- polylines and colors reset to initial state (even thought the actual state is not changed)

### Expected result

When navigation back, we should see the same polyline currently displayed when we left the screen

### Actual result

We see the polyline with it's initial state

### React Native Maps Version

1.27.2

### What platforms are you seeing the problem on?

Android

### React Native Version

0.85.1

### What version of Expo are you using?

Not using Expo

### Device(s)

emulator

CptFabulouso:
Still relevant

PR #5902, opened 2026-04-14
fix: remove features from map when MapView is detached, clear polylineOptions on polyline remove

CptFabulouso:

Did not find one

https://github.com/react-native-maps/react-native-maps/issues/5901

### How did you test this PR?

Affects only Android, Tested on emulator. Problem is that when polylines are added back when MapView is attached to window, there already exists `polylineOptions` with old (initial) `coordinates` and other values. This change:
a) calls `removeFromMap` on all features when MapView is detached from window
b) sets `polylineOptions` to `null`, so they are recreated after re-attaching

I am really not sure about a), I may be missing something, but since all features are added back when the MapView is attached, it makes sense to me to remove them all when MapView is detached.

Anyway re-creating `polylineOptions` on every `addToMap` call would be enough to fix the issue.

I've tested if it's not fixed by this PR https://github.com/react-native-maps/react-native-maps/pull/5866, since it does some cleanup, but no.

### DIFF
--- a/android/src/main/java/com/rnmaps/maps/MapPolyline.java
+++ b/android/src/main/java/com/rnmaps/maps/MapPolyline.java
@@ -190,6 +190,7 @@ public class MapPolyline extends MapFeature {
     @Override
     public void removeFromMap(Object collection) {
         PolylineManager.Collection polylineCollection = (PolylineManager.Collection) collection;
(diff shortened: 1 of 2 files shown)


CptFabulouso:
Still relevant

Issue #5906, opened 2026-04-21
Android Custom Map Marker / Thumbnail Pin


Cwilliams1997:
### Summary

**Android Custom Map Marker / Thumbnail Pin — Investigation Summary**

**Environment**

- React Native / Expo (JavaScript)
- react-native-maps@1.26.18
- react-native-map-clustering
- Android New Architecture (Fabric renderer)
- Physical Pixel device (~3x density)

**Core Problem**

Custom marker views containing remote images (<Image source={{ uri: ... }}>) inside <Marker> children do not render on Android with React Native New Architecture. Android's Fabric renderer rasterises the custom marker view into a bitmap before async image data has finished painting, resulting in blank/empty pins. The same code works correctly on iOS.

**What Was Tried (All Failed or Caused New Issues)**

**1. tracksViewChanges toggling**
Setting tracksViewChanges={true} until image loads, then flipping to false. Theoretically forces a re-snapshot after image paints. In practice caused addViewAt: failed to insert view crashes on Android physical devices at scale (100+ markers). Unreliable.

**2. onLoad-based preloading with hidden preloader views**
Rendering hidden <Image> components off-screen to warm the cache before mounting markers. Did not reliably solve the timing issue — images cached to disk but not yet decoded into memory at snapshot time.

**3. expo-image with cachePolicy="memory-disk"**
Swapping <Image> for <Image> from expo-image inside the custom marker view. Keeps decoded bitmaps in memory rather than just disk. Partial improvement but did not fully solve blank pins on cluster expansion or navigation return.

**4. Navigation bounce workaround (Profile → Explore)**
On first load: navigate to Profile tab then immediately back to Explore. Forces the map to remount, clearing the native layer, then zoom out/in cycle re-snapshots markers after images are in memory cache. Worked reliably for cold launch but did not solve cluster expansion. Caused noticeable flash/jank and broke on child cluster expansion. Reverted.

**5. Zoom out/in cycle on handleSearchSelect**
Zooming to latitudeDelta: 8 instantly, then zooming back to target location after delay. Cleared the native marker layer so new markers mounted with images already in cache. Worked for search location changes but not for cluster expansion.

**6. renderCluster onPress zoom intercept**
Intercepting cluster press, zooming out briefly, calling onPress(), then zooming back in. Worked for single-level clusters but broke on nested/child clusters because the fixed latitudeDelta target overrode the library's own zoom logic. Reverted to original renderCluster with plain onPress={onPress}.

[...]

Issue #5877, opened 2026-03-06
[Android] Custom View markers broken with New Architecture on Expo SDK 54 (RN 0.81)


kmiller9393:
### Summary

Custom React Native View markers (children inside `<Marker>`) do not render correctly on **Android** when **New Architecture is enabled** on Expo SDK 54 (React Native 0.81). Markers are invisible, clipped to quarter circles, or flicker constantly. This blocks adoption of New Architecture for any app relying on custom markers. I previously hit this same issue on Expo SDK 52 (RN 0.76.9) and had to disable New Architecture as a workaround. This is not an issue on iOS. Thank you for the help!

### Reproducible sample code

```TSX
import React from 'react';
  import { View, Text, StyleSheet } from 'react-native';
  import MapView, { Marker } from 'react-native-maps';

  // Minimal reproduction of custom View markers broken on Android + New Architecture
  export default function App() {
    return (
      <MapView
        style={{ flex: 1 }}
        initialRegion={{
          latitude: 37.78, longitude: -122.42,
          latitudeDelta: 0.05, longitudeDelta: 0.05,
        }}
      >
        {/* Individual marker: ring + dot (how we render item locations) */}
        <Marker
          coordinate={{ latitude: 37.78, longitude: -122.42 }}
          tracksViewChanges={false}
          anchor={{ x: 0.5, y: 0.5 }}
        >
          <View style={styles.wrap}>
            <View style={styles.ring} />
            <View style={styles.dot} />
          </View>
        </Marker>

        {/* Cluster marker: circle with count text (how we render Supercluster clusters) */}
        <Marker
          coordinate={{ latitude: 37.79, longitude: -122.43 }}
          tracksViewChanges={false}
          anchor={{ x: 0.5, y: 0.5 }}
        >
          <View style={styles.clusterWrap}>
            <View style={styles.clusterHalo} />
            <View style={styles.clusterCore}>
              <Text style={styles.clusterCount}>24</Text>
            </View>
          </View>
        </Marker>
      </MapView>
    );
  }

  const styles = StyleSheet.create({
    wrap: {
      width: 34,
      height: 34,
      alignItems: 'center',
      justifyContent: 'center',
    },
    ring: {
      position: 'absolute',
      width: 34,
      height: 34,
      borderRadius: 17,
      backgroundColor: 'rgba(2,27,156,0.3)',
      borderWidth: 1,
      borderColor: 'rgba(2,27,156,0.5)',
    },
    dot: {
[...]

PR #5860, opened 2026-02-12
fix(android): update marker size in onLayout to fix clipping

jk99k:
### Does any other open PR do the same thing?

I have searched through the existing PRs related to issue #5165. While there have been discussions regarding this behavior, I did not find a pull request that correctly addresses the synchronization between the Android native View layout and the Google Maps marker bitmap dimensions using the onLayout override. My PR provides a robust fix that works even when tracksViewChanges is set to false.

### What issue is this PR fixing?

#5165

### How did you test this PR?

- Platform: Android
- Maps API: Google Maps
- Device: Tested on a real Android device.
- Verification: Verified using a custom marker with dynamic content (text/images).
Test Plan:
1. Use a Marker with a custom View that changes its size dynamically (e.g., toggling between a short and long string).
2. Without this PR: The marker's hit area and visual bounds remain at the initial size (often defaulting to 100x100px or clipped), causing the new content to be partially invisible or misaligned.
3. With this PR: The onLayout event captures the new dimensions, updates the internal width/height, and triggers update(true) to refresh the marker's bitmap on the Google Maps layer. The marker now displays correctly at any size.

Issue #5728, opened 2025-09-09
Android: Custom markers don't display fully


stjasink:
### Summary

I have some markers that use Views and CSS to display a custom marker as an outline containing a number, with a pin below. On iOS these display as intended. On Android, only the outline displays and the number text and pin are missing.

This worked fine in version 1.20.1 before I upgraded to 1.26.1

I have seen a few issues raised about Android markers on here, but not specifically like this. So this might be a symptom of a more general issue with Android markers.

On IOS, displays as intended:

<img width="241" height="312" alt="Image" src="https://github.com/user-attachments/assets/c358601d-c635-4093-91f8-5130e99d6ed2" />

On Android, only partially displayed:

<img width="281" height="346" alt="Image" src="https://github.com/user-attachments/assets/8fe0884b-2cf1-43b6-9ef0-46c31f0a9c2e" />

### Reproducible sample code

```TSX
import MapView, { Marker } from 'react-native-maps';
import { StyleSheet, Text, View } from 'react-native';

function App() {

  return (
    <View style={styles.container}>
      <MapView
        provider="google"
        style={styles.map}
        initialRegion={{
          latitude: 37.78825,
          longitude: -122.4324,
          latitudeDelta: 0.0922,
          longitudeDelta: 0.0421,
        }}
      >
        <Marker
          coordinate={{latitude: 37.79, longitude: -122.45}}
          tracksViewChanges={false}>
          <View style={styles.markerContainer}>
            <View style={styles.markerTop}>
              <Text style={styles.markerText}>5</Text>
            </View>
            <View style={styles.markerPin} />
          </View>
        </Marker>
      </MapView>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    ...StyleSheet.absoluteFillObject,
    justifyContent: 'flex-end',
    alignItems: 'center',
  },
  map: {
    ...StyleSheet.absoluteFillObject,
  },
  markerContainer: {
    flexDirection: 'column',
    alignItems: 'center',
  },
  markerTop: {
    backgroundColor: 'white',
    borderRadius: 10,
    padding: 5,
    borderWidth: 3,
    borderColor: 'green',
  },
  markerPin: {
    width: 3,
    height: 5,
    backgroundColor: 'green',
  },
  markerText: {
    fontWeight: 'bold',
    color: 'black',
  },
});

export default App;
```

### Steps to reproduce

Create a new project with react native 0.81.1
Install react-native-maps 1.26.1
[...]

PR #5913, opened 2026-05-06
fix(android): custom Marker views clipped due to undersized bitmap

bkingery:

[#5860](https://github.com/react-native-maps/react-native-maps/pull/5860) targets the same symptom (clipped custom markers on Android) but uses a different approach — it updates this.width/this.height in onLayout and triggers a marker refresh when dimensions change. That fix helps when onLayout eventually fires with correct dimensions, but does not cover the Fabric case where the marker's own Yoga layout box is "correct" from Yoga's perspective yet still smaller than the laid-out React subtree. This PR fixes it at createDrawable() time by walking the subtree to compute the union of descendant bounds for the bitmap. The two approaches are complementary. PR [#5860](https://github.com/bkingery/react-native-maps/issues/5860) has been open since February 2026 with no merge activity.

#5877 Custom View markers broken with New Architecture (invisible, clipped to quarter circles, or flickering)
#5728 Custom markers don't display fully with tracksViewChanges={false}
#5165 Markers are clipped on Android
#5906 is related but describes a different root cause — it's about the Fabric renderer rasterizing the bitmap before async remote images finish loading (a timing/snapshot problem)

### How did you test this PR?

Verified using a local patch on Pixel 10 device. See screenshots below.

1. Create a Fabric-enabled app with a custom `Marker` whose child views are larger than the Yoga-assigned marker box (e.g., the reproduction cases from [#5877](https://github.com/bkingery/react-native-maps/issues/5877) — concentric circle markers, cluster markers with text)
2. Verify custom markers render fully without clipping
3. Verify no Fabric ReactViewGroup assertion crashes ("must be measured before being added to a parent")
4. Verify default pin markers and image-only markers still render correctly
5. Verify markers with MapCallout children still work correctly
6. Verify tracksViewChanges={false} works without visual regressions

Example code:
```js
import { Platform, StyleSheet, Text, View } from 'react-native';
import MapView, { Marker, PROVIDER_GOOGLE } from 'react-native-maps';

export function CustomMarkerRepro() {
  return (
    <MapView
      style={StyleSheet.absoluteFill}
      provider={Platform.OS === 'android' ? PROVIDER_GOOGLE : undefined}
      initialRegion={{
        latitude: 37.78825,
        longitude: -122.4324,
        latitudeDelta: 0.05,
        longitudeDelta: 0.05,
      }}
    >
      <Marker coordinate={{ latitude: 37.78825, longitude: -122.4324 }}>
        <View style={styles.pill}>
          <Text style={styles.price}>$1,234</Text>
        </View>
      </Marker>
    </MapView>
  );
}

  pill: {
    flexShrink: 0,
    alignItems: 'center',
    justifyContent: 'center',
    paddingHorizontal: 16,
    paddingVertical: 4,
    borderRadius: 16,
    borderWidth: 1,
    backgroundColor: '#fff',
  },
  price: {
    fontSize: 14,
    fontWeight: '600',
  },
});
```

### Summary

- Fix custom `<Marker>` views on Android being clipped when rasterized to a Google Maps icon on React Native Fabric
- Add `expandSnapshotSizeFromSubtree()` helper that walks the marker's view hierarchy and computes the union of descendant bounds to determine minimum bitmap dimensions
- Only uses laid-out dimensions (`getWidth`/`getHeight`, fallback `getMeasuredWidth`/`getMeasuredHeight`) — never calls `View#measure()`, which would trigger Fabric's `ReactViewGroup` assertion

### Problem

On Fabric, the Yoga layout box for a `MapMarker` can be smaller than its laid-out React subtree. Since `createDrawable()` uses the marker view's `width`/`height` to allocate the `Bitmap`, custom marker content gets clipped.

### Approach

A recursive helper accumulates `left`/`top` offsets while walking the subtree, computing `offset + size` for each descendant to find the true right/bottom edges (union of bounds). This aligns with the iOS approach in `AIRGoogleMapMarker` (`layoutSubviews` uses max right/bottom from child frames).

Non-custom markers (default pin, image-only) are unaffected — the new path only runs when `hasCustomMarkerView` is true. `MapCallout` children are skipped, consistent with existing library convention.

### Alignment with iOS behavior
On Google Maps iOS, custom marker content is hosted in an iconView whose size is derived from subview frames in AIRGoogleMapMarker’s layoutSubviews (effectively growing the container to fit laid-out children). On Apple Maps iOS, AIRMapMarker similarly adjusts bounds when the first child’s frame is larger than the marker’s React frame (layoutSubviews → reactSetFrame:). Android has no live iconView path—markers use a rasterized bitmap—so this change applies the same idea at snapshot time: bitmap dimensions are expanded using already laid-out subtree sizes instead of trusting only the parent Yoga box, avoiding clipped custom markers without calling measure() on Fabric-managed views. See [AIRGoogleMapMarker](https://github.com/react-native-maps/react-native-maps/blob/99183afdd8c4d814cc05e324c9e8c60fd812c2da/ios/AirGoogleMaps/AIRGoogleMapMarker.m#L62-L76) and [AIRMapMarker](https://github.com/react-native-maps/react-native-maps/blob/99183afdd8c4d814cc05e324c9e8c60fd812c2da/ios/AirMaps/AIRMapMarker.m#L421-L433) for current iOS behavior.

| Before | After |
|--|--|
|Screenshot 2026-05-06 at 3 32 51 PM|Screenshot 2026-05-06 at 3 32 40 PM|

### DIFF
--- a/android/src/main/java/com/rnmaps/maps/MapMarker.java
+++ b/android/src/main/java/com/rnmaps/maps/MapMarker.java
@@ -9,6 +9,7 @@ import android.graphics.drawable.Animatable;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
 import android.view.View;
+import android.view.ViewGroup;
 import android.widget.LinearLayout;
 import android.animation.ObjectAnimator;
 import android.util.Property;
@@ -658,6 +659,22 @@ public class MapMarker extends MapFeature {
 
     private Bitmap mLastBitmapCreated = null;
 
+    // Walks the subtree to find the union of descendant bounds for bitmap sizing.
+    // Uses only laid-out dimensions — do NOT call View#measure() here (Fabric assertion).
+    private void expandSnapshotSizeFromSubtree(View view, int offsetX, int offsetY, int[] outWh) {
+        int w = view.getWidth() > 0 ? view.getWidth() : view.getMeasuredWidth();
+        int h = view.getHeight() > 0 ? view.getHeight() : view.getMeasuredHeight();
+        outWh[0] = Math.max(outWh[0], offsetX + w);
+        outWh[1] = Math.max(outWh[1], offsetY + h);
+        if (view instanceof ViewGroup) {
+            ViewGroup group = (ViewGroup) view;
+            for (int i = 0; i < group.getChildCount(); i++) {
+                View child = group.getChildAt(i);
+                expandSnapshotSizeFromSubtree(child, offsetX + child.getLeft(), offsetY + child.getTop(), outWh);
+            }
+        }
+    }
+
     private void clearDrawableCache() {
         mLastBitmapCreated = null;
     }
@@ -665,6 +682,27 @@ public class MapMarker extends MapFeature {
     private Bitmap createDrawable() {
         int width = this.width <= 0 ? 100 : this.width;
         int height = this.height <= 0 ? 100 : this.height;
+
+        // Bitmap must cover laid-out custom-marker subtree; Yoga box alone can be too small (Fabric).
+        if (hasCustomMarkerView) {
+            int[] expanded = new int[] { width, height };
+            for (int i = 0; i < getChildCount(); i++) {
+                View child = getChildAt(i);
+                if (child instanceof MapCallout) {
+                    continue;
+                }
+                expandSnapshotSizeFromSubtree(child, child.getLeft(), child.getTop(), expanded);
+            }
+            width = expanded[0];
+            height = expanded[1];
+        }
+
+        if (width <= 0) {
+            width = 100;
+        }
+        if (height <= 0) {
+            height = 100;
+        }
 
         // Do not create the doublebuffer-bitmap each time. reuse it to save memory.
         Bitmap bitmap = mLastBitmapCreated;


swingler-andavo (approved):

Ishan5hrestha:
This is by far most important fix for map apps. Thank You! Hope this will be merged soon!

Ishan5hrestha:
@bkingery really waiting for this PR merge

bkingery:
@Ishan5hrestha The PR requires review from the project maintainers. Until then you could apply this patch in your project.

```patch
diff --git a/node_modules/react-native-maps/android/src/main/java/com/rnmaps/maps/MapMarker.java b/node_modules/react-native-maps/android/src/main/java/com/rnmaps/maps/MapMarker.java
--- a/node_modules/react-native-maps/android/src/main/java/com/rnmaps/maps/MapMarker.java
+++ b/node_modules/react-native-maps/android/src/main/java/com/rnmaps/maps/MapMarker.java
@@ -9,6 +9,7 @@
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
 import android.view.View;
+import android.view.ViewGroup;
 import android.widget.LinearLayout;
 import android.animation.ObjectAnimator;
 import android.util.Property;
@@ -519,11 +520,60 @@
 
   private void clearDrawableCache() {
     mLastBitmapCreated = null;
+  }
+
+  /**
+   * Expand bitmap bounds using sizes already produced by RN layout. Do not call {@link
+   * View#measure} on Catalyst/Fabric views — that triggers "must be measured before being added to
+   * a parent" (explicit MeasureSpec assertion).
+   */
+  private static void expandSnapshotSizeFromSubtree(View root, int[] maxWh) {
+    if (root == null || root.getVisibility() == View.GONE) {
+      return;
+    }
+    int w = root.getWidth();
+    int h = root.getHeight();
+    if (w <= 0) {
+      w = root.getMeasuredWidth();
+    }
+    if (h <= 0) {
+      h = root.getMeasuredHeight();
+    }
+    if (w > 0) {
+      maxWh[0] = Math.max(maxWh[0], w);
+    }
+    if (h > 0) {
+      maxWh[1] = Math.max(maxWh[1], h);
+    }
+    if (root instanceof ViewGroup) {
+      ViewGroup vg = (ViewGroup) root;
+      for (int i = 0; i < vg.getChildCount(); i++) {
+        View child = vg.getChildAt(i);
+        if (child instanceof MapCallout) {
+          continue;
+        }
+        expandSnapshotSizeFromSubtree(child, maxWh);
+      }
+    }
   }
 
   private Bitmap createDrawable() {
     int width = this.width <= 0 ? 100 : this.width;
     int height = this.height <= 0 ? 100 : this.height;
+    // Yoga can report a small marker frame while children laid out larger — grow bitmap using
+    // laid-out dimensions only (Fabric-safe; see expandSnapshotSizeFromSubtree).
+    if (hasCustomMarkerView && getChildCount() > 0) {
+      int[] maxWh = new int[] {width, height};
+      for (int i = 0; i < getChildCount(); i++) {
+        View child = getChildAt(i);
+        if (child instanceof MapCallout) {
+          continue;
+        }
+        expandSnapshotSizeFromSubtree(child, maxWh);
+      }
+      width = maxWh[0];
+      height = maxWh[1];
+    }
     this.buildDrawingCache();
 
     // Do not create the doublebuffer-bitmap each time. reuse it to save memory.
```

Ishan5hrestha:
Still relevant

salah-ghanim (approved):

PR #5923, opened 2026-06-02
feat(ios): allow GoogleMaps pod versions to be overridden

MathieuWhite:

No.

Provides the ability to hold off on GoogleMaps for iOS SDK updates or update ahead of time without being blocked by this package.

### How did you test this PR?

Tested on iOS by updating the example app's Podfile and re-installing pods. 
Verified the default values are used if the Podfile does not provide the variables.
Using the current versions as defaults for backward compatibility.


salah-ghanim (approved):

ChromeQ:
Will this be exposed via the expo plugin to provide config?

PR #5937, opened 2026-06-08
fix(Android): paint and remove New-Arch Polygon/Polyline overlays reliably (subset of #5932)

### LINKED ISSUE #5932: [New Architecture][Android] Polygon/Polyline/Geojson overlays don't paint on add, don't repaint after camera move, and ghost on removal
﻿## Summary
On Fabric / Android with `provider="google"`, polygon/polyline/geojson overlays behave unreliably:
- A freshly-added overlay often doesn't paint until a property mutates or the user zooms.
- Removing overlays (e.g. toggling a layer off) can leave "ghost" shapes painted on the surface.
- Overlays that survive a camera move aren't repainted.

Markers are unaffected.

## Environment
- react-native-maps: 1.27.2
- Platform: Android, `provider="google"`
- New Architecture (Fabric): enabled
- React Native 0.83.x / Expo SDK 55

## Reproduction
With the New Architecture enabled, render a handful of `<Geojson>` polygons:
```jsx
<MapView provider="google" style={{ flex: 1 }} initialRegion={/* ... */}>
  <Geojson geojson={featureCollectionWithAFewPolygons} fillColor="rgba(0,128,255,0.4)" strokeColor="#0066ff" />
</MapView>
```
- On first render the polygons frequently don't appear until you pinch-zoom.
- Conditionally unmount the `<Geojson>` â†’ one or more polygons remain painted ("ghosts").

## Root cause
Under the New Architecture, the GoogleMap surface does not composite freshly-added polygon/polyline overlays until a property changes, and does not repaint unchanged overlays after a camera move. Removal is also unreliable: feature removal is index-based (`MapView.removeFeatureAt`), and the per-view Fabric teardown is not dependably invoked for these reparented children, so removed overlays can linger.

## Suggested fix (approach)
- Re-commit overlays (re-apply their points) once the map is loaded and again on camera-idle, so overlays added before the surface was ready actually paint.
- Nudge each newly added overlay on the next frame to force its first paint.
- Remove overlays by identity when the view is dropped/recycled (override `onDropViewInstance` in the Fabric polygon/polyline managers â†’ remove the GoogleMap overlay directly) instead of relying solely on the index-based path.
- Optionally a deterministic reconcile: a prop carrying the identifiers that should be on the map, with the native side removing any tagged overlay whose id is absent â€” a reliable backstop for the unreliable Fabric teardown.

A complete, working `patch-package` patch for 1.27.2 implementing the above is available; happy to open a PR.

Related: #3386 (older Android Geojson fill-on-zoom), #5355 (Fabric support).

ThomasPonzoVisma:
Fixes #5932 (partial — focused subset)

## Problem
On the New Architecture (Android, `provider="google"`), `<Polygon>` / `<Polyline>` / `<Geojson>` overlays:
- often don't paint until a property mutates or the user zooms,
- aren't repainted after a camera move, and
- can linger ("ghost") on removal.

Markers are unaffected.

## Fix (focused subset)
- **Paint:** re-commit overlays (re-apply their points — visually a no-op) on map-load and on camera-idle, and nudge each newly added overlay on the next frame, so they paint without needing a zoom. (`MapView.redrawAllOverlays`, `MapPolygon`/`MapPolyline.forceRedraw`)
- **Removal:** remove the overlay by identity when Fabric drops/recycles the view (`PolygonManager`/`PolylineManager.onDropViewInstance` → `removeFromMapDirectly`), instead of relying solely on the index-based `removeFeatureAt` path (unreliable for batches under Fabric).

## Scope / caveats
- This is a **subset** of the full fix for #5932. It deliberately omits a `visible`-prop / deterministic `overlayReconcileJSON` backstop, because those require **spec + codegen** changes; this PR stays within the hand-editable Java so it's easy to review.
- **Not build-verified in this repo** — the changes are ported from a working `patch-package` patch validated in an app against 1.27.2 (the same source as `master`). Please run CI / a device check.

### DIFF
--- a/android/src/main/java/com/rnmaps/fabric/PolygonManager.java
+++ b/android/src/main/java/com/rnmaps/fabric/PolygonManager.java
@@ -40,6 +40,15 @@ public class PolygonManager extends ViewGroupManager<MapPolygon> implements RNMa
         return new MapPolygon(context);
     }
 
+    @Override
+    public void onDropViewInstance(MapPolygon view) {
+        // Identity-based overlay removal when Fabric drops/recycles the view, so a
+        // removed overlay can't linger as a "ghost" under the New Architecture (the
+        // index-based removeFeatureAt path is unreliable for batches).
+        view.removeFromMapDirectly();
+        super.onDropViewInstance(view);
+    }
+
 
 
     public static final String REACT_CLASS = "RNMapsGooglePolygon";
--- a/android/src/main/java/com/rnmaps/fabric/PolylineManager.java
+++ b/android/src/main/java/com/rnmaps/fabric/PolylineManager.java
@@ -50,6 +50,13 @@ public class PolylineManager extends ViewGroupManager<MapPolyline> implements RN
         return new MapPolyline(context);
     }
 
+    @Override
+    public void onDropViewInstance(MapPolyline view) {
+        // Identity-based overlay removal — see PolygonManager#onDropViewInstance.
+        view.removeFromMapDirectly();
+        super.onDropViewInstance(view);
+    }
+
 
     public static final String REACT_CLASS = "RNMapsPolyline";
 
--- a/android/src/main/java/com/rnmaps/maps/MapPolygon.java
+++ b/android/src/main/java/com/rnmaps/maps/MapPolygon.java
@@ -218,7 +218,33 @@ public class MapPolygon extends MapFeature {
 
   @Override
   public void removeFromMap(Object collection) {
+    if (polygon == null) return;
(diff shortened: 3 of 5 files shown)


samzmann:
Applied this as a patch on 1.26.20 and it did fix the ghost polygon issue (polygon lines still visible after unmount). Thanks.

samzmann:
The whole set of PRs seems AI generated though, what does the human have to say @ThomasPonzoVisma  ?

ThomasPonzoVisma:
@samzmann that we need better support for new arch officially without needing patches

PR #5938, opened 2026-06-09
fix(ios): replace RCTImageLoader with NSURLSession for bridgeless mode compatibility (#5912)

### LINKED ISSUE #5912: [iOS] [Apple Maps] Marker image prop fails to load under New Architecture (bridgeless mode) in RN >= 0.82
### Summary

Since React Native 0.82+ (https://github.com/facebook/react-native/pull/53026) bridgeless mode is enabled by default, both self.bridge (on view managers) and [RCTBridge currentBridge] return nil. This causes marker images set via the image prop to silently fail to load on iOS, because the lookup chain is sent to nil and the entire expression evaluates to nil.

### Affected code:
ios/AirMaps/AIRMapMarker.m in setImageSrc:
```obj-c
_reloadImageCancellationBlock = [[[RCTBridge currentBridge] moduleForName:@"ImageLoader"] loadImageWithURLRequest:...
```

### Root cause:

self.bridge and RCTBridge.currentBridge return nil under bridgeless mode. The Fabric component view RNMapsMarkerView does have access to RCTImageLoader via its module registry.

### Reproducible sample code

```TSX
<Marker
     coordinate={{ latitude: 37.78, longitude: -122.41 }}
     image={require('./assets/flag-pink.png'}
/>
```

### Steps to reproduce

1. React Native 0.82+ app with New Architecture enabled (the default)
2. Install react-native-maps
3. Render a <Marker> with a remote or local image prop

### Expected result

Marker renders with the image specified by the image prop, as it does under the old architecture.

### Actual result

No marker renders on the map at all

### React Native Maps Version

1.27.2

iOS (Apple Maps)

### React Native Version

\>= 0.82.0

Not using Expo

### Device(s)

iPhone 14 Pro (26.3.1) 

### Additional information

_No response_

mikkoseppa-tommy:
No

In bridgeless mode (New Architecture, RN >= 0.82), `[RCTBridge currentBridge]`
returns `nil`, causing marker images to silently fail to load. This affects any
app running the New Architecture with image-based markers.

Fixes #5912

### Solution

Replace the `RCTImageLoader`-based image loading in `AIRMapMarker` with a
standalone `NSURLSession` implementation.

The image URI is already fully resolved by the JS side before it reaches native
(local assets become `file://` URIs, Metro dev server assets become
`http://localhost:8081/...` URIs), so `RCTImageLoader` is not needed.

The replacement implementation:
- Uses `NSURLSession` directly to fetch the resolved URI
- Coalesces concurrent requests for the same URL so multiple markers sharing
  one image asset fire only a single network request
- Caches results in an `NSCache`, which evicts automatically under memory
  pressure
- Cancellation removes the caller's interest without aborting the underlying
  request, which other waiting markers may still need

The now-unused `bridge` property is removed from `AIRMapMarker`.

### How did you test this PR?
- Platforms affected: Affects only iOS.
- Tested both on a simulator and on a real device.
- Stress tested with 10 000+ markers rendered on screen.

### Future work
The proper long-term fix is to migrate `AIRMapMarker` to a true Fabric native
component. Fabric components declare their props in a typed component spec and
receive images through the Fabric renderer's built-in image pipeline, the same
way React Native's own `<Image>` component works. This patch is a pragmatic fix
to unblock bridgeless mode users without the scope of a full Fabric migration.

### DIFF
--- a/ios/AirMaps/AIRMapMarker.m
+++ b/ios/AirMaps/AIRMapMarker.m
@@ -9,22 +9,78 @@
 
 #import "AIRMapMarker.h"
 
-#import <React/RCTBridge.h>
-#import <React/RCTEventDispatcher.h>
-#import <React/RCTImageLoaderProtocol.h>
 #import <React/RCTUtils.h>
 #import <React/UIView+React.h>
-#import <React/RCTImageLoader.h>
-#import <React/RCTBridge+Private.h>
 
 NSInteger const AIR_CALLOUT_OPEN_ZINDEX_BASELINE = 999;
+
+// ---------------------------------------------------------------------------
+// Image loading — coalescing cache
+//
+// Multiple markers sharing the same image URL (common case: all markers use
+// the same pin asset) would each fire an independent NSURLSession request
+// without this. The coalescer ensures exactly one in-flight request per URL.
+// All callers that arrive while a request is in-flight are queued and notified
+// together when it completes. Results are cached in memory for the process
+// lifetime so subsequent mounts are instant.
+//
+// All access to these dictionaries happens on the main thread:
+//   - setImageSrc: is called from React Native's UI thread
+//   - the NSURLSession completion block dispatches back to the main queue
+// ---------------------------------------------------------------------------
+static NSCache<NSString *, UIImage *>                                          *AIRImageCache;
+static NSMutableDictionary<NSString *, NSMutableArray<void (^)(UIImage *)> *>  *AIRPendingHandlers;
+
+// Requests the image at urlString, returning a cancel block.
+// "Cancel" means this caller no longer wants the result — it does NOT abort
+// the network request, which other waiting callers still need.
+static dispatch_block_t AIRLoadImage(NSString *urlString, CGFloat scale, void (^completion)(UIImage *)) {
+    if (!AIRImageCache)      AIRImageCache      = [NSCache new];
+    if (!AIRPendingHandlers) AIRPendingHandlers = [NSMutableDictionary new];
+
+    // Cache hit — deliver on next run-loop tick so callers always get async behaviour
+    UIImage *cached = [AIRImageCache objectForKey:urlString];
+    if (cached) {
+        dispatch_async(dispatch_get_main_queue(), ^{ completion(cached); });
+        return ^{};
+    }
+
+    // Coalesce: register callback and let the existing in-flight request deliver it
+    __block BOOL cancelled = NO;
+    void (^handler)(UIImage *) = ^(UIImage *image) { if (!cancelled) completion(image); };
+
+    NSMutableArray *handlers = AIRPendingHandlers[urlString];
+    if (handlers) {
+        [handlers addObject:handler];
+        return ^{ cancelled = YES; };
+    }
+
+    // First request for this URL
+    AIRPendingHandlers[urlString] = [NSMutableArray arrayWithObject:handler];
+
+    NSURL *url = [NSURL URLWithString:urlString];
+    [[[NSURLSession sharedSession]
+        dataTaskWithURL:url
+      completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
+        UIImage *image = (!error && data) ? [UIImage imageWithData:data scale:scale] : nil;
+        if (!image) NSLog(@"AIRMapMarker failed to load image: %@", error);
+        dispatch_async(dispatch_get_main_queue(), ^{
+            if (image) [AIRImageCache setObject:image forKey:urlString];
+            NSArray *pending = AIRPendingHandlers[urlString];
+            [AIRPendingHandlers removeObjectForKey:urlString];
+            for (void (^h)(UIImage *) in pending) h(image);
+        });
+    }] resume];
+
+    return ^{ cancelled = YES; };
+}
 
 @implementation AIREmptyCalloutBackgroundView
 @end
 
 @implementation AIRMapMarker {
     BOOL _hasSetCalloutOffset;
-    RCTImageLoaderCancellationBlock _reloadImageCancellationBlock;
+    dispatch_block_t _imageLoadCancel;
     MKMarkerAnnotationView *_markerView;
     MKPinAnnotationView *_pinView;
     BOOL _calloutIsOpen;
@@ -362,29 +418,17 @@ NSInteger const AIR_CALLOUT_OPEN_ZINDEX_BASELINE = 999;
 {
     _imageSrc = imageSrc;
 
-    if (_reloadImageCancellationBlock) {
-        _reloadImageCancellationBlock();
-        _reloadImageCancellationBlock = nil;
+    // Deregister from any previous in-flight load
+    if (_imageLoadCancel) {
+        _imageLoadCancel();
+        _imageLoadCancel = nil;
     }
-    __weak __typeof(self) weakSelf = self;
+    if (!imageSrc) return;
 
-    _reloadImageCancellationBlock = [[[RCTBridge currentBridge] moduleForName:@"ImageLoader"] loadImageWithURLRequest:[RCTConvert NSURLRequest:_imageSrc]
-                                                                                                                 size:self.bounds.size
-                                                                                                                scale:RCTScreenScale()
-                                                                                                              clipped:YES
-                                                                                                           resizeMode:RCTResizeModeCenter
-                                                                                                        progressBlock:nil
-                                                                                                     partialLoadBlock:nil
-                                                                                                      completionBlock:^(NSError *error, UIImage *image) {
-        if (error) {
-            // TODO(lmr): do something with the error?
-            NSLog(@"failed to load image: %@", error);
-        }
-        dispatch_async(dispatch_get_main_queue(), ^{
-            __strong __typeof(weakSelf) strongSelf = weakSelf;
-            strongSelf.image = image;
-        });
-    }];
+    __weak __typeof(self) weakSelf = self;
+    _imageLoadCancel = AIRLoadImage(imageSrc, RCTScreenScale(), ^(UIImage *image) {
+        weakSelf.image = image;
+    });
 }
 
 - (void)setPinColor:(UIColor *)pinColor
--- a/ios/AirMaps/AIRMapMarker.m
+++ b/ios/AirMaps/AIRMapMarker.m
@@ -31,13 +31,34 @@ NSInteger const AIR_CALLOUT_OPEN_ZINDEX_BASELINE = 999;
 static NSCache<NSString *, UIImage *>                                          *AIRImageCache;
 static NSMutableDictionary<NSString *, NSMutableArray<void (^)(UIImage *)> *>  *AIRPendingHandlers;
 
+// True for URIs that came out of RN's asset pipeline rather than an arbitrary
+// remote URL: an on-device bundled asset (file:// under the app's bundle,
+// library, or home directory — what RCTIsLocalAssetURL already recognizes
+// for RN's own image loaders) or the Metro dev-server form produced by
+// AssetSourceResolver's assetServerURL(), e.g. ".../assets/Foo/icon@2x.png
+// ?platform=ios&hash=...".
+static BOOL AIRIsPackagerAssetURL(NSURL *url) {
+    if (RCTIsLocalAssetURL(url)) {
+        return YES;
+    }
+    NSString *query = url.query;
+    return url.path != nil && [url.path containsString:@"/assets/"]
+        && query != nil && [query containsString:@"hash="];
+}
+
 // React Native's asset resolver bakes the chosen density into the filename
 // (e.g. "pin@2x.png"), omitting the suffix for 1x. That resolved density is
-// what the pixel data actually is, so it — not the screen's scale — is what
-// UIImage must be decoded with.
+// what the pixel data actually is, so for RN assets it — not the screen's
+// scale — is what UIImage must be decoded with. Arbitrary remote URLs carry
+// no such baked-in density, so they keep using the screen's scale as before.
 static CGFloat AIRScaleFromAssetURL(NSString *urlString) {
+    NSURL *url = [NSURL URLWithString:urlString];
+    if (!AIRIsPackagerAssetURL(url)) {
+        return RCTScreenScale();
+    }
+
     CGFloat scale = 1;
-    NSString *filename = [NSURL URLWithString:urlString].path.lastPathComponent;
+    NSString *filename = url.path.lastPathComponent;
     NSRange at = [filename rangeOfString:@"@" options:NSBackwardsSearch];
     if (at.location != NSNotFound) {
         NSString *suffix = [filename substringFromIndex:at.location + 1];
@@ -445,6 +466,15 @@ static dispatch_block_t AIRLoadImage(NSString *urlString, CGFloat scale, void (^
         _imageLoadCancel = nil;
     }
     if (!imageSrc) return;
+
+    // A bare name with no URL scheme (e.g. image={{uri: 'custom_pin'}}) refers
+    // to an Xcode asset-catalog image, not something NSURLSession can fetch.
+    // Load it synchronously via the catalog instead, as RCTLocalAssetImageLoader
+    // did before this component moved off RCTImageLoader.
+    if ([NSURL URLWithString:imageSrc].scheme == nil) {
+        self.image = [UIImage imageNamed:imageSrc];
+        return;
+    }
 
     __weak __typeof(self) weakSelf = self;
     _imageLoadCancel = AIRLoadImage(imageSrc, AIRScaleFromAssetURL(imageSrc), ^(UIImage *image) {
--- a/ios/AirMaps/AIRMapMarker.m
+++ b/ios/AirMaps/AIRMapMarker.m
@@ -31,6 +31,27 @@ NSInteger const AIR_CALLOUT_OPEN_ZINDEX_BASELINE = 999;
 static NSCache<NSString *, UIImage *>                                          *AIRImageCache;
 static NSMutableDictionary<NSString *, NSMutableArray<void (^)(UIImage *)> *>  *AIRPendingHandlers;
 
+// React Native's asset resolver bakes the chosen density into the filename
+// (e.g. "pin@2x.png"), omitting the suffix for 1x. That resolved density is
+// what the pixel data actually is, so it — not the screen's scale — is what
+// UIImage must be decoded with.
+static CGFloat AIRScaleFromAssetURL(NSString *urlString) {
+    CGFloat scale = 1;
+    NSString *filename = [NSURL URLWithString:urlString].path.lastPathComponent;
+    NSRange at = [filename rangeOfString:@"@" options:NSBackwardsSearch];
+    if (at.location != NSNotFound) {
+        NSString *suffix = [filename substringFromIndex:at.location + 1];
+        NSRange x = [suffix rangeOfString:@"x"];
+        if (x.location != NSNotFound) {
+            CGFloat parsed = [[suffix substringToIndex:x.location] doubleValue];
+            if (parsed > 0) {
+                scale = parsed;
+            }
+        }
+    }
+    return scale;
+}
+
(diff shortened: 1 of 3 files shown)


brianp-tommycarwash (approved):

simon-thuresson-md:
We are currently experiencing this issue and would love to get this reviewed and merged

brianp-tommycarwash:
@salah-ghanim We've got all green. Can you please review and merge this in.

NicholasBoccuzzi:
Hit #5912 on RN 0.86 with the New Architecture and Apple Maps (react-native-maps 1.27.2) and independently landed on the same `NSURLSession` approach, so this matches what we ended up carrying as a local patch. The request coalescing and `NSCache` here are nicer than what I wrote.

One thing worth changing before merge: the decode scale.

This PR passes `RCTScreenScale()` for every image:

```objc
_imageLoadCancel = AIRLoadImage(imageSrc, RCTScreenScale(), ^(UIImage *image) {
```

But the URI that reaches native has already been resolved to one *specific* density by `AssetSourceResolver`, which encodes that density in the filename ([`AssetSourceResolver.js`](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Image/AssetSourceResolver.js)):

```js
const scaleSuffix = scale === 1 ? '' : '@' + scale + 'x';
```

So the scale that matters is the one in the resolved filename, not the screen's. Those coincide only when the asset happens to ship the screen's density.

The common failure is an asset that ships a single resolution: `pin.png` resolves to the 1x file, `RCTScreenScale()` returns 3 on most devices, and `[UIImage imageWithData:scale:3]` produces a `UIImage.size` one third of the intended points — so the marker draws at a third of its size. `MKAnnotationView` sizes itself from `image.size`, and `AIRMapMarker`'s `reactSetFrame` reads it too, so the anchor drifts along with it.

Deriving it from the filename instead, defaulting to 1 when there's no suffix:

```objc
static CGFloat AIRScaleFromAssetURL(NSString *urlString) {
    // RN omits the suffix for 1x, so "no suffix" means scale 1 — not the screen scale.
    CGFloat scale = 1;
    NSString *filename = [NSURL URLWithString:urlString].path.lastPathComponent;
    NSRange at = [filename rangeOfString:@"@" options:NSBackwardsSearch];
    if (at.location != NSNotFound) {
        NSString *suffix = [filename substringFromIndex:at.location + 1];
        NSRange x = [suffix rangeOfString:@"x"];
        if (x.location != NSNotFound) {
            CGFloat parsed = [[suffix substringToIndex:x.location] doubleValue];
            if (parsed > 0) {
                scale = parsed;
            }
        }
    }
    return scale;
}
```

I had the same bug in our patch with a `RCTScreenScale()` fallback and only caught it while comparing against this PR — it's invisible if your assets ship every density, which is why it's easy to miss.

Happy to open a follow-up PR against your branch if that's easier than folding it in.

mikkoseppa-tommy:
> Hit #5912 on RN 0.86 with the New Architecture and Apple Maps (react-native-maps 1.27.2) and independently landed on the same `NSURLSession` approach, so this matches what we ended up carrying as a local patch. The request coalescing and `NSCache` here are nicer than what I wrote.
> 
> One thing worth changing before merge: the decode scale.
> 
> This PR passes `RCTScreenScale()` for every image:
> 
> ```objc
> _imageLoadCancel = AIRLoadImage(imageSrc, RCTScreenScale(), ^(UIImage *image) {
> ```
> 
> But the URI that reaches native has already been resolved to one _specific_ density by `AssetSourceResolver`, which encodes that density in the filename ([`AssetSourceResolver.js`](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Image/AssetSourceResolver.js)):
> 
> ```js
> const scaleSuffix = scale === 1 ? '' : '@' + scale + 'x';
> ```
> 
> So the scale that matters is the one in the resolved filename, not the screen's. Those coincide only when the asset happens to ship the screen's density.
> 
> The common failure is an asset that ships a single resolution: `pin.png` resolves to the 1x file, `RCTScreenScale()` returns 3 on most devices, and `[UIImage imageWithData:scale:3]` produces a `UIImage.size` one third of the intended points — so the marker draws at a third of its size. `MKAnnotationView` sizes itself from `image.size`, and `AIRMapMarker`'s `reactSetFrame` reads it too, so the anchor drifts along with it.
> 
> Deriving it from the filename instead, defaulting to 1 when there's no suffix:
> 
> ```objc
> static CGFloat AIRScaleFromAssetURL(NSString *urlString) {
>     // RN omits the suffix for 1x, so "no suffix" means scale 1 — not the screen scale.
>     CGFloat scale = 1;
>     NSString *filename = [NSURL URLWithString:urlString].path.lastPathComponent;
>     NSRange at = [filename rangeOfString:@"@" options:NSBackwardsSearch];
>     if (at.location != NSNotFound) {
>         NSString *suffix = [filename substringFromIndex:at.location + 1];
>         NSRange x = [suffix rangeOfString:@"x"];
>         if (x.location != NSNotFound) {
>             CGFloat parsed = [[suffix substringToIndex:x.location] doubleValue];
>             if (parsed > 0) {
>                 scale = parsed;
>             }
>         }
>     }
>     return scale;
> }
> ```
> 
> I had the same bug in our patch with a `RCTScreenScale()` fallback and only caught it while comparing against this PR — it's invisible if your assets ship every density, which is why it's easy to miss.
> 
> Happy to open a follow-up PR against your branch if that's easier than folding it in.

Thank you for the feedback. This is now fixed in the latest commit.

ios/AirMaps/AIRMapMarker.m
```diff
@@ -362,29 +439,17 @@ - (void)setImageSrc:(NSString *)imageSrc
...
-            // TODO(lmr): do something with the error?
-            NSLog(@"failed to load image: %@", error);
-        }
-        dispatch_async(dispatch_get_main_queue(), ^{
-            __strong __typeof(weakSelf) strongSelf = weakSelf;
-            strongSelf.image = image;
-        });
-    }];
+    __weak __typeof(self) weakSelf = self;
+    _imageLoadCancel = AIRLoadImage(imageSrc, AIRScaleFromAssetURL(imageSrc), ^(UIImage *image) {
```
brianp-tommycarwash, line 480: `AIRScaleFromAssetURL` is applied to every `imageSrc`, but only bundled RN assets carry the density in the filename. For a remote URI like `https://cdn.example.com/pin.png` there is no `@Nx`, so this returns 1 where the old code returned `RCTScreenScale()`.

`reactSetFrame` derives the annotation bounds from `self.image.size` (line 130), so on a 3x device a 90x90 remote image goes from 30x30pt to 90x90pt. That's a silent 3x size jump for remote marker images.

Suggest gating on whether the URI is actually an RN asset before defaulting to 1: `file://` and the Metro `/assets/...?platform=ios&hash=...` form can use the parsed scale (defaulting to 1, which is correct for 1x-only assets), and anything else keeps `RCTScreenScale()` so existing remote-image users don't shift.

ios/AirMaps/AIRMapMarker.m
```diff
@@ -9,22 +9,99 @@
...
+static NSCache<NSString *, UIImage *>                                          *AIRImageCache;
+static NSMutableDictionary<NSString *, NSMutableArray<void (^)(UIImage *)> *>  *AIRPendingHandlers;
+
+// React Native's asset resolver bakes the chosen density into the filename
+// (e.g. "pin@2x.png"), omitting the suffix for 1x. That resolved density is
+// what the pixel data actually is, so it — not the screen's scale — is what
+// UIImage must be decoded with.
+static CGFloat AIRScaleFromAssetURL(NSString *urlString) {
+    CGFloat scale = 1;
+    NSString *filename = [NSURL URLWithString:urlString].path.lastPathComponent;
```
brianp-tommycarwash: If `NSURL URLWithString:` returns nil — unescaped characters in the path, a literal space in an asset name — `filename` is nil and the checks below fall through to `scale = 1` only because messaging nil zeroes out the returned `NSRange`. That works, but it's accidental rather than deliberate. Worth an explicit guard so the fallback is a decision:

```objc
NSString *filename = [NSURL URLWithString:urlString].path.lastPathComponent;
if (filename.length == 0) return 1;
```

brianp-tommycarwash (changes requested):
Re-reviewing after 23dbda8d. The scale change is right for bundled assets but it also changes behavior for remote image URIs, and I don't think that part is intended.

Decoding at the asset's baked-in density is correct: RN's resolver picks `pin@2x.png` and the pixel data really is 2x, so `RCTScreenScale()` was wrong any time the picked density didn't match the screen. No argument there.

The issue is that `AIRScaleFromAssetURL` runs on every `imageSrc`, including plain remote URLs. `<Marker image={{uri: 'https://cdn.example.com/pin.png'}}>` has no `@Nx` suffix, so it now decodes at scale 1 instead of screen scale. Since `reactSetFrame` sizes the annotation from `self.image.size`, a 90x90 remote pin that drew 30x30pt on a 3x device now draws 90x90pt. That's a visible regression for anyone using remote marker images — master used screen scale here too, so it's not just a delta against the first commit.

Note you can't simply fall back to `RCTScreenScale()` when there's no suffix: a bundled asset that only ships 1x resolves to `pin.png` with genuine 1x data, and screen scale would render it a third of the intended size. So the fallback has to key off whether the URI is an RN asset versus an arbitrary remote URL.

- ios/AirMaps/AIRMapMarker.m:450 — remote URIs with no `@Nx` suffix now decode at scale 1, so those markers render 2-3x larger than before
- ios/AirMaps/AIRMapMarker.m:40 — a nil `NSURL` degrades to scale 1 by accident rather than by decision

Everything else still looks good to me — coalescing, the NSCache, and the cancel-means-deregister semantics all read correctly. Happy to re-approve once the remote case is either handled or called out in the description as an intentional break.

mikkoseppa-tommy:
@brianp-tommycarwash @NicholasBoccuzzi

I have update the PR with the following changes:
- Gated asset-scale decoding to packager URLs (bundled assets via RCTIsLocalAssetURL(), Metro dev-server asset form) — fixes remote image URIs regressing to scale 1
- Nil/unparseable URLs now explicitly fall back to RCTScreenScale() instead of accidentally landing on scale 1
- Added UIImage imageNamed: fallback for schemeless imageSrc values (e.g. image={{uri: 'custom_pin'}}) to support - Xcode asset-catalog images

Tested with:
- React Native assets with different scales
- Web URLs
- Xcode assets

Everything seems to be working

Fgabz:
Hey, we hit the same thing but on Google Maps, not Apple.

Same root cause, but this PR only fixes `ios/AirMaps/AIRMapMarker.m`. The Google side has the same call twice, in `AIRGoogleMapMarker.m` — `setImageSrc:` (L369) and `setIconSrc:` (L436). Those two plus yours are the only `currentBridge` uses in the lib.

We're on RN 0.85.3 / Expo SDK 56, new arch, GoogleMaps pod 9.4.0, iOS 26.5 sim and a real device.

The symptom is different depending on the prop, which is confusing:

- `image={require(...)}` → nothing on the map at all, not even the default pin. The code inserts an empty placeholder view as `iconView` before loading, so the SDK thinks there's a custom icon and doesn't fall back.
- `icon={require(...)}` → default red pin, the image never lands.
- no props → default pin, fine.

Took us a while because there is no error, no log, no network request. Everything is just sent to nil.

Stuff we checked that is NOT the cause: marker mounting, component registration, provider, asset resolution, `tracksViewChanges`, the GoogleMaps pod version (we're on 9.4.0 so it's not the 10.x thing from #5964), and props not arriving — `pinColor` is applied in the same `updateProps` block and works, so `setImageSrc:` really is called.

I applied your fix to the Google file for both call sites and it works. Pins are back, and swapping the icon on focus works too.

Two notes if you want to take it:

- I renamed the statics `AIRGoogle*`, because if both files get the same helpers you end up with duplicate symbols. Probably nicer in a shared header.
- the Google `setImageSrc:` does more in the completion (frame, `layoutSubviews`, `setIconView:`) so I kept that part and only swapped the loader.

Happy to open a PR if that helps.

mikkoseppa-tommy:
> Hey, we hit the same thing but on Google Maps, not Apple.
> 
> Same root cause, but this PR only fixes `ios/AirMaps/AIRMapMarker.m`. The Google side has the same call twice, in `AIRGoogleMapMarker.m` — `setImageSrc:` (L369) and `setIconSrc:` (L436). Those two plus yours are the only `currentBridge` uses in the lib.
> 
> We're on RN 0.85.3 / Expo SDK 56, new arch, GoogleMaps pod 9.4.0, iOS 26.5 sim and a real device.
> 
> The symptom is different depending on the prop, which is confusing:
> 
> * `image={require(...)}` → nothing on the map at all, not even the default pin. The code inserts an empty placeholder view as `iconView` before loading, so the SDK thinks there's a custom icon and doesn't fall back.
> * `icon={require(...)}` → default red pin, the image never lands.
> * no props → default pin, fine.
> 
> Took us a while because there is no error, no log, no network request. Everything is just sent to nil.
> 
> Stuff we checked that is NOT the cause: marker mounting, component registration, provider, asset resolution, `tracksViewChanges`, the GoogleMaps pod version (we're on 9.4.0 so it's not the 10.x thing from #5964), and props not arriving — `pinColor` is applied in the same `updateProps` block and works, so `setImageSrc:` really is called.
> 
> I applied your fix to the Google file for both call sites and it works. Pins are back, and swapping the icon on focus works too.
> 
> Two notes if you want to take it:
> 
> * I renamed the statics `AIRGoogle*`, because if both files get the same helpers you end up with duplicate symbols. Probably nicer in a shared header.
> * the Google `setImageSrc:` does more in the completion (frame, `layoutSubviews`, `setIconView:`) so I kept that part and only swapped the loader.
> 
> Happy to open a PR if that helps.

Hey, I'm glad to hear this worked for you. If you can create a PR for the Google Maps changes that would be helpful.

ios/AirMaps/AIRMapMarker.m
```diff
@@ -9,22 +9,120 @@
...
+static NSMutableDictionary<NSString *, NSMutableArray<void (^)(UIImage *)> *>  *AIRPendingHandlers;
+
+// True for URIs that came out of RN's asset pipeline rather than an arbitrary
+// remote URL: an on-device bundled asset (file:// under the app's bundle,
+// library, or home directory — what RCTIsLocalAssetURL already recognizes
+// for RN's own image loaders) or the Metro dev-server form produced by
+// AssetSourceResolver's assetServerURL(), e.g. ".../assets/Foo/icon@2x.png
+// ?platform=ios&hash=...".
+static BOOL AIRIsPackagerAssetURL(NSURL *url) {
+    if (RCTIsLocalAssetURL(url)) {
```
brianp-tommycarwash, line 41: `RCTIsLocalAssetURL` isn't an "is this an RN asset" predicate — it's an image-content-type predicate wearing that name. In RN 0.85 (`React/Base/RCTUtils.mm`):

```objc
static BOOL RCTIsImageAssetsPath(NSString *path) {
  NSString *extension = [path pathExtension];
  return [extension isEqualToString:@"png"] || [extension isEqualToString:@"jpg"];
}
BOOL RCTIsBundleAssetURL(NSURL *url) { return RCTIsImageAssetsPath(RCTBundlePathForURL(url)); }
BOOL RCTIsLocalAssetURL(NSURL *url) { return RCTIsBundleAssetURL(url) || RCTIsHomeAssetURL(url); }
```

Only `png` and `jpg` pass. Metro's default `assetExts` also ships `jpeg`, `gif`, `webp` and `bmp`, so a bundled `pin@2x.jpeg` is a real asset that this returns `NO` for.

That matters only in release, which is what makes it easy to miss. In dev the URL is `http://localhost:8081/assets/.../pin@2x.jpeg?platform=ios&hash=…`, so the second clause catches it via `hash=`. In release `scaledAssetURLNearBundle()` produces `file:///…/MyApp.app/assets/…/pin@2x.jpeg` — no query at all, so `query != nil` fails and the whole function returns `NO`. Result is `RCTScreenScale()` on an asset whose density is baked into the filename: `pin@2x.jpeg` on a 3x device decodes at 3 and draws two-thirds size, and a 1x-only `pin.gif` draws at a third. Same failure 23dbda8d fixed, just narrowed to the non-png/jpg extensions, and it passes dev testing.

Testing origin directly rather than borrowing RN's predicate covers every extension:

```objc
static BOOL AIRIsPackagerAssetURL(NSURL *url) {
    if (url.isFileURL) {
        NSString *path = url.path;
        return [path hasPrefix:[NSBundle mainBundle].resourcePath] || [path hasPrefix:NSHomeDirectory()];
    }
    NSString *query = url.query;
    return url.path != nil && [url.path containsString:@"/assets/"]
        && query != nil && [query containsString:@"hash="];
}
```

Worth noting while you're in here: for the `file://` case RN doesn't decode by hand at all — `RCTImageFromLocalAssetURL` falls through to `[UIImage imageWithContentsOfFile:]`, and UIKit reads the `@Nx` off the filename itself. Using that for local assets would sidestep the scale question for bundled files entirely and skip an `NSURLSession` round trip for something already on disk. Not asking for it in this PR, just flagging it as the simpler end state.

One more on the comment above: it says `RCTIsLocalAssetURL` covers "bundle, library, or home" — it's bundle or home. `RCTIsLibraryAssetURL` exists but isn't part of it.

ios/AirMaps/AIRMapMarker.m
```diff
@@ -362,29 +460,26 @@ - (void)setImageSrc:(NSString *)imageSrc
...
+        _imageLoadCancel();
+        _imageLoadCancel = nil;
+    }
+    if (!imageSrc) return;
+
+    // A bare name with no URL scheme (e.g. image={{uri: 'custom_pin'}}) refers
+    // to an Xcode asset-catalog image, not something NSURLSession can fetch.
+    // Load it synchronously via the catalog instead, as RCTLocalAssetImageLoader
+    // did before this component moved off RCTImageLoader.
+    if ([NSURL URLWithString:imageSrc].scheme == nil) {
```
brianp-tommycarwash, line 474: Nit, not blocking. `scheme == nil` is true both for a bare catalog name and for a string `NSURL` refuses to parse — `URLWithString:` returns nil there, and `nil.scheme` is nil. So `{{uri: 'https://cdn.example.com/my pin.png'}}` (unescaped space) takes the catalog path, `imageNamed:` returns nil for it, and the marker silently drops its image instead of logging like the load failure path does.

Not a regression — that URL didn't work before either — but since this branch is new, distinguishing the two is nearly free:

```objc
NSURL *parsed = [NSURL URLWithString:imageSrc];
if (!parsed) {
    NSLog(@"AIRMapMarker could not parse image URI: %@", imageSrc);
    return;
}
if (parsed.scheme == nil) {
    self.image = [UIImage imageNamed:imageSrc];
    return;
}
```

brianp-tommycarwash (changes requested):
Re-reviewing 88fd6291. The remote-URI regression I raised is properly fixed, and the asset-catalog branch is a good catch I'd missed. One gap left in the packager-asset test, and it's a one-liner.

What's fixed:

- Remote URIs are back on `RCTScreenScale()`, so `<Marker image={{uri: 'https://cdn.example.com/pin.png'}}>` keeps the size it had on master. Gating on origin rather than on whether an `@Nx` suffix happens to be present is the right shape.
- The nil-`NSURL` nit is moot now, and by a better route than the guard I suggested: `setImageSrc:` returns early when the string doesn't parse to a URL with a scheme, so `AIRScaleFromAssetURL` can no longer be reached with an unparseable string at all.
- The scheme-less asset-catalog branch fixes a real hole I didn't flag — `image={{uri: 'custom_pin'}}` had no working path at all once this moved off `RCTImageLoader`, since `NSURLSession` can't fetch a bare catalog name.

What's still open:

- ios/AirMaps/AIRMapMarker.m:41 — `RCTIsLocalAssetURL` is an image-content-type test, not an origin test: it only accepts `.png` and `.jpg`. Bundled `.jpeg`/`.gif`/`.webp` assets fall through to `RCTScreenScale()` in release builds, which is the original bug this PR set out to fix.
- ios/AirMaps/AIRMapMarker.m:474 (nit, not blocking) — a URL string `NSURL` can't parse also lands in the catalog branch.

Happy to approve once the first one is handled.

PR #5940, opened 2026-06-20
fix(ios): guard Google Maps subview insertion

kaspizzo:
## Summary

Guard Google Maps iOS subview insertion against invalid native inputs before updating `_reactSubviews`.

This mirrors the defensive insertion behavior proposed for Apple Maps in #5871, but applies it to `ios/AirGoogleMaps/AIRGoogleMap.mm`, which is the crash path reported for Google Maps in #5217.

## Root cause

`insertReactSubview:atIndex:` inserted directly into `_reactSubviews` using the incoming `subview` and `atIndex`. When React Native passes a nil subview or an index outside the current array bounds during dynamic marker updates, NSMutableArray raises an exception.

## Changes

- Return early when `subview` is nil.
- Clamp `atIndex` into `[0, _reactSubviews.count]` before inserting.
- Keep the change limited to the Google Maps iOS implementation; no JS/API changes.

## Branch target

This PR targets `master` because the crash was validated against `react-native-maps@1.27.2`, and `master` currently represents that release line. If maintainers prefer this on `beta`, I can port the same guard to the v2 file layout under `ios/GoogleMaps/RNMGoogleMap.m`.

## Validation

- `yarn --frozen-lockfile`
- `yarn lint`
- `yarn tscheck`
- `yarn format-check`
- `yarn test`

I did not run the local iOS example build because this machine is missing the repo-locked Bundler version (`2.3.26`). Xcode and CocoaPods are installed, but I avoided changing the local Ruby setup just for this PR.

References #5217
Related #5871

### DIFF
--- a/ios/AirGoogleMaps/AIRGoogleMap.mm
+++ b/ios/AirGoogleMaps/AIRGoogleMap.mm
@@ -245,6 +245,10 @@ id regionAsJSON(MKCoordinateRegion region) {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wobjc-missing-super-calls"
 - (void)insertReactSubview:(id<RCTComponent>)subview atIndex:(NSInteger)atIndex {
+  if (subview == nil) {
+    return;
+  }
+
   // Our desired API is to pass up markers/overlays as children to the mapview component.
   // This is where we intercept them and do the appropriate underlying mapview action.
   if ([subview isKindOfClass:[AIRGoogleMapMarker class]]) {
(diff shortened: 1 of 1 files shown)


lejahmie:
+1 — this is the Google Maps counterpart to the crash we hit and symbolicated on Apple Maps (see #5871 for the full crash report + RCA).

Same underlying Fabric mechanism: when a `RNMapsMapView` is recycled, `prepareForRecycle` nils `_view`, so on reuse a fresh map is created with an **empty** `_reactSubviews`. Fabric then replays `mountChildComponentView:index:` with a child's original index, and the unguarded

```objc
[_reactSubviews insertObject:(UIView *)subview atIndex:(NSUInteger) atIndex];
```

throws either:
- `NSInvalidArgumentException: *** -[__NSArrayM insertObject:atIndex:]: object cannot be nil` (the `nil` case — see #5217), or
- `NSRangeException: *** -[__NSArrayM insertObject:atIndex:]: index N beyond bounds for empty array` (the out-of-range case we captured on Apple Maps).

This PR handles **both** (early `nil` return + clamp `atIndex` into `[0, _reactSubviews.count]`), which is the correct pair of guards. Removal is by object identity (`removeReactSubview:` → `removeObject:`), so clamping the insert index is behavior-preserving.

**Env:** react-native-maps 1.27.2, RN 0.85.3, New Architecture (Fabric), Hermes, iOS 26.5.1. The insert is still unguarded on `master`, so a version bump alone doesn't fix it.

We're running the equivalent change (this + #5871) as a local patch. Thanks @kaspizzo — would be great to land #5871 and #5940 together so Apple and Google paths are both covered.

salah-ghanim (approved):

PR #5950, opened 2026-07-11
fix(ios): keep google marker icon when the marker is re-added to the map

Bram-dc:

No.

On the iOS Google provider, `didInsertInMap:` creates a fresh `AIRGMSMarker` and restores every stored prop except the icon. Whenever a mounted marker is removed and re-inserted into the map — which Fabric does to reorder children when any sibling marker's `zIndex` changes — the re-inserted markers lose their custom icon and degrade to default red pins.

This PR caches the decoded `UIImage` when the icon loads and re-applies it synchronously in `didInsertInMap:` (falling back to a reload if no image is cached yet). Re-applying synchronously matters: triggering an async reload instead leaves the marker blank for a few frames, which shows as visible flicker of every reordered marker.

### How did you test this PR?

iOS simulator (iPhone 17 Pro, iOS 26), Google provider, new architecture, in an app with ~15 custom-icon markers. Changing one marker's `zIndex` on selection reorders its siblings: without this change the reordered markers turn into default red pins; with it, every icon persists with no flicker (verified with screenshot bursts during the reorder). Apple provider is unaffected (different marker implementation).


hbasement:
Independent confirmation of this diagnosis, with a second trigger that may help justify prioritizing it.

I hit the same bug from a different direction: on the new architecture, Fabric applies props (so `setIconSrc:`) **before** the marker view is mounted into the map. If the image load happens to finish before insertion, `didInsertInMap:` then replaces `_realMarker` and the icon is gone — exactly the carry-over gap you identified. That makes the failure *timing-dependent* rather than reorder-dependent, which is why it usually looks "random": only markers whose load won the race against insertion degrade to red pins. Anything that makes the icon apply faster (an `RCTImageLoader` cache hit, or local file sources) makes it hit far more markers — in my app every marker turned into a red pin once icon application became synchronous.

Environment: RN 0.81, new architecture, `GoogleMaps 9.4.0`, ~250 markers with locally generated PNG `icon` sources, reproduced on both a real device and the simulator. Your patch is the right shape, and caching the decoded `UIImage` for a synchronous re-apply is better than what I first tried (calling `setIconSrc:` again and relying on the loader cache) — it avoids a reload round-trip entirely.

I've opened #5969 for a separate, non-overlapping issue in the same method's neighborhood: the 0x0 `UIImage` used as the loading placeholder isn't a valid texture under the Metal renderer in Google Maps SDK 9.x, so it fails allocation and shows the default pin anyway on first load. It deliberately doesn't touch `didInsertInMap:`, so there's no conflict with this PR — the two are complementary, and this one should land first.

PR #5951, opened 2026-07-11
feat(ios): add iconScale prop to google maps marker

Bram-dc:

No.

Not a bug — a small feature for the iOS Google provider. The marker `icon` image is always decoded at the device screen scale, so its rendered size is dictated by the asset's pixel size and cannot be controlled (or changed) from JS.

This adds an `iconScale` prop: the pixel density the `icon` is decoded at, i.e. rendered size in points = pixel size / `iconScale` (0 / unset falls back to the screen scale, so existing behaviour is unchanged). Changing the prop reloads the icon in place, which allows e.g. enlarging a marker on selection while reusing a single raster asset.

Wired through the codegen spec, the checked-in generated props, the Fabric view and the paper view manager. iOS-only, matching the `icon` prop's Google-provider scope (the component spec already excludes Android).

### How did you test this PR?

iOS simulator (iPhone 17 Pro, iOS 26), Google provider, new architecture: markers rendered from a single 210 px asset at 56 pt (`iconScale` ≈ 3.75), and resized in place to 70 pt on selection by lowering `iconScale` — the icon swaps without remounting the marker. `yarn tscheck`, `yarn lint` and prettier pass.


Bram-dc:
Pushed a follow-up commit (367ab5b) that normalises the decoded icon back to the screen scale before it reaches the Maps SDK.

**Why.** As this PR stood, `iconScale` was passed straight to `RCTImageLoader`'s `scale:`, which is the point of the feature but leaves the SDK holding a `UIImage` whose `scale` is not the screen scale. Google Maps packs marker icons into a shared texture atlas and sizes each slot from `size * screenScale`; a 210 px asset shown at 56 pt is scale 3.75, so on a 3x screen the bitmap is 1.25x the slot it was given.

A user of my app hit this on a physical device in a release build: a rectangular block of the atlas rendered in place of a marker, showing a grid of *other* markers' artwork clipped to a rectangle, with partial icons at its edges — the signature of sampling neighbouring atlas entries. Notably the selected markers, the one case where `iconScale` happened to equal the screen scale (3.0), were unaffected.

**Honest limits on the evidence.** I have not reproduced this on a simulator — it appears on device only, which is consistent with a texture-packing difference, but it does mean the causal link is reasoned from the arithmetic and the shape of the artifact rather than from a controlled repro. I am confident enough that handing the SDK a conventional screen-scale image is the right thing to do regardless, since it removes the dependency on undefined behaviour, but I did not want to overstate it.

The commit only changes what is handed to the SDK; `iconScale` keeps its meaning (rendered size in points = pixel size / `iconScale`), and rendered marker sizes are unchanged. Verified on an iOS 26 simulator that markers still render at identical sizes before and after.

PR #5959, opened 2026-07-24
fix(android): prevent NullPointerException in applyBaseMapPadding whe…

### LINKED ISSUE #5837: Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.google.android.gms.maps.GoogleMap.setPadding(int, int, int, int)' on a null object reference
### Summary

The logic applies the mapPadding prop calls map.setPadding. If map is null, app crashes.

### Reproducible sample code

```TSX
import {AppState} from 'react-native';
import MapView, {PROVIDER_GOOGLE} from 'react-native-maps';

  const [appState, setAppState] = useState(AppState.currentState); // 'active' | 'background' | 'inactive'
  useEffect(() => {
    const sub = AppState.addEventListener('change', next => setAppState(next));
    return () => sub.remove();
  }, []);

  if (appState === 'background') {
    return null;
  }

  return (
      <MapView
        style={{flex: 1}}
        provider={PROVIDER_GOOGLE}
        mapPadding={{bottom: 160}}
        initialRegion={{
          latitude: 42,
          longitude: 30,
          latitudeDelta: 1,
          longitudeDelta: 1,
        }}
      />
  );
}
```

### Steps to reproduce

Use the new architecture, and set map padding via props. Include logic that unmounts the map when the app is backgrounded (uncommon, but used in some cases to release memory).

Background the app. Resume the app. Repeat until the crash occurs.

### Expected result

Don't crash.

### Actual result

Crashes.

### React Native Maps Version

1.26.20

Android

### React Native Version

0.83.1

Not using Expo

### Device(s)

Samsung Galaxy S24

### Additional information

_No response_

TwickE:

No.

Fixes a `NullPointerException` crash on Android when `applyBaseMapPadding` is called before the native `map` instance is initialized (`GoogleMap map == null`).

Fixes #5837 

### How did you test this PR?

- **Platform & Maps API**: Android (Google Maps)
- **Device**: Tested on a real Android device and Android emulator using React Native / Expo.
- **Verification**: Verified that calling `applyBaseMapPadding` or rendering a `MapView` with `mapPadding` before the map instance finishes loading no longer crashes with a `NullPointerException`. Map padding is deferred safely and applied once the native map instance is ready.

This fix was originally created by: @bradstiff


escapingyouth:
Heads up that this guard on its own may not stop the crash in #5837: `MapManager.setMapPadding` dereferences `view.map` again on the line right after `applyBaseMapPadding` returns —

```java
view.applyBaseMapPadding(left, top, right, bottom);
view.map.setPadding(left, top, right, bottom);   // NPE when the GoogleMap does not exist yet
```

so the deferral added here is reached, and then the next statement throws anyway. `MapView.updateExtraData` has the same problem in its deferred re-apply (`map.getCameraPosition()` guarded only by `setPaddingDeferred` and the view dimensions).

I opened #5983 with your guard plus those two, in case it's useful — happy to close it if you'd rather extend this one.

Issue #5964, opened 2026-07-27
[iOS][New Arch] Google Marker with custom children renders nothing — iconView keeps a CGRectZero frame


spsaucier:
### Summary

On iOS with `PROVIDER_GOOGLE` and the New Architecture, a `<Marker>` with custom children renders **nothing** — no custom view, and no fallback pin. A `<Marker>` with no children renders the default pin correctly, and `<Marker image={...} />` also renders nothing.

This is not a `tracksViewChanges` timing issue. It reproduces with `tracksViewChanges` pinned permanently `true`.

**Root cause: the marker's `iconView` keeps a `CGRectZero` frame.**

`RNMapsGoogleMarkerView.prepareMarkerView` builds its backing paper view standalone via the legacy manager:

```objc
// ios/AirGoogleMaps/RNMapsGoogleMarkerView.mm:99-105
_legacyMarkerManager = [[AIRGoogleMapMarkerManager alloc] init];
_view = (AIRGoogleMapMarker *)[_legacyMarkerManager view];
```

That `AIRGoogleMapMarker` is never added to a UIKit view hierarchy. `mountChildComponentView` forwards the React children into it, which lazily creates `_iconView` and assigns it to `GMSMarker.iconView`:

```objc
// ios/AirGoogleMaps/AIRGoogleMapMarker.m:151-159
if (!_iconView) { _iconView = [[UIView alloc] init]; }   // frame = CGRectZero
if (!_realMarker.iconView) { _realMarker.iconView = _iconView; }
[_iconView insertSubview:subview atIndex:atIndex];
```

`_iconView`'s frame is only ever assigned in two places (`AIRGoogleMapMarker.m:64-92`): `layoutSubviews`, which sizes it to the union of its subviews, and `didUpdateReactSubviews`, which sets it to `self.frame`. Neither runs under Fabric:

- `layoutSubviews` is a UIKit callback that only fires for views participating in a layout pass. This view is detached, so it never gets one.
- `didUpdateReactSubviews` is a Paper/UIManager callback that Fabric does not drive — and `self.frame` is `CGRectZero` regardless, because Fabric assigns layout metrics to `RNMapsGoogleMarkerView`, not to the detached paper view.

So `GMSMarker.iconView` ends up a 0×0 view containing a correctly-sized child. The GoogleMaps iOS SDK 10.x draws that as nothing. On 9.4.0 it rendered, so this is only visible once the pod is on 10.x.

[...]

PR #5966, opened 2026-07-29
fix(ios): lay out Google custom marker views before attachment

### LINKED ISSUE #5953: [iOS][Google Maps][Fabric] Custom markers disappear by viewport sector while panning — regression in 1.29.0 (works in 1.28.2)
### Summary

Since 1.29.0, on **iOS + PROVIDER_GOOGLE + New Architecture**, custom markers (child views, `tracksViewChanges` toggled `true → false` after mount) intermittently disappear when panning the map slightly and reappear after panning again. The dropout pattern follows viewport regions/sectors rather than individual markers. Android is unaffected. Downgrading to **1.28.2 fully resolves it** with no other changes.

### Reproducible sample code

```tsx
import {View, Text} from 'react-native';
import MapView, {Marker, PROVIDER_GOOGLE} from 'react-native-maps';

const VEHICLES = Array.from({length: 20}, (_, i) => ({
  id: `v${i}`,
  latitude: 52.5137 + (i % 5) * 0.001,
  longitude: 13.3903 + Math.floor(i / 5) * 0.0015,
}));

function VehicleMarker({vehicle, selected}) {
  const [tracks, setTracks] = useState(true);
  useEffect(() => {
    const t = setTimeout(() => setTracks(false), 300);
    return () => clearTimeout(t);
  }, []);
  return (
    <Marker
      coordinate={vehicle}
      tracksViewChanges={tracks}
      zIndex={selected ? 10 : 1}
      anchor={{x: 0.5, y: 1}}>
      <View style={{width: 40, height: 48, backgroundColor: 'black', borderRadius: 20, alignItems: 'center', justifyContent: 'center'}}>
        <Text style={{color: 'white', fontWeight: 'bold'}}>M</Text>
      </View>
    </Marker>
  );
}

  const [selectedId, setSelectedId] = useState(null);
  useEffect(() => {
    // Simulate selection changes / data refresh causing sibling zIndex changes
    const t = setInterval(() => setSelectedId(VEHICLES[Math.floor(Math.random() * VEHICLES.length)].id), 3000);
    return () => clearInterval(t);
  }, []);
  return (
    <MapView
      style={{flex: 1}}
      provider={PROVIDER_GOOGLE}
      initialRegion={{latitude: 52.5137, longitude: 13.3903, latitudeDelta: 0.02, longitudeDelta: 0.02}}>
      {VEHICLES.map(v => (
        <VehicleMarker key={v.id} vehicle={v} selected={v.id === selectedId} />
      ))}
    </MapView>
  );
}
```

### Steps to reproduce

1. Render a Google-provider map on iOS (New Architecture) with a couple dozen markers using custom child views and the common `tracksViewChanges` `true → false` toggle (needed so custom views rasterize after layout).
2. Let markers re-mount/reorder naturally (e.g. a sibling marker's `zIndex` changes on selection, or the marker list refreshes from polling).
3. Pan the map slightly.

### Expected result

All markers stay visible while panning; markers whose coordinates are inside the viewport are always drawn.

### Actual result

Groups of markers vanish while panning and reappear after panning further — visibility toggles by viewport region/sector, not per marker. The markers still exist in React state and native children (data is confirmed present; tapping the empty area does nothing until they reappear). Two consecutive screenshots one pan apart show a whole area of markers missing and then back.

### React Native Maps Version

1.29.0 (regression; 1.28.2 works, tested by pinning with no other change)

iOS (Google provider only; Android unaffected)

### React Native Version

0.83.0 (New Architecture, Fabric, bridgeless)

Not using Expo

### Device(s)

iPhone simulator

Happy to test a fix or provide more traces.

MadeinFrance:
## Summary

Fixes the iOS Google Maps custom-marker layout race introduced in 1.29.0.

- attaches a marker child before assigning the marker `iconView`
- lays out the icon view immediately and once on the next main-loop turn, after Fabric has completed child layout
- measures stable view bounds instead of a transient transformed frame, so Reanimated entrance transforms cannot produce a cropped marker snapshot

## Root cause

Under Fabric, custom marker children can be attached before their final layout is available. Google Maps then snapshots a zero or transform-scaled `iconView`; after `tracksViewChanges` is disabled, that invalid snapshot can disappear or render cropped while panning.

## Validation

- inspected against the 1.29.0 source used by the reproducer
- `git diff --check`

Fixes #5953

### DIFF
--- a/ios/AirGoogleMaps/AIRGoogleMapMarker.m
+++ b/ios/AirGoogleMaps/AIRGoogleMapMarker.m
@@ -67,8 +67,10 @@ CGRect unionRect(CGRect a, CGRect b) {
 
     for (UIView *v in [_iconView subviews]) {
 
-        float fw = v.frame.origin.x + v.frame.size.width;
-        float fh = v.frame.origin.y + v.frame.size.height;
+        // frame includes a UIView transform. Custom markers can be scaled by
+        // Reanimated while entering, so use the stable layout bounds instead.
+        float fw = v.bounds.size.width;
+        float fh = v.bounds.size.height;
 
         width = MAX(fw, width);
         height = MAX(fh, height);
@@ -145,6 +147,10 @@ CGRect unionRect(CGRect a, CGRect b) {
     if (_zIndex){
         [_realMarker setZIndex:_zIndex];
     }
+    [self layoutSubviews];
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self layoutSubviews];
+    });
     [_realMarker setMap:map];
 }
 
@@ -152,10 +158,16 @@ CGRect unionRect(CGRect a, CGRect b) {
     if (!_iconView){
         _iconView = [[UIView alloc] init];
     }
+    [_iconView insertSubview:subview atIndex:atIndex];
+    // Size _iconView to fit its children before setting it on the marker.
+    [self layoutSubviews];
     if (!_realMarker.iconView) {
         _realMarker.iconView = _iconView;
     }
-    [_iconView insertSubview:subview atIndex:atIndex];
+    // Fabric lays out children asynchronously, so measure once more next run loop.
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self layoutSubviews];
+    });
 }
 
 - (void)insertReactSubview:(id<RCTComponent>)subview atIndex:(NSInteger)atIndex {


A2999:
Adding a device report, since the PR description mentions only source inspection.

I've been shipping this patch in production on iOS: react-native-maps 1.29.0 + this PR via
patch-package, `PROVIDER_GOOGLE`, New Architecture, RN 0.81.5, Expo SDK 54, on an iPad mini (6th
gen, iPadOS 26.6). Workload is a few hundred custom-view markers continuously added, removed and
regrouped as the camera moves, plus fast multi-level pinch sweeps.

**This PR on its own is enough on my workload**. It's the only modification to the library in the
build I tested. Markers measure and attach correctly, no cropped or mis-sized markers, no survivors
blanking when unrelated markers change, correct first paint on async-loading images, and exact
camera tracking.

For contrast: before this I was pinned to 1.20.x and remounting *every* marker on any change (the
#5736 workaround) to stop survivors blanking out. That machinery is deleted now. I'd like to see
this land; I'm carrying it as a local patch pinned to 1.29.0.

One caveat on all of the above: iOS only. I have no Android measurement.

Separately, two things that aren't library bugs but that I think explain some of the "still broken
after patching" reports on #5971, #5964 and #5877:

1. **Fabric flattens layout-only marker child wrappers.** If a `<Marker>`'s direct child is a
   `View` with no appearance props, Fabric drops it from the native hierarchy and the icon view
   measures the inner content instead, markers draw cropped or mis-scaled. `collapsable={false}`
   fixes it, no native change. I got there by elimination: `onLayout` reported the correct frame,
   forcing `tracksViewChanges` on didn't help, re-latching the icon view didn't help, and simply
   giving the wrapper a border fixed it outright. Filed as #5980.
2. **The 1.29 path leaves Google's default bottom-centre `groundAnchor`** on custom-view markers.
   Mine were centred on 1.20. Since it's a fixed pixel offset, the apparent error grows as you zoom
   out, so it reads as drift rather than a constant offset, which is why it took me a while to
   spot. `anchor={{x: 0.5, y: 0.5}}` fixes it. Filed as #5982.

Anyone hitting either of those will still see broken markers with this patch applied, and may
reasonably conclude the patch doesn't work.

**One defect that does survive this patch**, and it's outside what the PR sets out to fix:
reordering the marker children array so React emits a *move* for an already-mounted `<Marker>`
(remove + re-insert at a different index, same key, same coordinate) flickers unrelated **sibling**
markers. Appends, removals and in-place prop updates are all clean. I isolated it with a JS-only
toggle on a single binary running this PR and no other native change. Stable render order, no
flicker; pre-fix order that relocates a mounted marker, flicker returns. I work around it by never
letting a mounted marker change index. Filed as #5981.

**One observation from reading the diff, which I could not trigger in practice.** The new
measurement uses `v.bounds.size` but no longer reads `v.frame.origin`, while the icon view is still
framed at `(0, 0, w, h)`, so a child laid out at a non-zero origin inside the icon view would be
clipped. I could not construct a case that actually hits this, so treat it as a code-reading note
rather than a bug report, flagging it only in case it matters for a layout I don't have.

Happy to test revisions on device.

PR #5151, opened 2024-08-28
fix(Android): fix multiple padding related issues on Android (#5150, #5152, #5153, #5154)

shoamano83:
### Does any other open PR do the same thing?

No

### What issue is this PR fixing?

- https://github.com/react-native-maps/react-native-maps/issues/5150
- https://github.com/react-native-maps/react-native-maps/issues/5152
- https://github.com/react-native-maps/react-native-maps/issues/5153
- https://github.com/react-native-maps/react-native-maps/issues/5154

### How did you test this PR?

I applied this change to my sample apps

- https://github.com/shoamano83/MapPaddingSample/tree/android_map_padding_on_map_ready
- https://github.com/shoamano83/MapPaddingSample/tree/android_initial_camera_and_map_padding
- https://github.com/shoamano83/MapPaddingSample/tree/android_map_padding_updated
- https://github.com/shoamano83/MapPaddingSample/tree/android_extra_padding

to confirm that padding now works on Android just same as iOS.

PR #5984, opened 2026-08-26
fix(android): restore map padding after reattach

### LINKED ISSUE #5948: [Android] mapPadding is lost when the native map is recreated on screen re-attach (onMapReady re-fires; value-equal prop is never re-sent)
### Summary

On Android (New Architecture / Fabric), when a MapView sits inside a react-navigation screen and that screen is detached and re-attached (e.g. the surrounding bottom-tab set changes while the user is on another screen), the native GoogleMap is recreated - `onMapReady` fires a second time for the same component instance. The `mapPadding` that was in effect is not re-applied to the recreated map: the value applied during the new view's mount is swallowed by the still-initializing map, and because the JS `mapPadding` prop is value-equal across renders, Fabric never re-sends it. The map is left with zero padding indefinitely - visibly, the native My Location button jumps to the raw top-right corner under the system status bar.

This looks like a residual case of the padding-application family fixed by #5151 (#5150 / #5152 / #5153): those fixes (contained in 1.28.2) cover the initial mount, but not a native map recreated after a screen re-attach.

### Reproducible sample code

```tsx
// Map screen registered inside a react-navigation bottom-tabs navigator
// whose tab set changes while this screen is blurred (a conditional tab).
import React from 'react';
import MapView, { PROVIDER_GOOGLE } from 'react-native-maps';

export default function MapScreen() {
  const [ready, setReady] = React.useState(false);
  return (
    <MapView
      style={{ flex: 1 }}
      provider={PROVIDER_GOOGLE}
      showsUserLocation
      showsMyLocationButton
      onMapReady={() => {
        console.log('onMapReady'); // fires AGAIN after the screen re-attaches
        setReady(true);
      }}
      mapPadding={
        ready ? { top: 120, right: 0, bottom: 0, left: 0 } : undefined
      }
      initialRegion={{
        latitude: 42.7,
        longitude: 25.3,
        latitudeDelta: 1,
        longitudeDelta: 1
      }}
    />
  );
}
```

### Steps to reproduce

1. Render the MapView above in a screen of a react-navigation bottom-tabs navigator, with the `mapPadding` top offset applied. The native My Location button renders below the offset (correct).
2. Navigate away (push a screen over the tabs), and while the map screen is blurred change the tab set (add or remove a conditional tab), so the map screen goes through a detach/re-attach.
3. Navigate back to the map screen.
4. Watch the console: `onMapReady` fires a second time - the native map view was recreated. The My Location button now renders under the status bar; the previously applied `mapPadding` is gone and never comes back (the JS prop value did not change, so Fabric dedupes it and nothing reaches the native side).

### Expected result

The recreated native map retains (or re-applies) the `mapPadding` currently set on the component - the My Location button stays below the padding offset.

### Actual result

The recreated map has zero padding; the My Location button renders under the system status bar and stays there indefinitely.

Workaround we ship: nudge the padding value by 1px inside `onMapReady`, forcing a post-init prop update Fabric cannot dedupe. Note for anyone else affected: anchoring the nudge on a navigation focus event is NOT reliable - it races the native map recreation (we reproduced both orderings on different devices); `onMapReady` is the deterministic signal that the recreated map can accept `setPadding`.

### React Native Maps Version

1.28.2

Android

### React Native Version

0.86.0

Not using Expo

### Device(s)

Pixel 9 emulator (Android 16) and physical Android device

### Additional information

- Reproduced on react-native-maps 1.28.2, React Native 0.86.0, New Architecture (Fabric), `googleRenderer` LATEST.
- Not fixed by upgrading: the 1.29.0 changelog contains no padding/restore-related changes.
- Related prior work already contained in 1.28.2: #5151 (fixes #5150/#5152/#5153, initial-mount padding), #5727 (state-restore double-initialization).
- The user-location dot (`showsUserLocation`) survives the recreation, so the loss is specific to the padding state, not the whole prop set.

huytdps13400:

No. I searched open and closed PRs for #5948 and the Android map-padding reattach behavior before starting.

Fixes #5948.

When Android recreates the underlying `GoogleMap`, `onMapReady()` restores cached camera, boundary, and style props through `applyBridgedProps()`, but not the cached base padding. Fabric does not resend a value-equal `mapPadding`, so the recreated map stays at zero padding.

This replays the four cached pixel padding values before restoring camera props. The order preserves the Android invariant established by #5151: apply padding before `moveCamera()` / `moveToRegion()` so both controls and the visible camera region are restored correctly.

### How did you test this PR?

Affected surface: Android + Google Maps. No iOS/Apple Maps code changes.

- RED source invariant against `beta`: `applyBridgedProps()` did not replay `base*MapPadding`
- GREEN invariant: cached padding is replayed before boundary/camera restoration
- `yarn lint`
- `yarn tscheck`
- `yarn format-check`
- `yarn test --runInBand` — 1 suite, 3 tests passed
- JDK 17: `example/android/gradlew -p example/android :react-native-maps:compileDebugJavaWithJavac --no-daemon` — `BUILD SUCCESSFUL`
- `git diff --check`

I could not run the navigation detach/reattach reproduction on an Android device locally. The native lifecycle and Java compilation were verified; the issue reproduction remains the reviewer-facing runtime test plan.


huytdps13400:
CI update: the change-relevant lanes are green (`lint-test` and `build-android`). Both iOS workflows fail during `pod install`, before compilation, because the React Native 0.70 Boost archive checksum differs from the podspec expectation (`expected f0397b…`, `got 71c32f…`). The two jobs fail at the same dependency-download step and no iOS files are changed in this PR. Android CI passed in 2m56s; the local JDK 17 Java compile also passed.

PR #6001, opened 2026-09-12
fix(ios): keep the marker anchored to its coordinate when its content is transformed

### LINKED ISSUE #5999: iOS: a Marker whose content is transformed drifts away from its own coordinate
## Summary

On Apple Maps, a custom `Marker` whose child view carries a `transform: [{scale}]` is drawn away from the coordinate it marks. Scaling the content is a common way to show that a pin is selected, and the pin then visibly slides up and to the left while the callout — anchored to the real coordinate — stays put, so the two look misaligned.

The offset is `size * (scale - 1) / 2`: a 44pt marker at `scale: 1.2` sits 4.4pt off, at `scale: 1.6` it is 13.2pt.

## Version

`react-native-maps` 1.27.2, also present on `master` (1.29.0). New architecture (Fabric), but the code involved is the same on both.

## Steps to reproduce

```jsx
<Marker coordinate={coordinate} anchor={{x: 0.5, y: 0.5}}>
  <Animated.View style={[{width: 44, height: 44, alignItems: 'center', justifyContent: 'center'},
                         {transform: [{scale: selected ? 1.2 : 1}]}]}>
    <View style={{width: 28, height: 28, borderRadius: 14, backgroundColor: 'orange'}} />
  </Animated.View>
</Marker>
```

**Expected:** the circle stays centred on `coordinate` and only grows.
**Actual:** it grows *and* moves 4.4pt up and to the left.

## Cause

`AIRMapMarker.layoutSubviews()` sizes the annotation from the `frame` of its first subview:

```objc
UIView *firstSubView = self.subviews.firstObject;
if (firstSubView && (CGRectGetWidth(firstSubView.frame) > CGRectGetWidth(reactFrame) ||
                     CGRectGetHeight(firstSubView.frame) > CGRectGetHeight(reactFrame))) {
    reactFrame = firstSubView.frame;
}
[self reactSetFrame:reactFrame];
```

In UIKit `frame` is the view's bounds **with its transform already applied**, so the scaled content grows the annotation. MapKit keeps the annotation centred on its coordinate, and the content — whose own layout frame never changed — is then drawn half the gain away from the point it belongs to.

## Suggested fix

Measure the subview by `bounds.size`, the size its own layout gave it, before any transform. PR follows.

## Workaround

Put the transform on a nested view and keep the outer one, the one MapKit measures, fixed. On Fabric that outer view also needs `collapsable={false}`, or view flattening removes it and the scaled view becomes the first subview again.

ivan-kolesov:
Fixes #5999.

## What

`AIRMapMarker.layoutSubviews()` measures its content by `bounds.size` instead of `frame`, so a transform on the marker content no longer moves the annotation off its coordinate.

## Why

In UIKit `frame` is the view's bounds with its transform already applied. Sizing the annotation from the first subview's `frame` therefore grew the annotation whenever the content was scaled — a common way to show that a pin is selected. MapKit keeps the annotation centred on its coordinate, so the content, whose own layout frame never changed, ended up drawn half the gain away from the point it marks: `size * (scale - 1) / 2`, which is 4.4pt for a 44pt marker at `scale: 1.2` and 13.2pt at 1.6. A callout anchored to the real coordinate then looks misaligned with its own pin.

`bounds.size` is the size the subview's own layout gave it, which is what the annotation should be sized from. The annotation still grows when the content really is larger than the marker frame — only a transform no longer counts as larger.

## Test plan

Verified on an iPhone 17 Pro simulator (iOS 26.5, Fabric) with a `Marker` whose child animates to `scale: 1.2` on selection, measuring the centroid of the drawn pin in screenshots before and after selecting it:

- before: the pin centre moves 4.4pt up and to the left when selected;
- after: it moves 0.1pt, while the drawn content still scales (the pin's drawn area grows by the same 1.39x in both builds).

As a sensitivity check, removing the `collapsable={false}` that the usual workaround needs — the condition that previously reproduced the drift and grew it to 13.2pt at `scale: 1.6` — no longer reproduces it at all.

Objective-C only — no JS touched.

## Note on `beta`

This is `master` only: the 2.0 line moved the marker to `ios/Maps/RNMMapMarker.m`, which has no `layoutSubviews` override, so the bug does not exist there and there is nothing to port. CONTRIBUTING asks for PRs against `beta`, but recent merges all land on `master` — happy to retarget if that is wrong.


salah-ghanim (approved):

salah-ghanim (approved):

salah-os-agent (approved):

salah-ghanim:
thanks @ivan-kolesov

